PR #13, opened 2022-10-04
ci: use a sha as an action's ref instead of a release tag

knanao:
This is a just suggestion to be more secure.

This might be helpful when a release tag is overridden or deleted suddenly, especially for un-official actions.
And I also highly recommend configuring `Actions permissions` to allow all actions that this project wanna use explicitly if you have not configured it yet.

https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#allowing-select-actions-and-reusable-workflows-to-run

### DIFF
--- a/.github/workflows/pr-title-validation.yaml
+++ b/.github/workflows/pr-title-validation.yaml
@@ -13,7 +13,7 @@ jobs:
     name: Validate PR title
     runs-on: ubuntu-latest
     steps:
-      - uses: amannn/action-semantic-pull-request@v4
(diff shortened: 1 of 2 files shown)


cre8ivejp:
recheck

knanao:
Oh, let me rebase this from main.

knanao:
recheck

cre8ivejp (approved):
Thank you for the suggestion and the PR 🎉

PR #16, opened 2022-10-05
fix(ui/web-v2): feature flag name not being checked

cre8ivejp:
The feature flag name is a required field, but it wasn't being checked.

Screen Shot 2022-10-05 at 19 58 02


masaaania (approved):
LGTM!!

PR #19, opened 2022-10-07
build: remove go_embed_data rule

masaaania:
- Removed `go_embed_data` rule from `pkg/locale/BUILD.bazel`.
- Introduced //go:embed to read files instead.
  - We need //go:embed to replace Bazel with `go build`.
- Even if this PR is merged, it can be still built with both `bazelisk` and `go build`.


cre8ivejp (approved):
Sorry for the delay!
Nice, thanks!

PR #23, opened 2022-10-12
build: add go build,test commands

masaaania:
- added `go build` and `go test` commands to Makefile.
- `make build-go-apps` creates the executable files (such as `account`, `auditlog`, ... ) in `./bin` directory.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,18 @@ endif
 DELETED_PACKAGES := //proto/external/googleapis/googleapis/83e756a66b80b072bd234abcfe89edf459090974/google/rpc,//proto/external/protocolbuffers/protobuf/v3.18.1/google/protobuf
 LOCAL_IMPORT_PATH := github.com/bucketeer-io/bucketeer
 
+# go applications
+GO_APP_DIRS := $(wildcard cmd/*)
+GO_APP_BUILD_TARGETS := $(addprefix build-,$(notdir $(GO_APP_DIRS)))
+
+ifndef GOOS
+	GOOS := $(shell go env GOOS)
+endif
+
+ifndef GOARCH
+	GOARCH=$(shell go env GOARCH)
+endif
+
 #############################
 # All
 #############################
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -138,6 +150,19 @@ diff-check:
...
+	CGO_ENABLED=0 GOOS=$(GOOS) GOARCH=$(GOARCH) \
+		go build -ldflags "-s -w" \
+		-o bin/$* -mod=vendor cmd/$*/$*.go
+
+.PHONY: build-go-apps
+build-go-apps: $(GO_APP_BUILD_TARGETS)
+
+.PHONY: test-go-apps
+test-go-apps:
+	TZ=UTC CGO_ENABLED=0 go test -v ./pkg/...
```
masaaania, line 164: `TZ=UTC` is added because some tests assume timezone is UTC.

Makefile
```diff
@@ -14,6 +14,18 @@ endif
...
+# go applications
+GO_APP_DIRS := $(wildcard cmd/*)
+GO_APP_BUILD_TARGETS := $(addprefix build-,$(notdir $(GO_APP_DIRS)))
+
+ifndef GOOS
+	GOOS := $(shell go env GOOS)
+endif
+
+ifndef GOARCH
+	GOARCH=$(shell go env GOARCH)
```
cre8ivejp, line 26: nit: `=` -> `:=`

Makefile
```diff
@@ -138,6 +150,19 @@ diff-check:
...
 	go mod tidy
 
+.PHONY: $(GO_APP_BUILD_TARGETS)
+$(GO_APP_BUILD_TARGETS): build-%:
+	CGO_ENABLED=0 GOOS=$(GOOS) GOARCH=$(GOARCH) \
+		go build -ldflags "-s -w" \
+		-o bin/$* -mod=vendor cmd/$*/$*.go
+
+.PHONY: build-go-apps
+build-go-apps: $(GO_APP_BUILD_TARGETS)
```
cre8ivejp, line 160: Because we use the prefix `build-%` [here](https://github.com/bucketeer-io/bucketeer/pull/23/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R154), I think it is better if match the prefix for other places.

E.g.
```sh
make build-go # Build all apps
make build-go-account # Build a single app
make test-go # Test all select directories
```

masaaania: Thx, I fixed it 👍

Makefile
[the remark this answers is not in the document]
masaaania: `make test` is trying to test ./hack and ./cmd, but there doesn't seem to be xxx_test.go files actually in those directories.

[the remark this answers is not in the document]
cre8ivejp: Ok, thanks for checking it!

cre8ivejp:
@masaaania 
May I suggest renaming the commit name to `build: add rules to build and test go files using go command` when you merge the PR?

cre8ivejp (approved):
Thank you!

masaaania:
> May I suggest renaming the commit name to build: add rules to build and test go files using go command when you merge the PR?

Yes, thank you for your advice 😄

PR #25, opened 2022-10-12
build: change protoc to use external dependency files

cre8ivejp:
I'm changing to use the proto dependency files from the `proto/external` instead of [cloning the whole project](https://github.com/bucketeer-io/bucketeer/pull/25/files#diff-5900b5f2bf0529200b3db83b69a71c399f90db130199f220e1d9725454449af1L102-L105) every time the CI runs.
The main reason is that Python CI needs to set up the [Go environment](https://github.com/bucketeer-io/bucketeer/blob/main/python/Makefile#L35) just to clone the project, which is unnecessary.
Also, Web development already uses proto dependency files from the `proto/external`, so I wanted to unify this part.

### DIFF
--- a/python/Makefile
+++ b/python/Makefile
@@ -1,11 +1,11 @@
 PY_FILES := $$(find . -iname '*.py' -not -path "./.venv/*")
 GIT_TOP_DIR := $(shell git rev-parse --show-toplevel)
-PROTOBUF_INCLUDE_DIR := $(GIT_TOP_DIR)/proto/external/protocolbuffers/protobuf/v3.9.0
 PROTO_FOLDERS := $(filter-out $(GIT_TOP_DIR)/proto/external%, $(shell find $(GIT_TOP_DIR)/proto -name '*.proto' -print0 | xargs -0 -n1 dirname | sort --unique))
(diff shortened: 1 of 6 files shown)


python/Makefile
```diff
@@ -31,18 +31,19 @@ gen:
...
 		poetry run python -m grpc_tools.protoc \
 			-I"$(GIT_TOP_DIR)" \
-			-I"$(PROTOBUF_INCLUDE_DIR)" \
-			-I"${GOPATH}/src/github.com/googleapis/googleapis" \
+			-I"$(GIT_TOP_DIR)/proto/external/googleapis/googleapis/83e756a66b80b072bd234abcfe89edf459090974" \
 			--python_out=$(GIT_TOP_DIR)/python/src \
 			$$f/*.proto; \
+	done
+
+	for f in ${PROTO_SERVICE_FOLDERS}; do \
```
cre8ivejp, line 39: We didn't notice because the `test-py` job doesn't return an error, but there were many error logs while generating the proto files. So I split the loop to generate the correct files and deleted the unnecessary include `PROTOBUF_INCLUDE_DIR` directory.

Screen Shot 2022-10-12 at 23 00 47

masaaania (approved):
LGTM! 👍

PR #32, opened 2022-10-18
chore: add sdk version property to proto metrics events

cre8ivejp:
I'm adding `sdk_version` to the proto message so we can monitor the metrics by version instead of all SDKs.


ono-max:
Could you share your plan? I'm not familiar with this implementation context.

cre8ivejp:
The plan is to save the `sdk_version` in the Prometheus so we can filter the metrics by version on the Grafana dashboard.
We discussed this when you presented your RFC last week.

ono-max:
I'm not sure, but is it easy to get the version from all SDK itself?

ono-max (approved):

PR #36, opened 2022-10-19
refactor: do not use localizedError

ono-max:
To support multiple languages in the error message, we have to stop using `localizedError`. This PR fixes one part of them.

### DIFF
--- a/pkg/auditlog/api/api.go
+++ b/pkg/auditlog/api/api.go
@@ -96,7 +96,8 @@ func (s *auditlogService) ListAuditLogs(
 	ctx context.Context,
 	req *proto.ListAuditLogsRequest,
 ) (*proto.ListAuditLogsResponse, error) {
-	_, err := s.checkRole(ctx, accountproto.Account_VIEWER, req.EnvironmentNamespace)
+	localizer := locale.NewLocalizer(locale.NewLocale(locale.JaJP))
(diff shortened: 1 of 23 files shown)


masaaania (approved):
LGTM! 👍

pkg/auditlog/api/api.go
```diff
@@ -96,7 +96,8 @@ func (s *auditlogService) ListAuditLogs(
 	ctx context.Context,
 	req *proto.ListAuditLogsRequest,
 ) (*proto.ListAuditLogsResponse, error) {
-	_, err := s.checkRole(ctx, accountproto.Account_VIEWER, req.EnvironmentNamespace)
+	localizer := locale.NewLocalizer(locale.NewLocale(locale.JaJP))
```
cre8ivejp, line 99: The next implementation will extract the locale information from the request header and set it. Is that correct?

ono-max: Not next, but I'll do it in the near future.

There are still a lot of parts that I have to remove. After removing them, I'll get the language data from headers.

cre8ivejp: Ok, I was just confirming how the next implementation would be implemented.
Thanks.

cre8ivejp (approved):
👍

PR #37, opened 2022-10-20
chore: add json transcoding to track api for testing

cre8ivejp:
I'm adding JSON transcoding to track API for testing.
According to the google documentation, it will convert all query parameters automatically as defined in the `TrackRequest`.

Reference:
`
Any fields in the request message which are not bound by the path template automatically become HTTP query parameters if there is no HTTP request body.
`
https://cloud.google.com/endpoints/docs/grpc-service-config/reference/rpc/google.api#httprule

### DIFF
--- a/proto/gateway/service.proto
+++ b/proto/gateway/service.proto
@@ -66,6 +66,17 @@ message RegisterEventsResponse {
   map<string, Error> errors = 1;
 }
 
+message TrackRequest {
+  string apikey = 1;
+  string userid = 2;
+  string goalid = 3;
+  string tag = 4;
+  int64 timestamp = 5;
+  double value = 6;
+}
+
+message TrackResponse {}
+
 service Gateway {
   rpc Ping(PingRequest) returns (PingResponse) {
     option (google.api.http) = {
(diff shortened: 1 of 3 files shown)


proto/gateway/service.proto
```diff
@@ -66,6 +66,17 @@ message RegisterEventsResponse {
...
 }
 
+message TrackRequest {
+  string apikey = 1;
+  string userid = 2;
+  string goalid = 3;
+  string tag = 4;
+  int64 timestamp = 5;
+  double value = 6;
+}
```
masaaania, line 76: question:
Is it intentional not to write the query parameters in a snake case to match the rest of the document?

masaaania (commented):
lgtm!
But I made a question comment.

proto/gateway/service.proto
cre8ivejp: Yes.
I wanted to change it to camel or snake case, but it is already running in production, so I kept the original format.

masaaania (approved):
LGTM! 👍

PR #39, opened 2022-10-20
chore: change grpc track api name temporarily for testing

cre8ivejp:
I forgot to use a different name for this because we already have the `track` API running on the HTTP server.


masaaania (approved):
lgtm!

PR #43, opened 2022-10-21
chore: update api-gateway proto descriptor value

cre8ivejp:
Before running `make docker-gen`, we must ensure that the cache in the `bazel-proto` directory is up-to-date. Otherwise, it could update the proto descriptor value using old code.


masaaania (approved):
LGTM!

PR #44, opened 2022-10-21
refactor: move extract api key calling to checkRequest func

cre8ivejp:
I'm moving the `extractAPIKeyID` because the query parameter passes the API key in the Track API instead of the request header. 
So I'm refactoring the `getEnvironmentAPIKey` to work for both cases.


masaaania (commented):
I don't think this change will allow for both cases, will it be changed further in the following PR?

masaaania (approved):
OK, thank you!

PR #45, opened 2022-10-21
feat: add track api to grpc server

cre8ivejp:
Because we can use the JSON transcoding to convert to gRPC requests and simplify the code, I moved the Track API from the HTTP server to the gRPC server.

**Note:** The old `trackhandler` code generates a `GoalBatchEvent`. Because it is already deprecated, I implemented it to generate a `GoalEvent` instead.


masaaania (approved):
LGTM!

PR #48, opened 2022-10-24
chore: set version using ldflags

masaaania:
- Using ldflags to set version information for go applications.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,11 @@ ifndef GOARCH
 	GOARCH := $(shell go env GOARCH)
 endif
 
+LDFLAGS_PACKAGE := github.com/bucketeer-io/bucketeer/pkg/ldflags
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -151,8 +156,11 @@ tidy-deps:
 
 .PHONY: $(GO_APP_BUILD_TARGETS)
 $(GO_APP_BUILD_TARGETS): build-%:
+	$(eval VERSION := $(shell git describe --tags --always --abbrev=7))
```
cre8ivejp, line 159: Maybe we could create a rule in the Makefile to generate the version.
I'm saying this because we have many places using the same code to generate it.
It doesn't need to be done in this PR, but let's keep this in mind.

cre8ivejp (approved):
Thank you!

PR #50, opened 2022-10-24
chore: add data retention period to delete-e2e-data-mysql

masaaania:
- added data retention period to delete-e2e-data-mysql


masaaania:
close

PR #55, opened 2022-10-27
chore: change experiment updater cronjob settings

cre8ivejp:
Since there is no cost or performance issue, by executing it every 10 seconds, I'm changing it to improve the e2e test stability.


masaaania (approved):
LGTM!

PR #58, opened 2022-10-31
ci: add rule to update proto_descriptor using protoc instead of bazel

masaaania:
-  I added the rule to update proto_descriptor using protoc instead of bazel
    - So far, we have been updating value.yaml using proto_descriptor.pb from bazel.
- I changed `tools/gen/gen.sh` to `tools/gen/descriptor.sh` and sed command to yq command.
- I deleted `tools/gen/gen.sh`, and `docker-gen` rule from Makefile.


cre8ivejp (approved):
Good good!
Thanks!

PR #62, opened 2022-11-04
chore: support aws kms

ono-max:
I ran the following script and confirmed it works.

## Prototype
```go
package main

import (
	"context"
	"fmt"
	"log"

	"github.com/aws/aws-sdk-go-v2/config"
	"github.com/aws/aws-sdk-go-v2/service/kms"
)

const keyID = "<KEY ID>"

func main() {
	ctx := context.TODO()
	str := `{"body":{"Alert id": 123}}`

	a, _ := NewAwsKMSCrypto(ctx, keyID, "ap-northeast-1")
	json := []byte(str)
	result, err := a.Encrypt(ctx, json)
	if err != nil {
		log.Fatal(err)
	}
	decripted, err := a.Decrypt(ctx, result)
	if err != nil {
		log.Fatal(err)
	}
	fmt.Println(string(decripted)) // => {"body":{"Alert id": 123}}
}

type EncrypterDecrypter interface {
	Encrypt(ctx context.Context, data []byte) ([]byte, error)
	Decrypt(ctx context.Context, data []byte) ([]byte, error)
}

type awsKMSCrypto struct {
	client *kms.Client
	keyID  string
}

func NewAwsKMSCrypto(
	ctx context.Context,
	keyID, region string,
) (EncrypterDecrypter, error) {
	cfg, err := config.LoadDefaultConfig(
		ctx,
		config.WithRegion(region),
	)
	if err != nil {
		return nil, err
	}
	client := kms.NewFromConfig(cfg)
	return awsKMSCrypto{
		client: client,
		keyID:  keyID,
	}, nil
}

func (c awsKMSCrypto) Encrypt(ctx context.Context, data []byte) ([]byte, error) {
	resp, err := c.client.Encrypt(ctx, &kms.EncryptInput{
		Plaintext: data,
		KeyId:     &c.keyID,
	})
	if err != nil {
		return nil, err
	}
	return resp.CiphertextBlob, nil
}

func (c awsKMSCrypto) Decrypt(ctx context.Context, data []byte) ([]byte, error) {
	resp, err := c.client.Decrypt(ctx, &kms.DecryptInput{
		CiphertextBlob: data,
		KeyId:          &c.keyID,
	})
	if err != nil {
		return nil, err
	}
	return resp.Plaintext, nil
}

```


masaaania (dismissed):
lgtm!

pkg/autoops/cmd/server/server.go
```diff
@@ -188,14 +189,24 @@ func (s *server) Run(ctx context.Context, metrics metrics.Metrics, logger *zap.L
 	}
 	u.Path = path.Join(u.Path, webhookPath)
 
-	kmsClient, err := kms.NewKeyManagementClient(ctx)
-	if err != nil {
-		return err
+	// TODO: Get cloudService from command-line flags
+	cloudService := "gcp"
```
cre8ivejp, line 193: Instead of setting it here, you can use the Default property in the `cmd.Flag` even if there is no flag set in the `deployment.yaml`.

E.g.
```go
cloudService: cmd.Flag("cloud-service", "Cloud Service info").Default("gcp").String()
```

cre8ivejp (approved):
Thanks!

PR #65, opened 2022-11-08
chore: add env variables for postgres

ono-max:


manifests/bucketeer/values.yaml
```diff
@@ -1180,6 +1186,12 @@ event-persister-goal-events-kafka:
     alloyDBUser:
     alloyDBPass:
     alloyDBName:
```
cre8ivejp, line 1188: If it is not being used anymore, we could delete the alloy DB values.

ono-max: Can I do it in another PR?

ono-max: Also, I'll remove fields from the go struct then,

masaaania (approved):
lgtm!

manifests/bucketeer/values.yaml
cre8ivejp: No problem, thanks!

cre8ivejp (approved):
Thanks 👍

PR #66, opened 2022-11-09
ci: add validation to check proto-descriptor

masaaania:
- added validation to check proto-descriptor to actions workflow.

### DIFF
--- a/.github/workflows/pr-go.yaml
+++ b/.github/workflows/pr-go.yaml
@@ -104,6 +104,8 @@ jobs:
         run: go install github.com/golang/protobuf/protoc-gen-go@${{ env.PROTOC_GEN_GO_VERSION }}
       - name: Generate proto go files
         run: make proto-go
+      - name: Generate proto go descriptor files
+        run: make proto-go-descriptor
       - name: Check repository updates
         run: make update-repos-check
 
--- a/.github/workflows/pr-go.yaml
+++ b/.github/workflows/pr-go.yaml
@@ -89,6 +89,26 @@ jobs:
       - name: Check proto files
         run: make proto-check
 
+  proto-descriptor-check:
+  runs-on: ubuntu-latest
+  steps:
+    - uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


.github/workflows/pr-go.yaml
```diff
@@ -104,6 +104,8 @@ jobs:
         run: go install github.com/golang/protobuf/protoc-gen-go@${{ env.PROTOC_GEN_GO_VERSION }}
       - name: Generate proto go files
         run: make proto-go
+      - name: Generate proto go descriptor files
+        run: make proto-go-descriptor
```
cre8ivejp, line 108: The `proto-go-description` rule creates the descriptor if needed, but it doesn't return an error if there is a diff.
We should do something similar to this.

https://github.com/bucketeer-io/bucketeer/blob/main/Makefile#L65-L67

cre8ivejp (changes requested):

.github/workflows/pr-go.yaml
cre8ivejp: I just noticed that you are checking the diff by executing the `update-repos-check`.
It would be better to be more explicit by creating a rule to check it separately to keep consistency.

masaaania: Thanks, I fixed it 👍

proto/Makefile
```diff
@@ -38,6 +38,12 @@ remove-go:
 go-descriptor:
 	DIR=$(GIT_TOP_DIR) DESCRIPTOR_PATH=$(GIT_TOP_DIR)/proto bash $(GIT_TOP_DIR)/tools/gen/descriptor.sh
 
+.PHONY: go-descriptor-check
+go-descriptor-check:
```
cre8ivejp, line 42: Adding the dependency rule for check rules seems better, like the other places. What do you think?

E.g.
```sh
.PHONY: go-descriptor-check
go-descriptor-check: go-descriptor
```

PR #67, opened 2022-11-09
chore: remove bazel config files

masaaania:
- removed bazel config files (BUILD.bazel, WORKSPACE, repositories.bzl...)
  - the first commit in this PR removes just all BUILD.bazel.
  - It may be easier to review each commit individually.
- update README.md related bazel


cre8ivejp (approved):
Nice!
Thank you!

PR #68, opened 2022-11-09
chore: display variation name in the variation label

cre8ivejp:
I'm adding the variation name in the input variation label to make it easier to understand which variation is, as a user suggested.

E.g.

```
variation_name - variation_value
```

Screen Shot 2022-11-09 at 14 24 37


masaaania (approved):
LGTM!

PR #77, opened 2022-11-16
docs: add rfc for evaluation count improvements

cre8ivejp:
I'm writing this proposal to solve the following issues.

- The data retrieval is done in milliseconds instead of 5-20 seconds
- Split the data pipeline for Feature Flag and A/B Test, making the self-host process simple and cheaper
- Be able to see the default value count


masaaania (approved):
LGTM! 👍

PR #80, opened 2022-11-22
chore: add event counter and experiment service's envoy config in the event persister

cre8ivejp:
I'm adding these services as part of removing the Bigtable dependency.


masaaania (approved):
LGTM!

PR #84, opened 2022-11-29
fix: event user persister should not ack message when DB returns an error

masaaania:


cre8ivejp (approved):
nit: `fix: event user persister should not ack message when DB returns an error`

PR #87, opened 2022-12-01
chore: remove unnecessary health checks in the envoy settings

cre8ivejp:
I'm removing unnecessary health checks that overwhelm other services when it deploys many pods.
For example, the api-gateway also checks for feature and account services, but they already check their health status.


masaaania (approved):
LGTM!

PR #96, opened 2022-12-07
chore: remove postgreSQL test implementation from the event persister

cre8ivejp:
I'm removing the PostgreSQL test implementation from the event persister because we decided to use BigQuery.


masaaania (approved):
LGTM!! 👍

PR #100, opened 2022-12-08
chore: add expiration to keys for evaluation counts

ono-max:
Sample script to try this implementation

```go
func ExampleClient() {
	rdb := goredis.NewClient(&goredis.Options{
		Addr:     "localhost:6379",
		Password: "", // no password set
		DB:       0,  // use default DB
	})

	key := "key"

	pipe := rdb.Pipeline()

	d := pipe.TTL(ctx, key)

	incr := pipe.Incr(ctx, key)

	_, err := pipe.Exec(ctx)

	if d.Val() == -1 {
		rdb.Expire(ctx, key, time.Second*5)
	}

	fmt.Println(incr.Val(), err)
}


```

### DIFF
--- a/pkg/eventpersister/persister/persister.go
+++ b/pkg/eventpersister/persister/persister.go
@@ -23,6 +23,7 @@ import (
 	"strconv"
 	"time"
 
+	goredis "github.com/go-redis/redis"
 	"github.com/golang/protobuf/proto" // nolint:staticcheck
 	"github.com/golang/protobuf/ptypes"
 	"github.com/golang/protobuf/ptypes/wrappers"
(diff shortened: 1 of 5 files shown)


pkg/eventpersister/persister/persister.go
```diff
@@ -666,11 +667,17 @@ func (p *Persister) createUserEvent(
 func (p *Persister) upsertEvaluationCount(event proto.Message, environmentNamespace string) error {
 	if e, ok := event.(*eventproto.EvaluationEvent); ok {
 		eck := p.key(eventCountKey, e.FeatureId, e.VariationId, environmentNamespace, e.Timestamp)
+		if err := p.setExpiration(eck); err != nil {
```
cre8ivejp, line 670: This implementation will query every time to check if the key exists, which is not efficient.
What we need is to check when the count is created.

E.g.
```go
count, err := p.evaluationCountCacher.Increment(eck)
if err != nil {
	return err
}
if count == 1 {
  p.evaluationCountCacher.Expire(key, time)
}
```

Doing this solves the efficiency problem, but if the EXPIRE fails or a race condition happens, the key could stay in the memory.

We need to execute the INCR and EXPIRE as one command to fix that.
You probably will need to use a Lua script to do this.
Reference:
https://stackoverflow.com/a/21897569

First, I would recommend checking the go-redis library to see if there is a way to achieve this efficiently.

Also, how are you planning to update the current keys in the memory without the expiration?

ono-max:
Resolved conflicts.

pkg/eventpersister/persister/persister.go
```diff
@@ -688,3 +688,43 @@ func (p *Persister) getUserEvaluation(
...
+	// The value of command is available only after the pipeline is executed.
+	// https://redis.uptrace.dev/guide/go-redis-pipelines.html#pipelines
+	d, err := dCmd.Result()
+	if err != nil {
+		return err
+	}
+	if d == -1 {
+		// The expiration may be overriden because of race condition.
+		// However, we don't have to care it because this expiration is not have to be strict.
+		_, err := p.evaluationCountCacher.Expire(key, oneMonth)
```
ono-max, line 702: Even though this implementation makes sure that expiration is set, we have to allow the following case:

1. Persister failed to set expiration.
2. The next evaluation event is sent after 15 days.
3. The first setting evaluation counts will be removed after 46 days, not 31 days.

pkg/eventpersister/persister/persister.go
```diff
@@ -688,3 +688,43 @@ func (p *Persister) getUserEvaluation(
...
+			return err
+		}
+	}
+	return nil
+}
+
+func (p *Persister) countEvent(key string) error {
+	pipe := p.evaluationCountCacher.Pipeline()
+	pipe.Incr(key)
+	dCmd := pipe.TTL(key)
```
cre8ivejp, line 713: I understand that the go-redis only returns the incremented value after the pipe is executed. But using the `pipe.TTL`, won't change the issue of calling two operations every time we increment a key.

I think we can only achieve this using the Lua script.
We can simplify all this using the example below.

E.g

```go
var incr = goredis.NewScript(`
local key = KEYS[1]
local val = redis.call("INCR", key)
if val == 1 then
  redis.call('EXPIRE', key, ARGV[1])
end
return v
`)
keys := []string{"key"}
values := []interface{}{1670819943}
_, err := incr.Run(ctx, rdb, keys, values...).Int()
```

Reference:
https://redis.uptrace.dev/guide/lua-scripting.html#redis-script

pkg/eventpersister/persister/persister.go
ono-max: Okimoto-san said it's a rare case, so we don't have to care it.

masaaania (approved):
Sorry for the late review, LGTM!

cre8ivejp (dismissed):
Nice 🎉

cre8ivejp (approved):
Thank you!

PR #101, opened 2022-12-08
chore: link the goal event to the experiment before sending it to kafka

cre8ivejp:
To solve the ordering of the events published from the api-gateway to Pubsub, I'm changing the event persister to link the goal event to the experiment before sending it to Kafka. 

To link the goal event, we need the evaluation event. If the goal event comes before the evaluation event, it will not acknowledge the Pubsub message until the evaluation event is sent to Kafka. If the evaluation event never came for some reason in the client after 24 hours, the persister will acknowledge the goal event, but it will not send to Kafka.

Also, I have changed the e2e tests to send the goal events before sending the evaluation events to ensure that implementation works.

### DIFF
--- a/pkg/eventpersister/persister/persister.go
+++ b/pkg/eventpersister/persister/persister.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/golang/protobuf/proto" // nolint:staticcheck
 	"github.com/golang/protobuf/ptypes"
+	"github.com/golang/protobuf/ptypes/wrappers"
 	"go.uber.org/zap"
 	"golang.org/x/sync/singleflight"
 
@@ -632,6 +633,7 @@ func (p *Persister) listExperiments(
 						exproto.Experiment_FORCE_STOPPED,
 						exproto.Experiment_STOPPED,
 					},
+					Archived: &wrappers.BoolValue{Value: false},
 				})
 				if err != nil {
 					return nil, err
--- a/pkg/eventpersister/persister/persister.go
+++ b/pkg/eventpersister/persister/persister.go
@@ -25,6 +25,7 @@ import (
 	"github.com/golang/protobuf/proto" // nolint:staticcheck
 	"github.com/golang/protobuf/ptypes"
 	"go.uber.org/zap"
+	"golang.org/x/sync/singleflight"
 
 	"github.com/bucketeer-io/bucketeer/pkg/cache"
 	"github.com/bucketeer-io/bucketeer/pkg/errgroup"
@@ -42,11 +43,21 @@ import (
 	"github.com/bucketeer-io/bucketeer/pkg/storage/v2/mysql"
 	eventproto "github.com/bucketeer-io/bucketeer/proto/event/client"
 	esproto "github.com/bucketeer-io/bucketeer/proto/event/service"
+	exproto "github.com/bucketeer-io/bucketeer/proto/experiment"
 	featureproto "github.com/bucketeer-io/bucketeer/proto/feature"
 )
 
 var (
-	ErrUnexpectedMessageType = errors.New("eventpersister: unexpected message type")
+	ErrUnexpectedMessageType     = errors.New("eventpersister: unexpected message type")
+	ErrExperimentNotFound        = errors.New("eventpersister: experiment not found")
+	ErrNoExperiments             = errors.New("eventpersister: no experiments")
+	ErrInvalidGoalEventTimestamp = errors.New("eventpersister: invalid goal event timestamp")
+)
+
+const (
+	listRequestSize        = 500
+	furthestEventTimestamp = 24 * time.Hour
+	oldestEventTimestamp   = 24 * time.Hour
 )
 
 const (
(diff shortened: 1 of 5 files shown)


pkg/eventpersister/persister/persister.go
```diff
@@ -614,3 +548,141 @@ func (p *Persister) key(
...
+	}
+	return ev, false, nil
+}
+
+func (p *Persister) listExperiments(
+	ctx context.Context,
+	environmentNamespace string,
+) ([]*exproto.Experiment, error) {
+	exp, err, _ := p.flightgroup.Do(
+		fmt.Sprintf("%s:%s", environmentNamespace, "listExperiments"),
```
cre8ivejp, line 621: Using `environment_namespace:listExperiments` as the key for the single flight.

pkg/eventpersister/persister/persister.go
```diff
@@ -614,3 +548,141 @@ func (p *Persister) key(
...
+			cursor := ""
+			for {
+				resp, err := p.experimentClient.ListExperiments(ctx, &exproto.ListExperimentsRequest{
+					PageSize:             listRequestSize,
+					Cursor:               cursor,
+					EnvironmentNamespace: environmentNamespace,
+					Statuses: []exproto.Experiment_Status{
+						exproto.Experiment_RUNNING,
+						exproto.Experiment_FORCE_STOPPED,
+						exproto.Experiment_STOPPED,
```
cre8ivejp, line 633: Because we can still receive delayed goal events, we should count them up to 24hrs after the experiment is finished.

pkg/eventpersister/persister/persister.go
```diff
@@ -419,92 +443,19 @@ func (p *Persister) marshalGoalEvent(e *eventproto.GoalEvent, environmentNamespa
...
 	return string(b), false, nil
 }
 
-func (p *Persister) getEvaluations(
-	e *eventproto.GoalEvent,
-	environmentNamespace string,
-) ([]*featureproto.Evaluation, bool, error) {
-	// Evaluations field in the GoalEvent is deprecated.
-	// The following conditions should be removed once all client SDKs are updated.
-	if e.SourceId == eventproto.SourceId_GOAL_BATCH {
```
cre8ivejp, line 462: This old implementation is not needed anymore, so I removed it.

pkg/eventpersister/persister/persister.go
```diff
@@ -419,92 +443,19 @@ func (p *Persister) marshalGoalEvent(e *eventproto.GoalEvent, environmentNamespa
...
-			zap.String("environmentNamespace", environmentNamespace),
-			zap.String("sourceId", e.SourceId.String()),
-			zap.String("goalId", e.GoalId),
-			zap.String("userId", e.UserId),
-			zap.String("tag", e.Tag),
-			zap.String("timestamp", time.Unix(e.Timestamp, 0).Format(time.RFC3339)),
-		)
-	}
-	m["evaluations"] = evaluations
+	eval := fmt.Sprintf("%s:%d:%s:%s", ev.FeatureId, ev.FeatureVersion, ev.VariationId, reason)
```
cre8ivejp, line 450: This is key to linking the Goal event to the Experiment.

pkg/eventpersister/persister/persister.go
```diff
@@ -614,3 +548,141 @@ func (p *Persister) key(
...
+	}
+	if len(experiments) == 0 {
+		handledCounter.WithLabelValues(codeNoExperiments).Inc()
+		return nil, false, ErrNoExperiments
+	}
+	// Find the experiment by goal ID
+	// TODO: we must change the console UI not to allow creating
+	// multiple experiments running at the same time,
+	// using the same feature flag id and goal id
+	var experiment *exproto.Experiment
```
cre8ivejp, line 584: As described in the comment, we need to change the Experiment API to not allow creating experiments using the same feature flag id, version, and goal id. This case is rare, but we should make the changes soon.

test/e2e/autoops/auto_ops_test.go
```diff
@@ -295,12 +295,12 @@ func TestOpsEventRateBatchWithoutTag(t *testing.T) {
 	time.Sleep(40 * time.Second)
 
 	userIDs := createUserIDs(t, 10)
-	for _, uid := range userIDs {
-		grpcRegisterEvaluationEvent(t, featureID, feature.Version, uid, feature.Variations[0].Id, "")
-	}
 	for _, uid := range userIDs[:6] {
 		registerGoalEventWithEvaluations(t, featureID, feature.Version, goalID, uid, feature.Variations[0].Id)
 	}
+	for _, uid := range userIDs {
```
cre8ivejp, line 301: I changed the sending order of the events to ensure the linking implementation is working in the event persister.

pkg/eventpersister/persister/persister.go
masaaania: listExperiments() should discard the experiments finished more than 24 hours ago?

masaaania (changes requested):

pkg/eventpersister/persister/persister.go
cre8ivejp: Before we list the experiments, we check the goal event timestamp. If it is more than 24 hours, the event will be acknowledged without being sent to Kafka, so it is ok as it is.
Also, I just noticed that we could list only experiments that are not archived to speed up the linking process. I will fix it. Thanks!

masaaania (approved):
Thank you!!

PR #102, opened 2022-12-12
ci: add the e2e and prd-pr steps as options when manually running the push-image workflow

masaaania:


cre8ivejp (approved):
nit: `ci: add the e2e and prd-pr steps as options when manually running the push-image workflow.

Thanks!

PR #103, opened 2022-12-12
chore: add auto ops service in the event persister envoy settings

cre8ivejp:
Adding the auto ops service in the event persister as part of linking the goal event to auto ops rules.


masaaania (dismissed):
LGTM!

masaaania (approved):
👍

PR #105, opened 2022-12-13
ci: add skip_release_note option when running publish_chart workflow manually

cre8ivejp:


.github/workflows/publish_chart.yaml
```diff
@@ -42,6 +47,7 @@ jobs:
           # Because the version is based on the tag, if the publish_chart starts before
           # the release workflow, it will create a chart with an old version.
       - name: Wait for release note to succeed
+        if: ${{ github.event.inputs.skip_release_note == 'false' }}
```
masaaania, line 50: Perhaps `== 'false'` is not needed because `skip_release_note` is boolean.

cre8ivejp: Even if the default value is set to true?
https://github.com/bucketeer-io/bucketeer/pull/105/files#diff-f82bec78431ce3f9b6d83311c38c57f8f0371009add85bd5388f10984fb689b4R12

masaaania: I think `'false'` is determined as string but I'm not sure.

masaaania (approved):

PR #108, opened 2022-12-13
ci: fix api rate limit exceeded for arduino/setup-protoc

cre8ivejp:
Fix reference:
https://github.com/actions/runner-images/issues/602#issuecomment-603035769


masaaania (approved):
LGTM! Thank you!

PR #111, opened 2022-12-14
chore: add default variation id for evaluation count key

ono-max:


pkg/eventpersister/persister/persister.go
```diff
@@ -523,12 +524,16 @@ func userMetadataColumn(environmentNamespace string, key string) string {
 
 func (p *Persister) upsertEvaluationCount(event proto.Message, environmentNamespace string) error {
 	if e, ok := event.(*eventproto.EvaluationEvent); ok {
-		eck := p.newEvaluationCountkey(eventCountKey, e.FeatureId, e.VariationId, environmentNamespace, e.Timestamp)
+		vID := e.VariationId
+		if e.Reason.Type == featureproto.Reason_DEFAULT || e.Reason.Type == featureproto.Reason_CLIENT {
```
cre8ivejp, line 528: The Client SDK doesn't use the `Reason_DEFAULT`, so we can remove it from the condition.
Also, can you add uni tests for those cases?

cre8ivejp (changes requested):

cre8ivejp (approved):
Thanks!

PR #121, opened 2022-12-22
chore: remove expiration setting from the event and user count

cre8ivejp:
Due to performance issues, I'm removing the expiration setting when incrementing the event and user count.
I will add a cronjob to delete keys once daily instead of checking the TTL every time we increment the counter.

Also, I'm changing the key format so we can easily scan the keys by day when we delete them.

```
environment_namespace:kind:feature-id:variation-id:timestamp
↓
environment_namespace:kind:timestamp:feature-id:variation-id
```

### DIFF
--- a/pkg/eventpersister/persister/persister.go
+++ b/pkg/eventpersister/persister/persister.go
@@ -535,6 +535,8 @@ func getVariationID(reason featureproto.Reason_Type, vID string) string {
 func (p *Persister) upsertEvaluationCount(event proto.Message, environmentNamespace string) error {
 	if e, ok := event.(*eventproto.EvaluationEvent); ok {
 		vID := getVariationID(e.Reason.Type, e.VariationId)
+		// To avoid duplication when the request fails, we increment the event count in the end
+		// because the user count is an unique count, and there is no problem adding the same event more than once
 		uck := p.newEvaluationCountkey(userCountKey, e.FeatureId, vID, environmentNamespace, e.Timestamp)
 		if err := p.countUser(uck, e.UserId); err != nil {
 			return err
--- a/pkg/eventpersister/persister/persister.go
+++ b/pkg/eventpersister/persister/persister.go
@@ -23,7 +23,6 @@ import (
 	"strconv"
 	"time"
 
-	goredis "github.com/go-redis/redis"
 	"github.com/golang/protobuf/proto" // nolint:staticcheck
 	"github.com/golang/protobuf/ptypes"
 	"github.com/golang/protobuf/ptypes/wrappers"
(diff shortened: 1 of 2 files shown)


ono-max:
Could you stop removing pipeline implementation since I use it in the event counter?

pkg/eventpersister/persister/persister.go
```diff
@@ -535,14 +535,14 @@ func getVariationID(reason featureproto.Reason_Type, vID string) string {
...
 		vID := getVariationID(e.Reason.Type, e.VariationId)
-		eck := p.newEvaluationCountkey(eventCountKey, e.FeatureId, vID, environmentNamespace, e.Timestamp)
-		if err := p.countEvent(eck); err != nil {
-			return err
-		}
 		uck := p.newEvaluationCountkey(userCountKey, e.FeatureId, vID, environmentNamespace, e.Timestamp)
 		if err := p.countUser(uck, e.UserId); err != nil {
 			return err
 		}
+		eck := p.newEvaluationCountkey(eventCountKey, e.FeatureId, vID, environmentNamespace, e.Timestamp)
```
cre8ivejp, line 542: I changed the order of EventCount and UserCount because if the EventCount succeeded but the UserCount failed, it will increment one more to EventCount until the UserCount succeeds.
The UserCount won't change if the same user sends the same event because it is unique.

pkg/eventpersister/persister/persister.go
```diff
@@ -310,6 +300,16 @@ func (p *Persister) send(messages map[string]*puller.Message) {
 				continue
 			}
 			evs[id] = eventJSON
+			if err := p.upsertEvaluationCount(event, environmentNamespace); err != nil {
```
cre8ivejp, line 303: I moved it to the end because if the upsert in the Bigtable fails, it will duplicate the same event in the Redis until it succeeds.

pkg/eventpersister/persister/persister.go
```diff
@@ -535,14 +535,16 @@ func getVariationID(reason featureproto.Reason_Type, vID string) string {
 func (p *Persister) upsertEvaluationCount(event proto.Message, environmentNamespace string) error {
 	if e, ok := event.(*eventproto.EvaluationEvent); ok {
 		vID := getVariationID(e.Reason.Type, e.VariationId)
-		eck := p.newEvaluationCountkey(eventCountKey, e.FeatureId, vID, environmentNamespace, e.Timestamp)
-		if err := p.countEvent(eck); err != nil {
-			return err
-		}
+		// To avoid duplication when the request fails, we increment the event count in the end
```
cre8ivejp, line 538: @masaaania
I have added the comment.

masaaania (approved):
LGTM!!

PR #122, opened 2022-12-22
chore: use redis instead of druid in GetEvaluationTimeseriesCount

ono-max:


cre8ivejp (changes requested):

pkg/eventcounter/api/api.go
```diff
@@ -315,11 +328,252 @@ func (s *eventCounterService) GetEvaluationTimeseriesCount(
...
+
+func getEventValues(vals []interface{}) ([]float64, error) {
+	eventVals := []float64{}
+	for _, v := range vals {
+		str, ok := v.(string)
+		if !ok {
+			eventVals = append(eventVals, 0)
+			continue
+		}
+		float, err := strconv.ParseFloat(str, 64)
```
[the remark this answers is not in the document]
ono-max: Type-assertion checks if the object has the type. String doesn't have float64 as an interface.

https://go.dev/play/p/hjV_fbJf6mo

manifests/bucketeer/values.yaml
```diff
@@ -1021,6 +1021,11 @@ event-counter:
     mysqlDbName:
     port: 9090
     metricsPort: 9002
+    redis:
+      serverName:
+      addr:
+      poolMaxIdle: 50
```
cre8ivejp, line 1027: Please set the default value to 25 for both.

pkg/eventcounter/api/api.go
```diff
@@ -315,11 +327,170 @@ func (s *eventCounterService) GetEvaluationTimeseriesCount(
...
+		dt, err := statusInternal.WithDetails(&errdetails.LocalizedMessage{
+			Locale:  localizer.GetLocale(),
+			Message: localizer.MustLocalize(locale.InternalServerError),
+		})
+		if err != nil {
+			return nil, statusInternal.Err()
+		}
+		return nil, dt.Err()
+	}
+
```
cre8ivejp, line 376: nit: remove the empty line.

pkg/eventcounter/api/api.go
```diff
@@ -315,11 +327,170 @@ func (s *eventCounterService) GetEvaluationTimeseriesCount(
...
+	timeStamps := getOneMonthTimeStamps(startAt)
+
+	vIDs := getVariationIDs(resp.Feature.Variations)
+
```
cre8ivejp, line 380: Same here.

pkg/eventcounter/api/api_test.go
```diff
@@ -919,6 +921,200 @@ func TestGenInterval(t *testing.T) {
...
+			desc: "error: ErrFeatureIDRequired",
+			input: &ecproto.GetEvaluationTimeseriesCountRequest{
+				EnvironmentNamespace: "ns0",
+			},
+			expectedErr: localizedError(statusFeatureIDRequired, locale.JaJP),
+		},
+		{
+			desc: "success",
+			setup: func(s *eventCounterService) {
+				s.featureClient.(*featureclientmock.MockClient).EXPECT().GetFeature(gomock.Any(), gomock.Any()).Return(
```
cre8ivejp, line 1017: Let's set the arguments for `GetFeature` to ensure that it is passing correctly instead of setting `gomock.Any()`.
Same for `GetEventCounts` and `GetUserCounts`.

E.g.
https://github.com/bucketeer-io/bucketeer/blob/main/pkg/eventpersister/persister/persister_test.go#L382-L389

pkg/eventcounter/api/api_test.go
```diff
@@ -919,6 +921,200 @@ func TestGenInterval(t *testing.T) {
...
+		})
+	}
+}
+
+func TestGetEvaluationTimeseriesCountV2(t *testing.T) {
+	t.Parallel()
+	mockController := gomock.NewController(t)
+	defer mockController.Finish()
+
+	patterns := []struct {
```
cre8ivejp, line 1000: Let's also add an internal error pattern for each mock interface to ensure the expected error is correct.

masaaania (approved):
LGTM! 👍

cre8ivejp (changes requested):

pkg/cache/v3/eventcounter.go
```diff
@@ -0,0 +1,92 @@
...
+func getEventValues(vals []interface{}) ([]float64, error) {
+	eventVals := make([]float64, 0, len(vals))
+	for _, v := range vals {
+		if v == nil {
+			eventVals = append(eventVals, 0)
+			continue
+		}
+		str, ok := v.(string)
+		if !ok {
+			return []float64{}, fmt.Errorf("failed to cast value: %v", v)
```
cre8ivejp, line 57: It would be helpful to know which key was when it failed.

pkg/eventcounter/api/api_test.go
ono-max: >Same for GetEventCounts and GetUserCounts.

We can't do that because the parameters always change in for loop.

ono-max:
@cre8ivejp 

Could you take a look at this PR? Thank you.

pkg/eventcounter/api/api_test.go
cre8ivejp: Since you are using slices, it can be done by looping the mock interface.
The mock interface can be called as much as needed. You just need to ensure that you call it in order.

- Create the test keys for each variation for `EventCount`.
- Create a slice that contains the test values to be returned for which key
- Loop the mock `GetEventCounts` interface and set the keys and the values to be returned

Doing this will ensure that the keys passing to the cacher, the returned values for which key, and the number of calls are correct.

E.g.
```go
keys := []string{...}
values := []string{...}
for i := 0; i < len(keys); i++ {
  s.evaluationCountCacher.(*eccachemock.MockEventCounterCache).EXPECT().GetEventCounts(keys[i]).Return([]float64{values[i]}, nil)
}
```

The same thing can be done for `UserCount`.

ono-max: Thank you. I was able to implement it 👍

ono-max:
@cre8ivejp 

I've updated 👍 Could you take a look at this one? Thank you.

pkg/eventcounter/api/api_test.go
```diff
@@ -1083,17 +1085,26 @@ func TestGetEvaluationTimeseriesCountV2(t *testing.T) {
...
+				endAt := time.Now()
+				startAt, err := genInterval(jpLocation, endAt, 30)
+				assert.NoError(t, err)
+				timeStamps := getOneMonthTimeStamps(startAt)
+				for _, vID := range vIDs {
+					ec := getEventCountKeys(vID, fID, environmentNamespace, timeStamps)
+					s.evaluationCountCacher.(*eccachemock.MockEventCounterCache).EXPECT().GetEventCounts(ec).Return(
+						[]float64{
+							1, 3, 5,
+						}, nil).AnyTimes()
```
cre8ivejp, line 1101: Nice!
Can you remove the `AnyTimes` for both mocks? To ensure that the code is not calling more than it should.

pkg/eventcounter/api/api_test.go
```diff
@@ -919,6 +921,294 @@ func TestGenInterval(t *testing.T) {
...
+	}{
+		{
+			desc: "error: ErrFeatureIDRequired",
+			input: &ecproto.GetEvaluationTimeseriesCountRequest{
+				EnvironmentNamespace: "ns0",
+			},
+			expectedErr: localizedError(statusFeatureIDRequired, locale.JaJP),
+		},
+		{
+			desc: "error: get event counts failed",
```
cre8ivejp, line 1031: Please add an internal error test for `GetFeature`.

pkg/eventcounter/api/api_test.go
```diff
@@ -1083,17 +1085,26 @@ func TestGetEvaluationTimeseriesCountV2(t *testing.T) {
...
+				vIDs := []string{vID0, vID1, defaultVariationID}
+				endAt := time.Now()
+				startAt, err := genInterval(jpLocation, endAt, 30)
+				assert.NoError(t, err)
+				timeStamps := getOneMonthTimeStamps(startAt)
+				for _, vID := range vIDs {
+					ec := getEventCountKeys(vID, fID, environmentNamespace, timeStamps)
+					s.evaluationCountCacher.(*eccachemock.MockEventCounterCache).EXPECT().GetEventCounts(ec).Return(
+						[]float64{
+							1, 3, 5,
```
cre8ivejp, line 1100: The `GetEventCounts` calls 30 times for each variation, but it returns only three values. Let's return the value we should expect and keep the consistency of the test. Otherwise, it looks like a bug.

Also, in the other comment, I mentioned that we should create the return values for each variation to ensure that the code is working as expected. By returning the same `1, 3, 5` value for all variations, we don't know if the code is working correctly.

Same for `GetUserCounts`.

ono-max:
@cre8ivejp Please review this PR🙏

ono-max:
@cre8ivejp Updated 👍

cre8ivejp (approved):
Excellent!
Thank you 🎉

PR #129, opened 2022-12-23
chore: add bigquery querier to eventcounter

masaaania:


Makefile
```diff
@@ -124,7 +124,7 @@ diff-check:
 
 .PHONY: tidy-deps
 tidy-deps:
-	go mod tidy
+	go mod tidy -compat=1.17
```
masaaania, line 127: We need to set this flag because of the change of go 1.17 module management.
https://go.dev/doc/go1.17#go-command

pkg/eventcounter/api/api.go
```diff
@@ -101,6 +106,62 @@ func (s *eventCounterService) Register(server *grpc.Server) {
...
+			Locale:  localizer.GetLocale(),
+			Message: localizer.MustLocalize(locale.InternalServerError),
+		})
+		if err != nil {
+			return nil, statusInternal.Err()
+		}
+		return nil, dt.Err()
+	}
+	s.logger.Debug("result", zap.Any("rows", variationCounts))
+
```
cre8ivejp, line 155: Please remove the empty line.

pkg/eventcounter/api/api_test.go
```diff
@@ -59,10 +59,153 @@ func TestNewEventCounterService(t *testing.T) {
...
+				EnvironmentNamespace: "ns0",
+				StartAt:              now.Add(-30 * 24 * time.Hour).Unix(),
+				EndAt:                now.Unix(),
+			},
+			expectedErr: localizedError(statusFeatureIDRequired, locale.JaJP),
+		},
+		{
+			desc: "success: one variation",
+			setup: func(s *eventCounterService) {
+				s.eventStorage.(*v2ecsmock.MockEventStorage).EXPECT().QueryEvaluationCount(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Return(
```
cre8ivejp, line 115: Please set the parameters using the data from the request, at least for the success cases, instead of `gomock.Any()`.

pkg/eventcounter/api/api_test.go
```diff
@@ -59,10 +59,153 @@ func TestNewEventCounterService(t *testing.T) {
...
+						},
+					},
+				},
+			},
+			expectedErr: nil,
+		},
+		{
+			desc: "success: all variations",
+			setup: func(s *eventCounterService) {
+				s.eventStorage.(*v2ecsmock.MockEventStorage).EXPECT().QueryEvaluationCount(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Return(
```
cre8ivejp, line 152: Same here.

pkg/eventcounter/storage/v2/event.go
```diff
@@ -0,0 +1,174 @@
...
+	rows := make([]*EvaluationEventCount, 0, iter.TotalRows)
+	for {
+		var row EvaluationEventCount
+		err := iter.Next(&row)
+		if err == iterator.Done {
+			break
+		}
+		if err != nil {
+			es.logger.Error(
+				"Failed to convert evaluation event count query result",
```
cre8ivejp, line 148: nit: `Failed to convert the evaluation event count from the query result`

pkg/storage/v2/bigquery/querier.go
```diff
@@ -0,0 +1,114 @@
...
+		registerMetrics(dopts.metrics)
+	}
+	logger := dopts.logger.Named("bigquery")
+	cli, err := bigquery.NewClient(ctx, project)
+	if err != nil {
+		logger.Error("Failed to create bigquery client", zap.Error(err))
+		return nil, err
+	}
+	cli.Location = location
+	return &querier{
```
ono-max, line 77: https://github.com/bucketeer-io/bucketeer/blob/main/pkg/storage/v2/mysql/client.go#L103

It's good to add an options field to querier struct.

pkg/storage/v2/bigquery/metrics.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+
+	"github.com/bucketeer-io/bucketeer/pkg/metrics"
+)
+
+const (
+	operationQuery = "Query"
+
+	codeOK      = "OK"
+	codeUnknown = "Unknown"
```
cre8ivejp, line 29: Can you also add the following codes?

- backendError
- internalError
- duplicate
- notFound
- jobBackendError
- jobInternalError

Error code reference:
https://cloud.google.com/bigquery/docs/error-messages

Implementation reference:
https://github.com/bucketeer-io/bucketeer/blob/main/pkg/storage/v2/mysql/metrics.go#L60-L79

pkg/eventcounter/storage/v2/event.go
```diff
@@ -0,0 +1,174 @@
...
+		{
+			Name:  "featureVersion",
+			Value: featureVersion,
+		},
+		{
+			Name:  "variationIDs",
+			Value: variationIDs,
+		},
+	}
+	es.logger.Debug("query evaluation count",
```
cre8ivejp, line 123: nit: `Query evaluation count`

pkg/eventcounter/cmd/server/server.go
```diff
@@ -253,3 +273,20 @@ func (s *server) createDruidQuerier(ctx context.Context, logger *zap.Logger) (dr
 	}
 	return druid.NewDruidQuerier(brokerClient, *s.druidDatasourcePrefix, druid.WithLogger(logger)), nil
 }
+
+func (s *server) createBigqueryQuerier(
```
cre8ivejp, line 277: Please rename here and the other places.
`Bigquery` -> `BigQuery`

pkg/eventcounter/cmd/server/server.go
```diff
@@ -189,12 +194,27 @@ func (s *server) Run(ctx context.Context, metrics metrics.Metrics, logger *zap.L
 	defer redisV3Client.Close()
 	redisV3Cache := cachev3.NewRedisCache(redisV3Client)
 
+	bigqueryQuerier, err := s.createBigqueryQuerier(ctx, *s.project, *s.bigqueryDataLocation, registerer, logger)
```
cre8ivejp, line 197: Please rename it from `bigquery` -> `bigQuery` and the other places.

pkg/eventcounter/storage/v2/event.go
```diff
@@ -0,0 +1,174 @@
...
+	"go.uber.org/zap"
+	"google.golang.org/api/iterator"
+
+	"github.com/bucketeer-io/bucketeer/pkg/log"
+	"github.com/bucketeer-io/bucketeer/pkg/storage/v2/bigquery"
+	ecproto "github.com/bucketeer-io/bucketeer/proto/eventcounter"
+)
+
+const (
+	DataTypeEvaluationEvents = "evaluation_event"
```
cre8ivejp, line 35: nit: `DataTypeEvaluationEvent`

cre8ivejp (changes requested):

pkg/storage/v2/bigquery/querier.go
```diff
@@ -0,0 +1,114 @@
...
+	dopts := &querierOptions{
+		logger: zap.NewNop(),
+	}
+	for _, opt := range opts {
+		opt(dopts)
+	}
+	if dopts.metrics != nil {
+		registerMetrics(dopts.metrics)
+	}
+	logger := dopts.logger.Named("bigquery")
```
ono-max, line 70: To avoid conflicts with writer for bigquery, can we change "bigquery" to something like "bigquery_querier"?

pkg/storage/v2/bigquery/querier/metrics.go
```diff
@@ -0,0 +1,104 @@
...
+func record() func(operation string, err *error) {
+	startTime := time.Now()
+	return func(operation string, err *error) {
+		if err == nil {
+			handledCounter.WithLabelValues(operation, codeOK).Inc()
+			handledHistogram.WithLabelValues(operation, codeOK).Observe(time.Since(startTime).Seconds())
+			return
+		}
+		var code string
+		var e *googleapi.Error
```
ono-max, line 70: Can we write unit tests for this part by cutting out this part or something?

masaaania: Thanks!
I added it 👍

ono-max:
@masaaania 

I'm worried if this implementation works because of adding service account settings to event counter.

Could you check if this implementation works after merging terraform files?

masaaania:
@ono-max 

> Could you check if this implementation works after merging terraform files?

Yes, I will check it. 
I'm worried too :sweat_smile:

ono-max:
Thank you 👍

pkg/eventcounter/api/api.go
```diff
@@ -89,6 +93,7 @@ func NewEventCounterService(
 		featureClient:                f,
 		accountClient:                a,
 		druidQuerier:                 d,
+		eventStorage:                 v2ecstorage.NewEventStorage(b, bigQueryDataset, l),
```
ono-max, line 96: Can we call this function in cmd/server/server.go such as kafka? If we do that, we don't have to assign bigQueryDataset in `NewEventCounterService` 👍 

https://github.com/bucketeer-io/bucketeer/blob/main/pkg/eventpersister/cmd/server/server.go#L346

masaaania: Although we can call it from cmd/server/server.go, that makes a dependency from `eventcounter/cmd/server` pkg to `eventcounter/storage/v2` pkg.

I don't want to create dependency from `{module}/cmd/server` to `{module}/storage` as possible.

ono-max: > I don't want to create dependency from {module}/cmd/server to {module}/storage as possible.
I thought the same thing, but it could confuse other developers because the way to write the code is different. How about adding comments about those things?

pkg/eventcounter/api/api.go
```diff
@@ -79,16 +81,20 @@ func NewEventCounterService(
...
 	return &eventCounterService{
-		experimentClient:             e,
-		featureClient:                f,
-		accountClient:                a,
-		druidQuerier:                 d,
+		experimentClient: e,
+		featureClient:    f,
+		accountClient:    a,
+		druidQuerier:     d,
+		// v2ecstorage.NewEventStorage() is executed at here instead of at eventcounter/cmd/server because {module}/cmd/server should not have a dependency to {module}/storage
```
cre8ivejp, line 96: Because we are not doing anything special here, I think this comment is unnecessary.
This is the way we implemented all the services.

pkg/eventcounter/api/api.go
```diff
@@ -101,6 +106,86 @@ func (s *eventCounterService) Register(server *grpc.Server) {
 	ecproto.RegisterEventCounterServiceServer(server, s)
 }
 
+// TODO temporary name
+func (s *eventCounterService) GetEvaluationCountBigQuery(
```
cre8ivejp, line 110: No need to change it now. I'm just sharing some ideas for the API name.

- GetExperimentEvaluationCount
- GetExperimentGoalCount

pkg/storage/v2/bigquery/querier/metrics.go
```diff
@@ -0,0 +1,103 @@
...
+
+func getCodeFromError(err error) string {
+	if err == nil {
+		return codeOK
+	}
+	var e *googleapi.Error
+	if ok := errors.As(err, &e); !ok {
+		return codeUnknown
+	}
+	switch e.Code {
```
ono-max, line 78: I think that it would be useful to use https://pkg.go.dev/net/http#StatusText.

ono-max: https://go.dev/src/net/http/status.go

masaaania: It may affect the prometheus or grafana because the string returned by `http.StatusText` is not under our control.
It takes effort, but it's better to return a code we define ourselves.

masaaania:
@ono-max 

> Could you check if this implementation works after merging terraform files?

I checked the event counter works as expected 👍

ono-max (approved):
👍

PR #130, opened 2022-12-23
chore: change the timezone to JP location when saving the count in redis

ono-max:


cre8ivejp (approved):
Thank you!

nit: `chore: change the timezone to JP location when saving the count in redis`

PR #137, opened 2022-12-27
fix: internal error when there is no value in the event count

cre8ivejp:
The service returns an error because SUM returns NULL when there is no value.


ono-max (approved):
I didn't know `IFNULL()`👍

Issue #142, opened 2022-12-28
Default and off variation not being updated when changing variation type


cre8ivejp:
When the user adds many variations and changes the default and off variation to use anything above index 1, changing the variation type from STRING to BOOLEAN will delete the extra variations. Still, it doesn't update the default and off variation.

#### How to reproduce

1- Add a new Feature Flag
2- Change the variation type to STRING
3- Add 3 new variations
4- Change the ON and OFF variation values to variations 4 and 5, respectively
5- Change the variation type to BOOLEAN

![Screen Shot 2022-12-28 at 21 50 35](https://user-images.githubusercontent.com/2486691/209815458-15b5b77a-a6ac-4846-8d28-db62074490f5.png)

Issue #138, opened 2022-12-28
Make it easier to understand why the user cannot delete some variations


cre8ivejp:
This is not an issue. The problem is the variation is being used in the `default variation` or the `off variation` settings, so it cannot be deleted until the user changes those settings to other variations.

It can be improved by adding a tooltip on the delete icon, letting the user know why he can't delete it.

#### If it is used in the default variation settings
- This variation cannot be deleted because it is used in the default variation settings.
- このバリエーションはデフォルトバリエーション設定で使用されているため、削除できません。

#### If it is used in the off-variation settings
- This variation cannot be deleted because it is used in the off-variation settings.
- このバリエーションはオフバリエーション設定で使用されているため、削除できません。

#### If it is used in both
- This variation cannot be deleted because it is used in the default and the off-variation settings
- このバリエーションは、デフォルトおよびオフバリエーション設定で使用されているため、削除できません。

**Note:** The default variation has two types. Fixed and Rollout. To make it simple, let's use the same message for both.

PR #143, opened 2022-12-30
fix(ui): show popup variation rule message while hovering on variation delete button

bimalgrg519:
Fix https://github.com/bucketeer-io/bucketeer/issues/142 https://github.com/bucketeer-io/bucketeer/issues/138

### DIFF
--- a/ui/web-v2/apps/admin/src/components/VariationInput/index.tsx
+++ b/ui/web-v2/apps/admin/src/components/VariationInput/index.tsx
@@ -193,7 +193,6 @@ export const VariationInput: FC<VariationInputProps> = memo(
           {variations.map((variation: any, idx) => {
             const disableRemoveBtn =
               variationType == Feature.VariationType.BOOLEAN.toString() ||
-              variations.length <= 2 ||
               removeDisabledIndexes.has(idx);
             return (
               <div key={idx} className="flex flex-row flex-wrap mb-2">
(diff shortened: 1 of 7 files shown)


cre8ivejp:
Thank you for your PR!

I deployed it and found a bug when the same variation is being used in the default and off variation. It shows the HoverPopover without a message.

Screen Shot 2022-12-31 at 19 57 53

It will be easier to understand and simple if we don't show the delete button when there are only two variations. What do you think?
Also, can you limit the popover's width to 350px?

bimalgrg519:
When there are only two variations and lets say no rule being applied on second variation then we can hide delete button on second variation right?

cre8ivejp:
@bimalgrg519 

> When there are only two variations and lets say no rule being applied on second variation then we can hide delete button on second variation right?

No. If we have only two variations, there is no need to check if there is a rule to the second variation.
We check the list length, and if it is two, we hide the delete button.
If we have three or more variations, then we show the delete buttons using the conditions that you implemented in the PR.

bimalgrg519:
Thank you for the explanation. Sent PR.

cre8ivejp:
@bimalgrg519 

I found that when you have four variations and delete variation 1, it doesn't update the delete button status. Can you take a look, please?
http://localhost:8000/development/features/feature-flag-string/variation

Added four variations
Screen Shot 2023-01-01 at 19 03 59

After you removed variation 1.
Screen Shot 2023-01-01 at 19 04 11

cre8ivejp (changes requested):

ui/web-v2/apps/admin/src/components/VariationInput/index.tsx
```diff
@@ -162,9 +162,9 @@ export const VariationInput: FC<VariationInputProps> = memo(
           case typeDisabled && variationId === offVariationId:
             return f(messages.feature.variationSettings.offVariation);
 
-          // Check if the variation is both the on variation and the off variation
+          // Check if the variation is both the on and off variation
```
cre8ivejp, line 165: nit: `Check if the variation is both on and off variations`

ui/web-v2/apps/admin/src/components/VariationInput/index.tsx
```diff
@@ -148,11 +148,11 @@ export const VariationInput: FC<VariationInputProps> = memo(
       (variationId) => {
         // Use a switch statement to determine the message
         switch (true) {
-          // Check if the variation is both the on variation and the off variation
+          // Check if the variation is both the on and off variations
```
cre8ivejp, line 151: nit: `Check if the variation is both on and off variations`

ui/web-v2/apps/admin/src/components/VariationInput/index.tsx
```diff
@@ -162,7 +162,7 @@ export const VariationInput: FC<VariationInputProps> = memo(
           case typeDisabled && variationId === offVariationId:
             return f(messages.feature.variationSettings.offVariation);
 
-          // Check if the variation is both the on and off variation
+          // Check if the variation is both on and off variation
```
cre8ivejp, line 165: nit: `Check if the variation is both on and off variations`

cre8ivejp (approved):
Thank you!!!

PR #144, opened 2023-01-01
fix: table has no partition for value from column_list error

masaaania:


cre8ivejp (approved):
nit: fix: table has no partition for value from column_list error

Issue #139, opened 2022-12-28
Validation is not working on the targeting rule when the segment user is selected


cre8ivejp:
### Things to fix

- When the user clicks on the `Add rule` button, the `Save with comment` button should stay disabled until the validation returns true
- When the rule condition is `Compare`, it must check the value field has at least one value
- When the rule condition is `User segment`, it must check the segment user ID is selected
- When the rule condition is `Date`, it must check the field

![Screen Shot 2022-12-28 at 20 40 15](https://user-images.githubusercontent.com/2486691/209806799-0aa23248-5b50-4b04-9e5a-d91d754cffd6.png)

PR #149, opened 2023-01-08
fix(ui): fix validation in targeting rule

bimalgrg519:
Fix https://github.com/bucketeer-io/bucketeer/issues/139

### DIFF
--- a/ui/web-v2/apps/admin/src/components/FeatureTargetingForm/index.tsx
+++ b/ui/web-v2/apps/admin/src/components/FeatureTargetingForm/index.tsx
@@ -251,7 +251,7 @@ export const RuleInput: FC<RuleInputProps> = memo(({ feature }) => {
         {
           type: ClauseType.COMPARE,
           attribute: '',
-          operator: Clause.Operator.EQUALS,
+          operator: Clause.Operator.EQUALS.toString(),
           values: [],
         },
       ],
--- a/ui/web-v2/apps/admin/src/components/FeatureTargetingForm/index.tsx
+++ b/ui/web-v2/apps/admin/src/components/FeatureTargetingForm/index.tsx
@@ -12,6 +12,7 @@ import {
 import { useIntl } from 'react-intl';
 import { shallowEqual, useDispatch, useSelector } from 'react-redux';
 import 'react-datepicker/dist/react-datepicker.css';
+import { v4 as uuid } from 'uuid';
 
 import { intl } from '../../lang';
 import { messages } from '../../lang/messages';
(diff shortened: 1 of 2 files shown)


ui/web-v2/apps/admin/src/components/FeatureTargetingForm/index.tsx
```diff
@@ -69,7 +71,17 @@ export const FeatureTargetingForm: FC<FeatureTargetingFormProps> = memo(
         label: createVariationLabel(v),
       };
     });
-    const isValid = Object.keys(errors).length == 0;
+
+    const checkSaveBtnDisabled = useCallback(() => {
```
cre8ivejp, line 75: @bimalgrg519 
Thank you for your PR!

I found a few issues. Can you take a look, please?

- When we select `User segment` and the segment from the menu [1], the `Save with comment` keeps disabled
- We have 2 conditions [2] in the rule. If you delete the first one and add a new one, it glitches the values.

[1]

Screen Shot 2023-01-10 at 11 53 10

[2]

#### How to reproduce
1- Add a `Compare` condition and set its values
2- Add a `Date` condition and set its values
3- Delete the `Compare` condition
4- Click to add a new condition

Screen Shot 2023-01-10 at 12 03 56

cre8ivejp (changes requested):

cre8ivejp:
@bimalgrg519 
I checked your last commit. Can you make not reset the values when we select the same type?

1- Select the `Compare` type and set its values
2- Select the `Compare` again, and the current values will be reset

If we select the same type, it shouldn't do anything.

cre8ivejp:
@bimalgrg519 

When I add a rule, it doesn't select the operator for the first condition.
If I add a second condition, it selects it.
Can you fix it?

Screen Shot 2023-01-11 at 17 31 10

cre8ivejp:
@bimalgrg519 

When I delete a rule, the `Save with comment` button keeps disabled. Can you take a look?

http://localhost:8000/development/features/feature-flag-string/targeting

cre8ivejp:
@bimalgrg519 

When I change the values in the condition, the `Save with comment` button becomes enabled as it should be, but if I revert the changes by putting back the original values, the button keeps enabled. It should be disabled because there are diffs.

It happens for all types. Can you take a look?

Also, when I select the same segment from the select menu, it enables the button. It shouldn't enable because the data is still the same.

cre8ivejp:
@bimalgrg519 

I found another issue when I delete the first condition from the rule.
If you delete the user segment condition, the page will become blank.

http://localhost:8000/development/features/feature-flag-string/targeting

cre8ivejp:
@bimalgrg519 

When I add and delete a new rule, it enables the save button. Can you take a look?

#### How to reproduce

http://localhost:8000/development/features/feature-flag-string/targeting

1- Click on `Add rule`
2- Delete the new rule
3- Check the save button

cre8ivejp (approved):
Thank you!

PR #152, opened 2023-01-11
chore: add service accounts binding to deployments

masaaania:


cre8ivejp (approved):
Thank you!
nit: `chore: add serviceAccountName spec to deployment template`

PR #159, opened 2023-01-16
fix: goal event not being acknowledged correctly

cre8ivejp:
Because the calculator service already adds a buffer of 48 hours before stopping the running experiments, there is no need to validate the timestamp. Also, there is no need to list stopped experiments.

- Changed to list only running experiments
- Removed timestamp validation


masaaania (approved):
Looks Good To Me 👍

cre8ivejp:
@masaaania 
I checked the condition to list the experiments in the calculator service, and it doesn't see the archive flag, so I removed it from the event persister to avoid future issues.

https://github.com/bucketeer-io/bucketeer/blob/28f91b67ebc1a40317c9364515c24815d53ceee1/python/src/lib/calculator/job/calculate_experiments.py#L271-L279

masaaania (approved):
OK, thank you 👍

PR #160, opened 2023-01-16
test: invalid message type in the e2e test

cre8ivejp:
I'm fixing the e2e test by creating a file with the register event type to avoid different implementations in the e2e tests and the gateway package.


masaaania (approved):
LGTM!

PR #166, opened 2023-01-18
chore: mark event id as a failure when failed to marshal

ono-max:


masaaania (approved):
LGTM!

pkg/eventpersisterdwh/persister/goal_event.go
```diff
@@ -118,12 +118,16 @@ func (w *goalEvtWriter) Write(
...
 	}
-	if err := w.writer.AppendRows(ctx, goalEvents); err != nil {
+	fs, err := w.writer.AppendRows(ctx, goalEvents)
+	if err != nil {
 		w.logger.Error(
 			"failed to append rows to goal event",
 			zap.Error(err),
 		)
 	}
+	for id, f := range fs {
```
cre8ivejp, line 128: The `fs` only returns the failed events while trying to Marshal them. If the AppendRows returns an error, it will Ack all the messages. As we spoke, we should Nack the messages if it returns an error.

We need to update the fails map before returning it.

https://github.com/bucketeer-io/bucketeer/pull/166/files#diff-97c603194608262028e81a60fdd000061c42a58131269e8e982772d1f8cd6e30R78

cre8ivejp (approved):
Thank you!

pkg/eventpersisterdwh/persister/goal_event.go
cre8ivejp: As we spoke, the implementation for error handling will be done in another PR.

PR #167, opened 2023-01-18
test: update eventcounter e2e

masaaania:
- Added eventcounter's e2e test for new APIs using BigQuery as upstream.
- Removed eventcounter's e2e test for APIs using Druid as upstream.


cre8ivejp (approved):
Thank you!

PR #168, opened 2023-01-18
chore: mark id as a fail when appends requests is failed

ono-max:


pkg/storage/v2/bigquery/writer/client.go
```diff
@@ -101,41 +93,41 @@ func NewWriter(
 
 func (w *writer) AppendRows(
 	ctx context.Context,
-	msgs [][]byte,
-) error {
+	batches [][][]byte,
+) ([]int, error) {
+	fails := make([]int, len(batches))
```
masaaania, line 98: Shouldn't this be `make([]int, 0, len(batches))` ?

masaaania (approved):
lgtm!

ono-max:
@cre8ivejp 

Could you review this PR? Thx.

cre8ivejp (approved):
Thank you!
nit: chore: mark event id as failed when append requests fail

PR #169, opened 2023-01-18
test: revert auto ops e2e

ono-max:


cre8ivejp (approved):
Thank you!
nit: `test: revert auto ops e2e`

PR #173, opened 2023-01-19
fix: insert timestamp correctly in bigquery

ono-max:
Here is a screen shot to prove this implementation works.
Screen Shot 2023-01-19 at 19 43 12


masaaania (approved):
Thank you!!

cre8ivejp (approved):
Thanks!

PR #175, opened 2023-01-20
chore: update BigQuery query files

masaaania:
- We chose the time-unit column partitioning instead of the ingestion time partitioning.
  - https://cloud.google.com/bigquery/docs/partitioned-tables
  - A pseudo column named `_PARTITIONTIME` doesn't exist when you use the time-unit column partitioning.


cre8ivejp (approved):
Thanks!

PR #176, opened 2023-01-20
fix: eventcounter doesn't return goal counts

masaaania:
- Fixed SQL 
- Added goalID to the SQL where clause to filter


cre8ivejp (approved):
Thanks!

PR #182, opened 2023-01-25
chore: remove needless helm values.yaml

masaaania:
- I added `default` namespace settings to each module.
- I updated hpa API version from v2beta1 to v1.


cre8ivejp (approved):
Nice 🎉

PR #184, opened 2023-01-25
fix: internal error when there are no counts for evaluation time-series count

ono-max:


cre8ivejp (approved):
Looks good!
Thanks!

nit: `fix: internal error when there are no counts for evaluation time-series count`

masaaania (approved):
LGTM!

PR #196, opened 2023-01-27
fix: mysql error alerts when the auto ops rule is already triggered

cre8ivejp:
If the user schedule to turn on a flag, and before the auto ops run, he enables the flag, it will treat it as an error, so I'm changing the ExecuteOperation to return nil if the operation is already done.

#### Things done

- Changed to check if the ops have been triggered outside of the transaction
- Changed not to treat as an error if no rows were affected
- Split the logs in the operation to make debugging easier

### DIFF
--- a/pkg/autoops/api/api.go
+++ b/pkg/autoops/api/api.go
@@ -16,7 +16,6 @@ package api
 
 import (
 	"context"
-	"errors"
 	"net/url"
 	"strconv"
 	"time"
@@ -47,8 +46,6 @@ import (
 	eventproto "github.com/bucketeer-io/bucketeer/proto/event/domain"
 	experimentproto "github.com/bucketeer-io/bucketeer/proto/experiment"
 )
-
-var errAlreadyTriggered = errors.New("auto ops Rule has already triggered")
 
 type options struct {
 	logger *zap.Logger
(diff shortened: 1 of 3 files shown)


pkg/autoops/api/api.go
```diff
@@ -836,6 +833,13 @@ func (s *AutoOpsService) ExecuteAutoOps(
 	if err := s.validateExecuteAutoOpsRequest(req, localizer); err != nil {
 		return nil, err
 	}
+	triggered, err := s.checkIfHasAlreadyTriggered(ctx, localizer, req.Id, req.EnvironmentNamespace)
```
cre8ivejp, line 836: I had to move this implementation from the transaction to check if it had already been triggered because the response returns the `AlreadyTriggered` boolean.

https://github.com/bucketeer-io/bucketeer/pull/196/files#diff-13112ed94680ecaec33373531ff5c65e48cecae142657203e04d967a185684d5L862-L864

pkg/autoops/api/api.go
```diff
@@ -859,23 +863,36 @@ func (s *AutoOpsService) ExecuteAutoOps(
...
 		}
-		if autoOpsRule.AlreadyTriggered() {
-			return errAlreadyTriggered
-		}
 		handler := command.NewAutoOpsCommandHandler(editor, autoOpsRule, s.publisher, req.EnvironmentNamespace)
 		if err := handler.Handle(ctx, req.ChangeAutoOpsRuleTriggeredAtCommand); err != nil {
 			return err
 		}
 		if err = autoOpsRuleStorage.UpdateAutoOpsRule(ctx, autoOpsRule, req.EnvironmentNamespace); err != nil {
+			if err == v2as.ErrAutoOpsRuleUnexpectedAffectedRows {
```
cre8ivejp, line 871: If there are no affected rows, it will not treat as an error. Otherwise, it might trigger the alert.

pkg/autoops/api/operation.go
```diff
@@ -86,15 +98,27 @@ func disableFeature(
 		EnvironmentNamespace: environmentNamespace,
 	}
 	_, err := featureClient.DisableFeature(ctx, req)
-	if code := status.Code(err); code == codes.FailedPrecondition {
-		logger.Warn(
-			"Failed to disable feature",
+	if err != nil {
+		if code := status.Code(err); code == codes.FailedPrecondition {
```
cre8ivejp, line 102: I split the implementation to print warning and error levels to make debugging easier.

masaaania (approved):
lgtm!

PR #199, opened 2023-01-30
chore: rename event persister kafka to evaluation count

cre8ivejp:
As part of migrating to BigQuery.


ono-max (approved):
👍

masaaania (approved):
👍 

I think the name of the service account that will be created is `event-persister`, so we don't have to do anything using terraform.

https://github.com/ca-dp/bucketeer-ops/blob/14afbcaf687d0ab7e339131a751a31d080033073/terraform/modules/srv/service_account.tf#L161

PR #200, opened 2023-01-31
chore: change auto ops count key format

cre8ivejp:
I'm changing the auto ops key format so that we can easily scan the keys by feature id.


masaaania (approved):
LGTM!!

PR #203, opened 2023-02-01
chore: add support for multiple experiments using the same goal id

cre8ivejp:
Currently, if the user runs two or more experiments using the same goal ID, it will only link the goal event to the first item in the experiment list. 
So I'm adding support to create one or more goal events according to the number of experiments that contain the same goal ID.


masaaania (approved):
LGTM!

ono-max (approved):
👍

Issue #180, opened 2023-01-23
feat: Prerequisite feature


cre8ivejp:
This feature allows controlling the feature flag dependencies. It can make flags depend on other flags, making them prerequisites to enable a feature.

PR #209, opened 2023-02-03
feat(ui): add prerequisites in targeting

bimalgrg519:
Solve https://github.com/bucketeer-io/bucketeer/issues/180


cre8ivejp:
@bimalgrg519 

Every time I change the dropdown menu for the prerequisite or delete an item, it makes a `ListFeatures` request.
It should only request once when the user clicks on the `Add prerequisite`.

- Request `ListFeatures` when there are no prerequisites and the user clicks on the `Add prerequisite`
- Request `ListFeatures` when the page is loaded and there are one or more prerequisites

Also, the feature flag ID and the variation are not being validated.

cre8ivejp (changes requested):

ui/web-v2/apps/admin/src/components/FeatureTargetingForm/index.tsx
```diff
@@ -107,6 +113,12 @@ export const FeatureTargetingForm: FC<FeatureTargetingFormProps> = memo(
                 />
               )}
             />
+            <div>
+              <label className="input-section-label">Prerequisites</label>
```
cre8ivejp, line 117: Please add the `Prerequisites` to the lang files.

cre8ivejp (changes requested):

ui/web-v2/apps/admin/src/components/FeatureTargetingForm/index.tsx
cre8ivejp: Please use this `前提条件` for the Japanese.

bimalgrg519: Could you please provide for :

- Add Prerequisite
- Select a flag
- Select a variation

cre8ivejp: @bimalgrg519 

Please use the following.

- Add Prerequisite (前提条件を追加)
- Select a flag (フィーチャーフラグを選択)
- Select a variation (バリエーションを選択)

bimalgrg519:
> Also, the feature flag ID and the variation are not being validated.

What type of validation should be applied for this?

cre8ivejp:
@bimalgrg519 

> What type of validation should be applied for this?

All the fields that are configured in the request and in the commands (AddPrerequisiteCommand, ChangePrerequisiteVariationCommand, etc) should be validated.

ui/web-v2/apps/admin/src/assets/lang/en.json
```diff
@@ -234,9 +235,12 @@
   "feature.list.noResult.searchKeyword": "ID, name and description",
   "feature.offVariation": "off variation",
   "feature.onVariation": "on variation",
+  "feature.prerequisites": "Prerequisites",
   "feature.resetRandomSampling": "Reset random sampling",
   "feature.rule": "Rule rollout percentage",
   "feature.search.placeholder": "ID, Name, Description",
+  "feature.selectFlag": "Select a flag",
```
cre8ivejp, line 242: Please change it to `Select a feature flag`

ui/web-v2/apps/admin/src/pages/feature/targeting.tsx
```diff
@@ -227,13 +253,14 @@ export const FeatureTargetingPage: FC<FeatureTargetingPageProps> = memo(
...
     }, [dispatch, currentEnvironment]);
 
-    if (isLoading) {
-      return (
-        <div className="p-9 bg-gray-100">
-          <DetailSkeleton />
-        </div>
-      );
-    }
+    // if (isLoading) {
```
cre8ivejp, line 256: Is there a reason to comment this code?

cre8ivejp:
@bimalgrg519 

We should disable the `Add prerequisites` button when there are feature flags to add.

Screen Shot 2023-02-08 at 14 03 54

cre8ivejp:
@bimalgrg519 

Can you remove the white background when there are no prerequisites like in the Rule section?

Screen Shot 2023-02-08 at 14 15 13

ui/web-v2/apps/admin/src/pages/feature/targeting.tsx
bimalgrg519: I will remove this. We don't need this anymore.

ui/web-v2/apps/admin/src/pages/feature/detail.tsx
```diff
@@ -73,7 +74,8 @@ export const FeatureDetailPage: FC = memo(() => {
     }
   }, [featureId, dispatch, currentEnvironment]);
 
-  if (!feature || isLoading) {
+  if (!feature) {
+    // if (!feature || isLoading) {
```
cre8ivejp, line 78: Please remove this line.

cre8ivejp:
@bimalgrg519 

After adding all the flags, if you remove them, the `Add Prerequisite` button stays disabled. Even if you try to reload the page, or access another feature flag, the button stays disabled.

http://localhost:8000/development/features/test-tag-filter/targeting

ui/web-v2/apps/admin/src/components/FeatureTargetingForm/index.tsx
```diff
@@ -312,9 +312,7 @@ export const PrerequisiteInput: FC<PrerequisiteInputProps> = memo(
     };
 
     useEffect(() => {
-      if (prerequisites?.length > 0) {
-        dispatchListFeatures();
-      }
+      dispatchListFeatures();
```
cre8ivejp, line 315: @bimalgrg519 

This way the dispatch will always trigger when the page is accessed.

We want to dispatch it when the page is loaded and it has prerequisites.
Or when it doesn't have prerequisites, we want to dispatch it when the `Add Prerequisites` button is clicked.

bimalgrg519: @cre8ivejp  Yeah i agree with your points. But to make add Prerequisites button enable or disable at first, we need total numbers of feature flag. That's why we need to dispatch feature flag list api.

cre8ivejp: @bimalgrg519 
Can't you can solve this by adding a condition inside of `disableAddPrerequisite`?

```ts
useEffect(() => {
  if (prerequisites.length > 0) {
    dispatchListFeatures();
  }
}, []);

const disableAddPrerequisite = useCallback(() => {
  if (prerequisites.length === 0) {
    return false;
  }
  return prerequisites.length === features.length - 1;
}, [prerequisites, features]);
```

bimalgrg519: Thank you so much @cre8ivejp . Perfect solution.

cre8ivejp (dismissed):
Thank you!

yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
cre8ivejp, line 1: @bimalgrg519, we don't need this file.

cre8ivejp: @bimalgrg519 can you take a look at the comments when you have time, please?

bimalgrg519: @cre8ivejp I mistakenly added this file. I have removed it now.

cre8ivejp (approved):
👍

PR #213, opened 2023-02-06
feat(ui): set lang in headers

bimalgrg519:


cre8ivejp (approved):
Thank you!

cre8ivejp:
@bimalgrg519 
I have checked the html source, and it still shows `lang=en` even if I switch to Japanese and refresh the page.
Can you take a look?

ui/web-v2/apps/admin/src/grpc/utils.ts
```diff
@@ -6,6 +7,7 @@ import { getToken } from '../storage/token';
 
 type MetaData = {
   authorization: string;
+  'accept-language': LanguageTypes;
```
cre8ivejp, line 10: To keep consistency, please add the quotes to the `authorization` field also. Or remove them from `accept-language`.

cre8ivejp (changes requested):

ui/web-v2/apps/admin/src/grpc/utils.ts
bimalgrg519: @cre8ivejp  Quotes is added because 'accept-language' has minus (-) symbol and we can not add quote in authorization. It's a javascript object key.

bimalgrg519:
> @bimalgrg519 I have checked the html source, and it still shows `lang=en` even if I switch to Japanese and refresh the page. Can you take a look?

@cre8ivejp  Could you please check this on safari browser once?

bimalgrg519:
> @bimalgrg519
> 
> I have checked on Safari already, and it is not changing.
> 
> Screen Shot 2023-02-07 at 18 03 11


https://user-images.githubusercontent.com/11039098/217221970-9cc9f20f-a3e2-4999-b30b-ccbcb0b47baa.mov

cre8ivejp:
@bimalgrg519 

Please check the page source from the right-click menu on the page, not on the developer tool.

#### Steps to reproduce

- Access http://localhost:8000
- Switch the language to Japanese
- Right-click on the page and click on View Page Source or Show Page Source.
- The page source doesn't update. Even if you refresh the page.

If you reproduce the same steps on ebay.com you can confirm that updates it correctly.

bimalgrg519:
> @bimalgrg519
> 
> Please check the page source by clicking on the right-click menu on the page, not on the developer tool.
> 
> #### Steps to reproduce
> * Access http://localhost:8000
> * Switch the language to Japanese
> * Right-click on the page and click on View Page Source or Show Page Source.
> * The page source doesn't update. Even if you refresh the page.
> 
> You can reproduce the same steps on ebay.com you can confirm that updated it correctly.

@cre8ivejp We can not change view source code html.
Please check this for the answer https://stackoverflow.com/questions/9654737/content-of-html-page-changed-by-jquery-but-view-source-dont-reflect-the-chang

cre8ivejp (approved):
Thank you!

PR #214, opened 2023-02-07
chore: add source id property to metrics event

cre8ivejp:
Because these changes are not in production yet, I'm changing the order of the proto properties to match the evaluation and goal event.


masaaania (approved):
LGTM!

PR #215, opened 2023-02-07
chore: remove bigtable from event persister DWH

cre8ivejp:
Instead of using Bigtable to save the user evaluations, I changed it to evaluate the user to get the needed data to save the goal event.
Doing this, will simplify the data pipeline when supporting a multi-cloud platform and reduce costs.

I will remove the Bigtable client in another PR.


ono-max (approved):
👍

masaaania (approved):
lgtm!

PR #216, opened 2023-02-08
chore: move function saving metrics event to gateway

masaaania:
- move function saving metrics event to the api-gateway
- add new metrics event according to HTTP status code
- add e2e test for metrics event saving

### DIFF
--- a/pkg/gateway/api/api_grpc.go
+++ b/pkg/gateway/api/api_grpc.go
@@ -134,7 +134,6 @@ type grpcGatewayService struct {
 	goalPublisher          publisher.Publisher
 	evaluationPublisher    publisher.Publisher
 	userPublisher          publisher.Publisher
-	metricsPublisher       publisher.Publisher
 	featuresCache          cachev3.FeaturesCache
 	segmentUsersCache      cachev3.SegmentUsersCache
 	environmentAPIKeyCache cachev3.EnvironmentAPIKeyCache
@@ -149,7 +148,6 @@ func NewGrpcGatewayService(
 	gp publisher.Publisher,
 	ep publisher.Publisher,
 	up publisher.Publisher,
-	mp publisher.Publisher,
 	v3Cache cache.MultiGetCache,
 	opts ...Option,
 ) rpc.Service {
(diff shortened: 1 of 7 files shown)


pkg/gateway/api/api_grpc.go
```diff
@@ -769,12 +773,23 @@ func (s *grpcGatewayService) RegisterEvents(
 				}
 				continue
 			}
-			metricsMessages = append(metricsMessages, event)
+			m := &eventproto.MetricsEvent{}
+			if err := ptypes.UnmarshalAny(req.Events[i].Event, m); err != nil {
```
ono-max, line 777: Can we write this code like this?
```suggestion
			if err := ptypes.UnmarshalAny(event.Event, m); err != nil {
```

pkg/gateway/api/api_grpc.go
```diff
@@ -794,6 +809,18 @@ func (s *grpcGatewayService) containsInvalidTimestampError(errs map[string]*gwpr
 	return false
 }
 
+func (s *grpcGatewayService) mergeMaps(
```
ono-max, line 812: nit
```suggestion
func (*grpcGatewayService) mergeMaps(
```

pkg/gateway/api/api_grpc.go
masaaania: https://qiita.com/uchiko/items/1c611f0db618ce9dc0a9

pkg/gateway/api/metrics_event.go
```diff
@@ -0,0 +1,513 @@
...
+func (s *grpcGatewayService) saveUnknownError(event *eventproto.MetricsEvent, env string) error {
+	errorType := ErrorTypeUnknown
+	ev := &eventproto.UnknownErrorMetricsEvent{}
+	if err := ptypes.UnmarshalAny(event.Event, ev); err != nil {
+		return err
+	}
+	if ev.ApiId == eventproto.ApiId_UNKNOWN_API {
+		return MetricsSaveErrUnknownApiId
+	}
+	var tag string
```
ono-max, line 500: Do we need to insert "none" if tag is empty?
https://github.com/bucketeer-io/bucketeer/blob/main/pkg/eventpersisterdwh/persister/evaluation_event.go#L166

cre8ivejp: Not anymore. I have confirmed that all projects are using the latest SDK.

ono-max (approved):
👍

cre8ivejp (approved):
Thank you!

PR #218, opened 2023-02-09
chore: change to always return the error messages in Japanese temporarily

ono-max:


cre8ivejp (approved):
Thank you!
nit: `chore: change to always return the error messages in Japanese temporarily`

PR #220, opened 2023-02-10
fix: metrics event count is not saving to Prometheus

masaaania:


cre8ivejp (approved):
Thank you!
nit: `fix: metrics event count is not saving to Prometheus`

PR #228, opened 2023-02-13
chore: add support for a single evaluation in the evaluate features api

cre8ivejp:
Part of removing Bigtable.


masaaania (dismissed):
lgtm!

pkg/feature/api/feature.go
```diff
@@ -1816,9 +1851,68 @@ func (s *FeatureService) EvaluateFeatures(
...
+			UserEvaluations: &featureproto.UserEvaluations{
+				Id:          userEvaluations.Id,
+				Evaluations: []*featureproto.Evaluation{eval},
+				CreatedAt:   userEvaluations.CreatedAt,
+			}}, nil
+	}
 	return &featureproto.EvaluateFeaturesResponse{UserEvaluations: userEvaluations}, nil
 }
 
+func (s *FeatureService) getTargetFeatures(
```
ono-max, line 1884: Can you add the test case like https://github.com/bucketeer-io/bucketeer/blob/main/pkg/gateway/api/api_grpc_test.go#L2156?

Although I don't add it in the api.go because we'll remove it. It's worth to add the test case here.

pkg/feature/api/feature_test.go
```diff
@@ -1013,6 +1013,244 @@ func TestEvaluateFeatures(t *testing.T) {
 	}
 }
 
+func TestEvaluateSingleFeature(t *testing.T) {
```
ono-max, line 1016: Can you add the test case if req.FeatureId is not set.

cre8ivejp: @ono-max 
That case is already being tested in the `TestEvaluateFeatures`.
I split the test because it was too long.

proto/feature/service.proto
```diff
@@ -301,6 +301,7 @@ message EvaluateFeaturesRequest {
   bucketeer.user.User user = 1;
   string environment_namespace = 2;
   string tag = 3;
+  string feature_id = 4;  // it will evaluate a single feature if set
```
ono-max, line 304: Can you explain the reason why we need to handle single feature instead of creating new API such as EvaluateFeature?

cre8ivejp: The main reason is to avoid code duplication.
Also, this API is for internal services only.

ono-max (approved):

PR #230, opened 2023-02-14
chore: add support to filter feature flags by prerequisites

cre8ivejp:


pkg/feature/domain/feature.go
```diff
@@ -77,6 +77,9 @@ func NewFeature(
 		VariationType: variationType,
 		CreatedAt:     time.Now().Unix(),
 		Maintainer:    maintainer,
+		Prerequisites: []*feature.Prerequisite{},
```
cre8ivejp, line 80: The reason for adding an empty array instead of a null value is that the query becomes more complex because the `JSON_LENGTH` command for a `'null'` or a `["ABC"]` value is 1. Also, `'null'` is not the same as `NULL`.

ono-max (dismissed):
LGTM except for Clone part 👍

cre8ivejp:
@ono-max I have added the empty prerequisite array to the clone feature.
Thanks!

ono-max (approved):

masaaania (approved):
lgtm!

PR #245, opened 2023-02-27
chore: change service account to be optional in the helm manifest

cre8ivejp:


masaaania:
@cre8ivejp 
We also need to update deployments.

https://github.com/bucketeer-io/bucketeer/blob/aeb5bf79ce31e37c8bac88f4fa5249f18b22da6e/manifests/bucketeer/charts/account-apikey-cacher/templates/deployment.yaml#L41

cre8ivejp:
@masaaania, thanks!
I have pushed it.

masaaania (approved):
LGTM!

PR #251, opened 2023-03-06
chore: change web-gateway's LB to L7

masaaania:
- I change web-gateway's LB to L7 load balancer
  - change service type to `NordPort` from `LoadBlancer`.
  - add ingress to create GCP Global HTTPS LoadBalancer.


manifests/bucketeer/charts/web-gateway/templates/envoy-configmap.yaml
```diff
@@ -871,6 +871,6 @@ data:
                     alpn_protocols: ["h2"]
                     tls_certificates:
                       - certificate_chain:
-                          filename: /usr/local/certs/bucketeer-jp/tls.crt
+                          filename: /usr/local/certs/service/tls.crt
                         private_key:
-                          filename: /usr/local/certs/bucketeer-jp/tls.key
+                          filename: /usr/local/certs/service/tls.key
```
masaaania, line 876: - We don't need any TLS certifications because the TLS termination is GCP HTTP LoadBalancer.
- But, `HTTP/2` health checks use TLS.
  -  https://cloud.google.com/load-balancing/docs/health-checks#optional-flags-hc-protocol-http
- GCP health checks don't perform certificate validation actually, so we can use any certs.
  - I selected `service/tls` rather than `bucketeer-jp/tls` because the web-gateway uses it for connecting with other services like `account`, `feature`, etc.
  - https://cloud.google.com/load-balancing/docs/health-check-concepts#criteria-certificates

cre8ivejp (approved):
Nice!

PR #252, opened 2023-03-07
chore: get evaluation count in hourly timestamp

ono-max:
I confirmed go-timezone package works in apine by running the following script:
```go
package main

import (
	"fmt"
	"log"
	"time"

	timezone "github.com/tkuchiki/go-timezone"
)

func main() {
	tz := timezone.New()

	_time := time.Date(2020, 1, 1, 0, 0, 0, 0, time.UTC)
	fixedTime, err := tz.FixedTimezone(_time, "Asia/Tokyo")
	if err != nil {
		fmt.Println("err", err) // => err unknown time zone Europe/Belgrade
	} else {
		fmt.Println(fixedTime)
	}
	tzInfo, err := tz.GetTzInfo("Asia/Tokyo")
	if err != nil {
		log.Fatal("%w", err)
	}
	fmt.Println(tzInfo.LongStandard())
	fmt.Println(tzInfo.ShortStandard())
	fmt.Println(tzInfo.StandardOffset())
	fmt.Println(tzInfo.StandardOffsetHHMM())
	fmt.Println(tzInfo.LongDaylight())
	fmt.Println(tzInfo.ShortDaylight())
	fmt.Println(tzInfo.DaylightOffset())
	fmt.Println(tzInfo.DaylightOffsetHHMM())
}

```

Also, I confirmed this implementation works in dev

Screen Shot 2023-03-08 at 15 07 34
Screen Shot 2023-03-08 at 15 07 42


manifests/bucketeer/charts/event-counter/values.yaml
```diff
@@ -25,6 +25,7 @@ env:
     addr:
     poolMaxIdle: 25
     poolMaxActive: 25
+  timezone: Asia/Tokyo
```
ono-max, line 28: TODO: change this to UTC after checking this implementation in dev.

pkg/cache/v3/eventcounter.go
```diff
@@ -106,6 +153,40 @@ func getUserValues(cmds []*goredis.IntCmd) ([]float64, error) {
...
+func getUserValuesV2(cmds [][]*goredis.IntCmd) ([]float64, error) {
+	userVals := make([]float64, 0, len(cmds))
+	for _, day := range cmds {
+		var totalVal float64
+		for _, hour := range day {
+			val, err := hour.Result()
+			if err != nil {
+				return []float64{}, err
+			}
+			totalVal += float64(val)
```
cre8ivejp, line 183: Because the user count is unique, we can't just sum the values. Otherwise, it can report wrong values.
If the same user sends an event every hour during the day, the current implementation will count as 24 events, but it should be reported as 1.

We will need to use `PFMERGE` instead.
https://redis.io/commands/pfmerge/

cre8ivejp (changes requested):

masaaania (approved):
lgtm!

ono-max:
@cre8ivejp 

I confirmed this implementation works in dev after changing codes after your review.

pkg/cache/v3/eventcounter.go
```diff
@@ -106,10 +156,106 @@ func getUserValues(cmds []*goredis.IntCmd) ([]float64, error) {
...
+
+func (c *eventCounterCache) getUserCountsV2(keys [][]string) (count []float64, err error) {
+	pipe := c.cache.Pipeline()
+	uniqueKeys, err := c.createUniqueKeys(len(keys))
+	if err != nil {
+		return
+	}
+	defer func() {
+		if e := c.deleteKeys(uniqueKeys, pipe); e != nil {
+			err = multierror.Append(err, e)
```
ono-max, line 175: To collect all errors, I use go-multierror package here.

Here is an example:

```
{"severity":"ERROR","logger":"bucketeer-event-counter.server.api","caller":"api/api.go:465","message":"Failed to get user counts","serviceContext":{"service":"bucketeer-event-counter.server","version":"-"},"error":"err: 2 errors occurred:\n\t* error1\n\t* error2\n\n, ...
```

ono-max: https://github.com/hashicorp/go-multierror

masaaania:
Looks good 👍

masaaania (dismissed):

ono-max:
@cre8ivejp 

I confirmed current implementation works correctly, but could you double-check it when having time? Thank you!

Daily timestamp
Screen Shot 2023-03-10 at 18 03 43


Hourly timestamp
Screen Shot 2023-03-10 at 18 35 39

ono-max:
@cre8ivejp 

Updated. Thank you!

cre8ivejp (approved):
Amazing work 🎉 
Thank you!

PR #256, opened 2023-03-10
chore(ui): add the option to delete values individually in the select menu

bimalgrg519:
Solve https://github.com/bucketeer-io/bucketeer/issues/257


cre8ivejp (approved):
Thanks!

PR #260, opened 2023-03-15
chore: set mysql client's connection max idle time instead of lifetime

masaaania:
- SetConnMaxLifetime
  - https://pkg.go.dev/database/sql#DB.SetConnMaxLifetime
- SetConnMaxIdleTime
  - https://pkg.go.dev/database/sql#DB.SetConnMaxIdleTime

### DIFF
--- a/pkg/storage/v2/mysql/client.go
+++ b/pkg/storage/v2/mysql/client.go
@@ -47,9 +47,9 @@ func defaultOptions() *options {
 
 type Option func(*options)
 
-func WithConnMaxIdleTime(cml time.Duration) Option {
+func WithConnMaxIdleTime(it time.Duration) Option {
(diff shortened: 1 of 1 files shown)


pkg/storage/v2/mysql/client.go
```diff
@@ -47,9 +47,9 @@ func defaultOptions() *options {
 
 type Option func(*options)
 
-func WithConnMaxLifetime(cml time.Duration) Option {
+func WithConnMaxIdleTime(cml time.Duration) Option {
```
masaaania, line 50: function `WithConnMaxLifetime()` has not been used from any code.

cre8ivejp (commented):
Thank you!

PR #262, opened 2023-03-22
fix: nil pointer if Reason field is nil

ono-max:


pkg/eventpersister/persister/persister.go
```diff
@@ -285,16 +285,16 @@ func (p *Persister) extractEvents(messages map[string]*puller.Message) environme
 	return envEvents
 }
 
-func getVariationID(reason featureproto.Reason_Type, vID string) string {
-	if reason == featureproto.Reason_CLIENT {
+func getVariationID(reason *featureproto.Reason, vID string) string {
+	if reason != nil && reason.Type == featureproto.Reason_CLIENT {
```
cre8ivejp, line 289: If the reason is nil, we must treat it as an error because it is not normal behavior.
Please create an error such as the one below.

```go
ErrReasonNil         = errors.New("eventpersister: reason is nil")
```

cre8ivejp (approved):
Thank you!

PR #264, opened 2023-03-28
test: update register metrics event test

masaaania:
- I added a seed to generate a random number.
- I changed the duration value in metrics events higher value.


cre8ivejp (approved):
👍

PR #270, opened 2023-04-03
chore: modify timeout second of web-gateway backend

masaaania:
- To handle requests to get MAU, which takes over 40 seconds.


cre8ivejp (approved):
Thank you!

PR #282, opened 2023-04-10
docs: rfc for progressive rollout

ono-max:


docs/rfcs/0041-progressive-rollout.md
```diff
@@ -0,0 +1,285 @@
...
+  string cursor = 3;
+  repeated string feature_ids = 4;
+  OrderBy order_by = 5;
+  OrderDirection order_direction = 6;
+  google.protobuf.BoolValue is_waiting = 7;
+  google.protobuf.BoolValue is_doing = 8;
+  google.protobuf.BoolValue is_done = 9;
+  string search_keyword = 10;
+  google.protobuf.BoolValue is_manual_scheduling = 11;
+  google.protobuf.BoolValue is_automatic_scheduling = 12;
```
masaaania, line 222: IMO:

It might be better to use `status` and `type` instead of bool values like `is_waiting` and `is_manual_scheduling`.
Because there may be more `status` and `type` in the future.

It's up to you to decide to keep or change it.

ono-max: Yeah, the reason for using bool is that I'd like to define these fields as optional. However, it seems to be that `optional` field is introduced in Protocol Buffers v3.12.0.

I'll use them. Thanks

masaaania (approved):

docs/rfcs/0041-progressive-rollout.md
```diff
@@ -0,0 +1,282 @@
...
+
+We'll create `progressive_rollout` table as follows. The `clause` always can be converted into `ProgressiveRolloutScheduleClause` when `type` is MANUAL_SCHEDULE or AUTOMATIC_SCHEDULE.
+
+```sql
+CREATE TABLE IF NOT EXISTS `progressive_rollout` (
+  `id` VARCHAR(255) NOT NULL,
+  `feature_id` VARCHAR(255) NOT NULL,
+  `clause` JSON NOT NULL,
+  `status` INT(11) NOT NULL,
+  `type` INT(11) NOT NULL,
```
cre8ivejp, line 79: Can't you check the type using JSON query in the clause column?

ono-max:
@cre8ivejp @masaaania 

Could you review this PR again? Thank you.

docs/rfcs/0041-progressive-rollout.md
```diff
@@ -0,0 +1,296 @@
...
+```
+
+* proto/autoops/clause.proto
+
+`ProgressiveRolloutManualScheduleClause` is set when Manual Setting is used by users.
+`ProgressiveRolloutAutomaticScheduleClause` is set when Template Setting is used by users.
+
+**NOTE**
+When `ProgressiveRolloutAutomaticScheduleClause` will be converted into `ProgressiveRollout`,
+We need to calculate `time` based on `started_at` and `interval`.
```
cre8ivejp, line 131: Please update this sentence because we no longer have a `started_at` property.

docs/rfcs/0041-progressive-rollout.md
```diff
@@ -0,0 +1,296 @@
...
+
+* proto/autoops/progressive_rollout.proto
+
+`Status` represents the state of operations for ProgressiveRollout. If a operation has not started, `Status` is WAITING.
+If a operation is in progress, `Status` is DOING. If a operaiton is finished, `Status` is DONE.
+All fields in `ProgressiveRollout` are stored to DB as columns.
+The reason of using `google.protobuf.Any` is for ease of expansion.
+
+```proto
+message ProgressiveRollout {
```
cre8ivejp, line 107: I was talking to the Designer about this feature today.
If we add the Type property in the proto and in the DB, we will have more flexibility to create the design making it easier to develop without checking the clause type in many places.
So let's go with the Type property as you suggested before.

E.g.
```proto
message ProgressiveRollout {
  enum Type {
    UNKNOWN = 0;
    MANUAL = 1;
    TEMPLATE = 2;
  }
  enum Status {
    UNKNOWN = 0;
    WAITING = 1;
    RUNNING = 2;
    FINISHED = 3;
    DISABLED = 4;
  }
  string id = 1;
  string feature_id = 2;
  Type type = 3;
  google.protobuf.Any clause = 4;
  Status status = 5;
  int64 created_at = 6;
  int64 updated_at = 7;
}
```

docs/rfcs/0041-progressive-rollout.md
```diff
@@ -0,0 +1,296 @@
...
+    DONE = 2;
+  }
+  string id = 1;
+  string feature_id = 2;
+  google.protobuf.Any clause = 3;
+  Status status = 4;
+  int64 created_at = 5;
+  int64 updated_at = 6;
+  // When the operation of Progressive Rollout is stopped temporary, `disabled` is true.
+  bool disabled = 7;
```
cre8ivejp, line 120: What do you think about adding a `DISABLED` in the Status instead of adding a bool? It would make the query easier and display the current status in the UI.

ono-max:
@cre8ivejp 

Thank you for your review. Updated.

docs/rfcs/0041-progressive-rollout.md
```diff
@@ -0,0 +1,291 @@
...
+			Time: 1672499160,
+			Weight: 40000,
+		},
+		{
+			// '2023-01-01 00:09:00'
+			Time: 1672499340,
+			Weight: 60000,
+		},
+	},
+}
```
masaaania, line 62: nit:

This example may be confusing because the interval is less than 5 minutes.

> The interval of time for each scheduled time must be at least 5 minutes.

docs/rfcs/0041-progressive-rollout.md
```diff
@@ -0,0 +1,291 @@
...
+## Progress bar of rollout
+
+![progress-bar-of-rollout](./images/0041-image4.png)
+
+# Important Notice
+
+* Users can use Progressive Rollout when the number of variations is equal to 2.
+* Users can't use same scheduled time in single auto ops rules. For example, users can not set true for 50% at 2023-01-01 00:06:00 and 80% at the same time.
+* The interval of time for each scheduled time must be at least 5 minutes.
+* Users can stop Progressive Rollout temporary.
```
masaaania, line 29: question:

When a user resumes a Progressive Rollout that has been stopped, how do we deal with Schedules that are already past their time?

docs/rfcs/0041-progressive-rollout.md
ono-max: Sorry, I think that example is 3:00, 6:00, and 9:00. I think that it is correct?

docs/rfcs/0041-progressive-rollout.md
ono-max: That's right. It seems to be that a temporary stop is impossible.

ono-max: Let me change it. Thank you.

docs/rfcs/0041-progressive-rollout.md
```diff
@@ -0,0 +1,291 @@
...
+4. AutoOps service calls `UpdateFeatureTargeting` to update feature rules. In this case, update the weight of vid-1 to 20000 and the weight of vid-2 to 80000.
+
+# Changes
+
+## Table
+
+We'll create `ops_progressive_rollout` table as follows. `ProgressiveRolloutManualScheduleClause` and `ProgressiveRolloutAutomaticScheduleClause` are converted into Any type and stored into `clause` column.
+
+```sql
+CREATE TABLE IF NOT EXISTS `ops_progressive_rollout` (
```
cre8ivejp, line 77: Please add the type field to the table definition as described here.
https://github.com/bucketeer-io/bucketeer/pull/282/files#r1166241238

docs/rfcs/0041-progressive-rollout.md
masaaania: Each example's schedule have 00:03:00, 00:06:00, and 00:09:00, whose interval is 3 minutes.

ono-max:
@cre8ivejp @masaaania

Could you review this PR again? Thank you.

docs/rfcs/0041-progressive-rollout.md
ono-max: Talked at discord.

masaaania (approved):
Looks Good To Me 👍

cre8ivejp (dismissed):
Nice!
Thank you!

PR #285, opened 2023-04-17
chore: add proto file for progressive rollout

ono-max:


proto/autoops/service.proto
```diff
@@ -160,6 +161,67 @@ message ListWebhooksResponse {
...
+}
+
+message GetProgressiveRolloutResponse {
+  ProgressiveRollout progressive_rollout = 1;
+}
+
+message DeleteProgressiveRolloutRequest {
+  string id = 1;
+  string environment_namespace = 2;
+  DeleteProgressiveRolloutCommand command = 3;
```
masaaania, line 183: question:

What is `DeleteProgressiveRolleoutCommand` needed for?

ono-max: This is used in https://github.com/bucketeer-io/bucketeer/blob/main/pkg/autoops/command/auto_ops_rule.go#L50.

For example DeleteFeatureCommand is defined in https://github.com/bucketeer-io/bucketeer/blob/main/proto/feature/command.proto#L50.

proto/autoops/progressive_rollout.proto
```diff
@@ -0,0 +1,39 @@
...
+
+package bucketeer.autoops;
+option go_package = "github.com/bucketeer-io/bucketeer/proto/autoops";
+
+import "google/protobuf/any.proto";
+
+message ProgressiveRollout {
+  enum Type {
+    MANUAL_SCHEDULE = 0;
+    AUTOMATIC_SCHEDULE = 1;
```
cre8ivejp, line 25: Since both types are automatic operations, what do you think about renaming this to `TEMPLATE_SCHEDULE`?

proto/autoops/service.proto
```diff
@@ -160,6 +161,67 @@ message ListWebhooksResponse {
...
+
+message ExecuteProgressiveRolloutRequest {
+  string environment_namespace = 1;
+  string id = 2;
+  ChangeProgressiveRolloutTriggeredAtCommand
+      change_progressive_rollout_triggered_at_command = 3;
+}
+
+message ExecuteProgressiveRolloutResponse {
+  bool already_triggered = 1;
```
cre8ivejp, line 222: Just confirming, are you going to use this flag?
I was checking the current code, and it only uses it for debugging.

https://github.com/bucketeer-io/bucketeer/blob/main/pkg/opsevent/batch/executor/executor.go#L74-L79

Removing it from the response will be better if you don't plan to use it.
If you want to use it, returning the `triggered_at` will be better.

proto/autoops/service.proto
```diff
@@ -160,6 +161,67 @@ message ListWebhooksResponse {
...
+  string environment_namespace = 2;
+}
+
+message GetProgressiveRolloutResponse {
+  ProgressiveRollout progressive_rollout = 1;
+}
+
+message DeleteProgressiveRolloutRequest {
+  string id = 1;
+  string environment_namespace = 2;
```
cre8ivejp, line 182: To keep the consistency, can you set the `environment_namespace` as the first property in the message?
Same for the other messages.

proto/autoops/service.proto
cre8ivejp: Giving it a second thought because this is different compared to the auto ops rules. If you need to return something, it should be the current status. What do you think?

proto/autoops/service.proto
```diff
@@ -160,6 +161,67 @@ message ListWebhooksResponse {
...
+  string cursor = 3;
+  repeated string feature_ids = 4;
+  OrderBy order_by = 5;
+  OrderDirection order_direction = 6;
+  optional ProgressiveRollout.Status status = 7;
+  string search_keyword = 8;
+  optional ProgressiveRollout.Type type = 9;
+}
+
+message ListProgressiveRolloutResponse {
```
cre8ivejp, line 209: The total count is missing.

proto/autoops/service.proto
```diff
@@ -160,6 +161,67 @@ message ListWebhooksResponse {
...
+  repeated ProgressiveRollout progressive_rollouts = 1;
+  string cursor = 2;
+}
+
+message ExecuteProgressiveRolloutRequest {
+  string environment_namespace = 1;
+  string id = 2;
+  ChangeProgressiveRolloutTriggeredAtCommand
```
cre8ivejp, line 217: I didn't understand what is this command for.
Can you describe how the `time` will be used on the server?

proto/autoops/service.proto
ono-max: To keep the consistency (for debugging the same as auto ops) with auto ops rules, I added already_triggered flag.

ono-max: Let me remove it. Thank you.

proto/autoops/service.proto
ono-max: @cre8ivejp 

I added the comment. Does it make sense?

cre8ivejp: The executor changes the `triggered_at` in the auto ops rules because it has that property in the proto message. But the progressive rollout doesn't have a `triggered_at` field.
So it's hard to understand what this command is for.

ono-max: Talked at discord.

ono-max:
@cre8ivejp 

Updated 👍 Thank you.

proto/autoops/clause.proto
```diff
@@ -58,3 +58,30 @@ message WebhookClause {
...
+    HOURLY = 1;
+    DAILY = 2;
+    WEEKLY = 3;
+  }
+  // The reason of setting `schedules` is to save `triggered_at` in each
+  // schedule.
+  repeated ProgressiveRolloutSchedule schedules = 1;
+  Interval interval = 2;
+  int64 increments = 3;
+  string variation_id = 4;
```
ono-max, line 86: FYI: I also added variation_id field.

proto/autoops/service.proto
```diff
@@ -160,6 +161,66 @@ message ListWebhooksResponse {
...
+  string environment_namespace = 1;
+  int64 page_size = 2;
+  string cursor = 3;
+  repeated string feature_ids = 4;
+  OrderBy order_by = 5;
+  OrderDirection order_direction = 6;
+  optional ProgressiveRollout.Status status = 7;
+  string search_keyword = 8;
+  optional ProgressiveRollout.Type type = 9;
+  int64 total_count = 10;
```
cre8ivejp, line 207: The `total_count` must be in the response.

ono-max:
@cre8ivejp 

Done. Thank you.

cre8ivejp (approved):
Thank you!

masaaania (approved):
LGTM!

PR #288, opened 2023-04-20
chore: remove feature persisting user data

masaaania:
- This PR removes the feature persisting user data to MySQL
  - We don't use the data now, and this process is using too many MySQL resources.
  - We decided to remove the service until we decide again what to do with the user metadata.

- This PR removes the dependency to feature service from user-persister.
  - This dependency was already not in use.


cre8ivejp (approved):
Thanks!

PR #289, opened 2023-04-20
test: delete user e2e test

masaaania:
- This PR deletes the e2e test for user module.
  - We will create an e2e test for MAU later.


cre8ivejp (approved):
Thanks!

PR #291, opened 2023-04-20
chore: add time range option in the proto for evaluation count request

cre8ivejp:
Added time range options to display the time-series count in the Evaluation tab.


proto/eventcounter/timeseries.proto
```diff
@@ -23,6 +23,11 @@ message VariationTimeseries {
 }
 
 message Timeseries {
+  enum Unit {
```
ono-max, line 26: I think that it'll be better to define this filed in GetEvaluationTimeseriesCountResponse. What do you think?

ono-max (approved):
👍

PR #292, opened 2023-04-21
chore: remove user metadata from user event

masaaania:
- This PR removes the user metadata from the user event.
  - We don't use the metadata now, so removes it until we decide again what to do with.


cre8ivejp (approved):
Looks good!

PR #293, opened 2023-04-24
chore:  add function to refresh features cache

masaaania:
- This PR adds a function to refresh the feature flags cache in the Redis.

### DIFF
--- a/pkg/feature/api/feature.go
+++ b/pkg/feature/api/feature.go
@@ -25,6 +25,7 @@ import (
 	"google.golang.org/genproto/googleapis/rpc/errdetails"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/types/known/wrapperspb"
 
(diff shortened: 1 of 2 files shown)


pkg/feature/api/feature.go
```diff
@@ -2106,3 +2215,27 @@ func (s *FeatureService) CloneFeature(
...
+
+func (s *FeatureService) refreshFeaturesCache(ctx context.Context, environmentNamespace string) error {
+	fs, _, _, err := s.listFeatures(
+		ctx,
+		mysql.QueryNoLimit,
+		"",
+		nil,
+		"",
+		nil,
+		wrapperspb.Bool(false),
```
masaaania, line 2227: To have the same behavior as existing implementations, I added the `wrapperspb.Bool(false)` temporarily to filter out  archived features from the Redis.

When we adopt the new evaluation method, we must set this arg to `nil` in order to include archived features to response.

cre8ivejp (approved):
Thanks!

cre8ivejp (approved):
Thanks!

PR #295, opened 2023-04-25
chore: add TTL to the put function of the cache

masaaania:
- This PR adds a TTL to the Put function of the cache.
- This PR also deletes obsolete codes related to the cache.


pkg/cache/v3/experiments.go
```diff
@@ -24,6 +24,10 @@ import (
 	experimentproto "github.com/bucketeer-io/bucketeer/proto/experiment"
 )
 
+const (
+	experimentsTTL = 0
+)
+
 type ExperimentsCache interface {
```
cre8ivejp, line 31: Since we are not using the `ExperimentsCache`, we can delete it as well.

masaaania: Thankyou, I deleted it 👍

cre8ivejp (approved):
Thank you!

PR #297, opened 2023-04-26
fix: out of memory while trying to delete a segment

cre8ivejp:
The app enters an infinite loop state when the environment has more than 500 feature flags (listRequestSize).

I could fix it by replacing the `mysql.QueryNoLimit` with the `listRequestSize` const, but to avoid this situation, I refactored it to get all features in one query. If we have any performance issues in the future, we can think of something else.


masaaania (approved):
LGTM!

PR #306, opened 2023-05-02
refactor: move functions getting prerequisite to feature domain

masaaania:
- This PR moves functions getting prerequisites from the api-gateway domain to the feature domain since those functions are used in domains other than the api-gateway.


pkg/feature/domain/evaluation.go
```diff
@@ -117,7 +117,7 @@ func TopologicalSort(features []*featureproto.Feature) ([]*featureproto.Feature,
 		for _, p := range f.Prerequisites {
 			pf, ok := mapFeatures[p.FeatureId]
 			if !ok {
-				return errFeatureNotFound
+				return ErrFeatureNotFound
```
ono-max, line 120: Do you have a reason why we need to change from `errFeatureNotFound` to `ErrFeatureNotFound`?

masaaania: There is no need, thx!

masaaania: @ono-max 
I fixed it 👍

ono-max (approved):
👍

PR #318, opened 2023-05-12
chore: add tag filtering to evaluation by evaluatedAt

masaaania:
- Adds tag filtering to the evaluation by `evaluatedAt`.
- Renames variable `isUserAttributesUpdated` to `userAttributesUpdated`

### DIFF
--- a/pkg/feature/domain/evaluation.go
+++ b/pkg/feature/domain/evaluation.go
@@ -160,6 +160,8 @@ func evaluate(
 		}
 		// VariationId is used to check if prerequisite flag's result is what user expects it to be.
 		flagVariations[f.Id] = variation.Id
+		// When the tag is set in the request, it will return only the evaluations of flags that match the tag configured when the flag was created on the dashboard.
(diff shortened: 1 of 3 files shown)


pkg/feature/domain/evaluation.go
```diff
@@ -158,6 +160,9 @@ func evaluate(
 		}
 		// VariationId is used to check if prerequisite flag's result is what user expects it to be.
 		flagVariations[f.Id] = variation.Id
+		if targetTag != "" && !tagExist(f.Tags, targetTag) {
```
cre8ivejp, line 163: It would be nice to comment on how the tag works here.

E.g.
```go
// When the tag is set in the request, it will evaluate only the flags that match the tag configured when the flag was created on the dashboard.
// When empty, it will evaluate all the flags in the environment. 
```

masaaania: @cre8ivejp 
Thx!
I added a comment!

cre8ivejp (approved):
Nice!

PR #329, opened 2023-05-22
chore: add Dependabot

ono-max:


.github/dependabot.yaml
```diff
@@ -0,0 +1,18 @@
...
+    schedule:
+      interval: "weekly"
+  - package-ecosystem: "npm"
+    directory: "/ui"
+    schedule:
+      interval: "weekly"
+  - package-ecosystem: "pip"
+    directory: "/python"
+    schedule:
+      interval: "weekly"
```
masaaania, line 18: It would be better to set the interval to monthly as in the existing flow to reduce the toil.

.github/dependabot.yaml
```diff
@@ -0,0 +1,18 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
```
cre8ivejp, line 3: Please add the prefix `build` to all the `package-ecosystem` and include the `scope`.

```yml
commit-message:
  prefix: "build"
  include: "scope"
```

[Reference](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#commit-message)

.github/dependabot.yaml
```diff
@@ -0,0 +1,18 @@
...
+    directory: "/"
+    schedule:
+      interval: "monthly"
```
cre8ivejp, line 6: Please set the timezone `Asia/Tokyo` for all the schedules and the time to `10:00`

masaaania (approved):

cre8ivejp (approved):
Nice, thanks!

Issue #325, opened 2023-05-19
chore: support multi language for audit log


ono-max:
### Describe the feature you'd like

support multi language for audit log

PR #335, opened 2023-05-24
chore: add language field to Recipient

ono-max:
Part of https://github.com/bucketeer-io/bucketeer/issues/325

Recipient is one of the columns in the subscription field. Here is the process flow.

1. Assign subscription.Recipient.Language in [send method](https://github.com/bucketeer-io/bucketeer/blob/main/pkg/notification/sender/sender.go#L114)
2. Create localizer in pkg/notification/sender/notifier/slack.go.
3. Call `locale.NewLocalizer`.


proto/notification/recipient.proto
```diff
@@ -19,8 +19,13 @@ option go_package = "github.com/bucketeer-io/bucketeer/proto/notification";
 
 message Recipient {
   enum Type { SlackChannel = 0; }
+  enum Language {
+    Japanese = 0;
```
cre8ivejp, line 23: Can you set English as the first option?

cre8ivejp: Also, please use upper case letters for enum values.

cre8ivejp (approved):
👍

masaaania (approved):
LGTM!

PR #338, opened 2023-05-25
fix: wrong user counts in 24 hours

ono-max:
This PR fixes the following things:

* Truncate hour in the current time when the unit is 24 hours
* PF count all keys when the unit is 24 hours. PF Merge keys when the unit is not 24 hours.


cre8ivejp (approved):
Thank you!

PR #355, opened 2023-05-29
revert: remove dependabot

ono-max:
Since we can not configure the `-compat` flag in Dependabot, we'll stop using Dependabot.

See https://github.com/dependabot/dependabot-core/issues/5463.


masaaania (approved):
lgtm!

PR #358, opened 2023-05-31
chore: change service-accounts name

masaaania:
- This PR changes the service-account name because they have conflicts. 
  - e.g., There is a name conflict between `event-persister-evaluation-events-dwh` and `event-persister-goal-events-dwh`.


cre8ivejp (approved):
Nice!
Thanks!

PR #360, opened 2023-05-31
chore: stop updating pip dependency group

ono-max:
We plan to remove python implementation in the feature. Thus, we don't need to update python dependencies.


cre8ivejp (approved):
Thank you!

masaaania (approved):
Thank you!

PR #372, opened 2023-06-01
chore: remove renovate.json

ono-max:
Since we use Dependabot to update dependencies, we don't need renovate.


masaaania (approved):
Thank you!

PR #382, opened 2023-06-02
chore: stop upserting daily evaluation count keys

ono-max:
We don't need to store daily evaluation count keys since we started using hourly evaluation count keys in https://github.com/bucketeer-io/bucketeer/pull/307.


cre8ivejp (approved):
Thanks!

masaaania (approved):
LGTM!

PR #394, opened 2023-06-05
feat: feature evaluation

bimalgrg519:


ui/web-v2/apps/admin/src/pages/feature/index.tsx
```diff
@@ -211,7 +211,12 @@ export const FeatureIndexPage: FC = memo(() => {
         options && options.hasExperiment
           ? options.hasExperiment === 'true'
           : null;
-      const tags = options && options.tagIds ? options.tagIds : [];
+
+      const tags = Array.isArray(options?.tagIds)
```
cre8ivejp, line 215: If the `options` can be null, it will crash.
Shouldn't this be as follows?

```js
const tags = options && Array.isArray(options.tagIds)
  ? options.tagIds
  : [];
```

bimalgrg519: @cre8ivejp  The variable options.tagIds can be either a string or a array or strings. Api will accept only array of strings thats why we need to make array of string if there is only one tagId.

cre8ivejp: The API sends the `tagIds` as an array. I don't understand why it can be a string if the length of the array is 1.
Are you saying when there is only one element, the `tagIds` will be a string `'test'` instead of an array `['test']`?

cre8ivejp: @bimalgrg519 

It will be safer if we check it is a string and is not empty.

```js
const tags = options && Array.isArray(options.tagIds)
  ? options.tagIds
  : typeof options.tagIds === 'string' && options.tagIds.length > 0
  ? [options.tagIds]
  : [];
```

bimalgrg519: Thank you @cre8ivejp .

cre8ivejp (approved):
Thanks!

PR #379, opened 2023-06-01
feat: build httpstan image

Ubisoft-potato:
### What I am doing?

This pull request is part of imlementation of this [issue](https://github.com/bucketeer-io/bucketeer/issues/356), used to create the Docker image for the [HttpStan](https://github.com/stan-dev/httpstan) 

### Summary
This commit adds four new files to the tools/httpstan directory: `Dockerfile`, `Makefile`, `README.md`, and `httpstan_api_test.py` 

* The `Dockerfile` is used to build the HttpStan Docker image
* The `Makefile` includes targets for building and pushing the Docker image
* The `httpstan_api_test.py` script contains a function that tests the `POST /v1/models` endpoint, which is used to compile a model. As you can see, the stan script that used to test HttpStan compile API is the script that Python Calculator was using
* The `README.md` explains how to build and run the HttpStan Docker image and how to test the HttpStan API using the `httpstan_api_test.py` script

Issue #356, opened 2023-05-30
feat: remove python dependency from calculator service


cre8ivejp:
## Summary

Currently, the calculator service is written in Python, using Bayesian Algorithm to calculate the experiment probabilities.
Due to the following reasons, we want to rewrite the service in Go and, for Bayesian analysis, use the HttpStan.

- Maintenance costs
- Contribution and Development Learning costs
- Improve the developing environment compatibility

Besides the Bayesian analysis, we can rewrite the entire calculator service in Go without any issues.
For the analysis, we can deploy a new service for HttpStan and the Go app to make requests to it.

It will also allow us to have the following features.

- Automatic caching of compiled Stan models
- Automatic caching of samples from Stan models
- Parallel sampling

Currently, the calculator service calculates the probabilities once at a time. The Parallel sampling will also help us to improve this.

## What the calculator does

- List all experiments from all environments
- Get the experiment's evaluation counter and goal counter from the event counter service
- Deliver those data to the Pystan to calculate the probabilities
- Save the experiment's probability into the DB
- Stop the experiment if needed

## Implementation

To simplify the implementation, we can implement it as an API in the experiment service, so we won't need to make changes in the helm chart, CI, etc. Also, most APIs we need for this task, the experiment service, have it already, and since it is related to the experiment, having this API in the experiment service makes sense.

- Create the docker image for the HttpStan and implement it in the helm chart ([Reference](https://github.com/bucketeer-io/bucketeer/tree/main/tools/dex))
- Reimplement the Python code in the experiment service as an API
- Write unit and e2e tests ([E2E test reference](https://github.com/bucketeer-io/bucketeer/blob/main/test/e2e/eventcounter/eventcounter_test.go#L525))
- Implement Kubernetes cron job ([Reference](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/))
- Delete unnecessary GitHub Action Workflows ([PR](https://github.com/bucketeer-io/bucketeer/blob/main/.github/workflows/pr-python.yaml), [Push image](https://github.com/bucketeer-io/bucketeer/blob/main/.github/workflows/push-image.yaml#L185-L252)) and their container images

### Source code

- https://github.com/bucketeer-io/bucketeer/tree/main/python
- https://github.com/stan-dev/httpstan

PR #395, opened 2023-06-05
feat: python calculator service refactor in golang

Ubisoft-potato:
## What I am doing?
* This pull request is part of imlementation of this [issue](https://github.com/bucketeer-io/bucketeer/issues/356).  
* Implement the Stan interface (using HttpStan, as we have already [created the HttpStan image](https://github.com/bucketeer-io/bucketeer/pull/379)) in the experiment service, so we can use it in the experiment service.
* Rewrite the [Python Calculator](https://github.com/bucketeer-io/bucketeer/tree/main/python)'s business logic using Goalng.

### New dependency
* [go-resty/resty](https://github.com/go-resty/resty): Simple HTTP and REST client library for Go 
* [go-gota/gota](https://github.com/go-gota/gota): Golang Pandas
* [gonum/gonum](https://github.com/gonum/gonum): Golang Numpy

### Main new files
* Stan interface implementation: `pkg/calculator/stan/stan.go`
* Stan interface implementation unit test: `pkg/calculator/stan/stan_test.go`
* Experiment calculator (Python Calculator Rewrite) : `pkg/calculator/experimentcalc/experiment_calculator.go`
* Experiment calculator unit test: `pkg/calculator/experimentcalc/experiment_calculator_test.go`
* Stan params probability statistics calculation: `pkg/calculator/experimentcalc/statistics.go`
* The `python/src/lib/calculator/stats/normal_inverse_gamma.py` Golang implementation: `pkg/calculator/experimentcalc/normal_inverse_gamma.go`
* Experiment result mysql storage: `pkg/calculator/storage/v2/experiment_result.go`

## `pr-go` workflow change
Due to the implementation of the Stan interface using HttpStan, I have added a HttpStan container service in `pr-go` workflow so we can use the API of HttpStan when we start unit testing.

## Code References

- Implementation Code

| Go Code | Python Code | Description |
|--------|--------|--------|
| [ExperimentCalculator.Run](https://github.com/Ubisoft-potato/bucketeer/blob/feature/refactor_calculator/pkg/calculator/experimentcalc/experiment_calculator.go#L76) | [ExperimentCalculator.run](https://github.com/bucketeer-io/bucketeer/blob/main/python/src/lib/calculator/job/calculate_experiments.py#L61) | This function is the entry of the experiment calculation |
| [ExperimentCalculator.createExperimentResult](https://github.com/Ubisoft-potato/bucketeer/blob/feature/refactor_calculator/pkg/calculator/experimentcalc/experiment_calculator.go#L125) | [ExperimentCalculator. _create_experiment_result](https://github.com/bucketeer-io/bucketeer/blob/main/python/src/lib/calculator/job/calculate_experiments.py#L87) | This function is used to calculate single experiment result (get the experiment's evaluation counter and goal counter from the event counter service then send to `Stan`) |
| [ExperimentCalculator.calcGoalResult](https://github.com/Ubisoft-potato/bucketeer/blob/feature/refactor_calculator/pkg/calculator/experimentcalc/experiment_calculator.go#L248) | [ExperimentCalculator._calc_goal_result](https://github.com/bucketeer-io/bucketeer/blob/main/python/src/lib/calculator/job/calculate_experiments.py#L417) | Prepare the data then send it to `Stan` and parse `Stan` calculation result  |
| [ExperimentCalculator.binomialModelSample](https://github.com/Ubisoft-potato/bucketeer/blob/feature/refactor_calculator/pkg/calculator/experimentcalc/experiment_calculator.go#LL404C36-L404C36) | [Binomial.run](https://github.com/bucketeer-io/bucketeer/blob/main/python/src/lib/calculator/stats/binomial.py#L75) | This is core function of experiment calculation: call `Stan` then calculate the statistics result using Numpy and Pandas based on the samplings |
| [ExperimentCalculator.convertFitSamples](https://github.com/Ubisoft-potato/bucketeer/blob/feature/refactor_calculator/pkg/calculator/experimentcalc/experiment_calculator.go#L589) | [Binomial._conv_fit](https://github.com/bucketeer-io/bucketeer/blob/main/python/src/lib/calculator/stats/binomial.py#LL113C15-L113C15)  | Calculate the statistics: `Mean` `Sd` `Rhat` `percentile025` `percentile975` |
| [normalInverseGamma](https://github.com/Ubisoft-potato/bucketeer/blob/feature/refactor_calculator/pkg/calculator/experimentcalc/normal_inverse_gamma.go#L47) | [normal_inverse_gamma.py](https://github.com/bucketeer-io/bucketeer/blob/main/python/src/lib/calculator/stats/normal_inverse_gamma.py#L37) | This function is used to calculate Normal Inverse Gamma posterior | 

- Tests Code

| Go Test | Python Test | Description |
|--------|--------|--------|
| [TestExperimentCalculatorBinomialModelSample](https://github.com/Ubisoft-potato/bucketeer/blob/feature/refactor_calculator/pkg/calculator/experimentcalc/experiment_calculator_test.go#L44) | [test_calc_result](https://github.com/bucketeer-io/bucketeer/blob/main/python/tests/lib/calculator/stats/binomial_test.py#L21) | This function is tested using `Stan` to do samping then calculate the statistics |
| [TestNormalInverseGamma](https://github.com/Ubisoft-potato/bucketeer/blob/feature/refactor_calculator/pkg/calculator/experimentcalc/normal_inverse_gamma_test.go#L27) | [normal_inverse_gamma_test](https://github.com/bucketeer-io/bucketeer/blob/main/python/tests/lib/calculator/stats/normal_inverse_gamma_test.py#L23) | This function is tested Normal Inverse Gamma posterior |


cre8ivejp:
@Ubisoft-potato, instead of adding another module for the experiment calculator API, I was thinking of adding this API as part of `experimentService` struct.

- Experiment calculator (Python Calculator Rewrite): pkg/experiment/api/calculator.go
- Experiment calculator unit test: pkg/experiment/api/calculator_test.go
- Stan interface implementation: pkg/experiment/calculator/stan.go
- Stan interface implementation unit test: pkg/experiment/calculator/stan_test.go

E.g.

```go
func (s *experimentService) CalculateExperiments(..) {
  listEnvironments...
  listExperiments...
  calculate..
}
```

@masaaania, please let me know your thoughts on this.

Ubisoft-potato:
> @Ubisoft-potato, instead of adding another module for the experiment calculator API, I was thinking of adding this API as part of `experimentService` struct.
> 
> * Experiment calculator (Python Calculator Rewrite): pkg/experiment/api/calculator.go
> * Experiment calculator unit test: pkg/experiment/api/calculator_test.go
> * Stan interface implementation: pkg/experiment/calculator/stan.go
> * Stan interface implementation unit test: pkg/experiment/calculator/stan_test.go
> 
> E.g.
> 
> ```go
> func (s *experimentService) CalculateExperiments(..) {
>   listEnvironments...
>   listExperiments...
>   calculate..
> }
> ```
> 
> @masaaania, please let me know your thoughts on this.

@cre8ivejp Yes, my thoughts was add this as an API that part of `experimentService` struct, also it wiil be added gRPC handler. But the Python Calculator service logic is massive, so I created a seperated Go pacakage `calculator` to implement the Python Calculator service logic; then we import this pacakage in `experimentService`.

cre8ivejp:
@Ubisoft-potato cc: @masaaania 

> Yes, my thoughts was add this as an API that part of experimentService struct, also it wiil be added gRPC handler. But the Python Calculator service logic is massive, so I created a seperated Go pacakage calculator to implement the Python Calculator service logic; then we import this pacakage in experimentService.

Indeed. I suggested that because maybe some functions in the `experimentService` struct could be reused and kept simple.
But if most code is new, your proposal sounds better.

Ubisoft-potato:
@cre8ivejp Yes, as you can see this pr will add many new functions that `experimentService` do not contain.

And except that, for example, we cannot reuse `experimentService.ListExperiments` directly, because `experimentService.ListExperiments` function is an gRPC handler that serves external requests, and it will check the role of the user before processing the request.

```golang
func (s *experimentService) ListExperiments(
	ctx context.Context,
	req *proto.ListExperimentsRequest,
) (*proto.ListExperimentsResponse, error) {
	localizer := locale.NewLocalizer(ctx)

	_, err := s.checkRole(ctx, accountproto.Account_VIEWER, req.EnvironmentNamespace, localizer)
	if err != nil {
		return nil, err
	}

	whereParts := []mysql.WherePart{
		mysql.NewFilter("deleted", "=", false),
		mysql.NewFilter("environment_namespace", "=", req.EnvironmentNamespace),
	}
        ...
```
But we don't need to check the role of the request because we will call `CalculateExperiments` using k8s cronjob internally, so I simplify `experimentService.ListExperiments` function in `calculator` package as an internal function.

pkg/experiment/calculator/stan.go
```diff
@@ -0,0 +1,318 @@
...
+	crateFitErrResp := CrateFitErrResp{}
+
+	resp, err := s.client.R().
+		SetPathParam("model_id", modelID).
+		SetBody(req).
+		SetResult(&createFitResp).
+		SetError(&crateFitErrResp).
+		Post("/v1/models/{model_id}/fits")
+	if err != nil {
+		s.logger.Info("HttpStan failed to create fit",
```
cre8ivejp, line 178: Please use `s.logger.Error` for errors.
Same for other places.

pkg/experiment/calculator/experiment_calculator.go
```diff
@@ -0,0 +1,270 @@
...
+	return variationCounts, nil
+}
+
+func (e ExperimentCalculator) calcGoalResult(
+	ctx context.Context,
+	evalVariationCounts,
+	goalVariationCounts map[string]*eventcounter.VariationCount,
+	baseVid string,
+) *eventcounter.GoalResult {
+	var (
```
cre8ivejp, line 229:
```suggestion
goalResult := &eventcounter.GoalResult{}
vids := make([]string, 0, len(goalVariationCounts))
```

pkg/experiment/calculator/experiment_calculator.go
```diff
@@ -0,0 +1,270 @@
...
+		goalResult = &eventcounter.GoalResult{}
+		vids       = make([]string, 0, len(goalVariationCounts))
+	)
+	for vid, _ := range goalVariationCounts {
+		vids = append(vids, vid)
```
cre8ivejp, line 234: The vids var is not being used. Is this function unfinished?

Ubisoft-potato: Yes, I combed through the logic and listed the functions that need to be implemented of the Python Calculator. So the function not be fully implemented yet.

pkg/experiment/calculator/experiment_calculator.go
```diff
@@ -0,0 +1,270 @@
...
+		}),
+	}
+	orders := []*mysql.Order{mysql.NewOrder("name", mysql.OrderDirectionAsc)}
+	experimentStorage := v2es.NewExperimentStorage(e.mysqlClient)
+
+	experiments, _, _, err := experimentStorage.ListExperiments(
+		ctx,
+		whereParts,
+		orders,
+		0,
```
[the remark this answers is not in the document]
Ubisoft-potato: OK

pkg/calculator/experimentcalc/experiment_calculator.go
```diff
@@ -0,0 +1,613 @@
...
+						log.FieldsFromImcomingContext(ctx).AddFields(
+							zap.String("fitId", fitId),
+							zap.Error(err),
+						)...,
+					)
+				}
+				if details.Done {
+					break
+				}
+				time.Sleep(50 * time.Millisecond)
```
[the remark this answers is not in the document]
Ubisoft-potato: I found that `pystan` was [using `50ms` to check result from `httpstan`](https://github.com/stan-dev/pystan/blob/main/stan/model.py#L222).
Actually when I am testing, this will return in around 1 seconds.

pkg/calculator/experimentcalc/experiment_calculator.go
```diff
@@ -0,0 +1,613 @@
...
+	baseLineIdx++
+	samplesChan := make(chan dataframe.DataFrame)
+	wg := sync.WaitGroup{}
+	wg.Add(numOfChains)
+	for i := 1; i <= numOfChains; i++ {
+		go func(chain int) {
+			defer wg.Done()
+			compiledModel, err := e.stan.CompileModel(ctx, stan.ModelCode())
+			if err != nil {
+				e.logger.Error("failed to compile model",
```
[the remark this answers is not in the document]
Ubisoft-potato: No problem! I will keep this consistency.

pkg/calculator/experimentcalc/experiment_calculator.go
```diff
@@ -0,0 +1,613 @@
...
+				if err != nil {
+					e.logger.Error("failed to get operation details",
+						log.FieldsFromImcomingContext(ctx).AddFields(
+							zap.String("fitId", fitId),
+							zap.Error(err),
+						)...,
+					)
+				}
+				if details.Done {
+					break
```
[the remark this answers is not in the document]
Ubisoft-potato: Yes, this need to add check logic to break loop.

> samples := make([]dataframe.DataFrame, 0, numOfChains)

And we will check the length of `samples` slice, if the length is not equal `numOfChains`, so we can consider this fit was failed.

pkg/calculator/experimentcalc/experiment_calculator.go
```diff
@@ -0,0 +1,650 @@
...
+					zap.String("namespace", env.Namespace),
+					zap.Error(experimentErr),
+				)...,
+			)
+			return
+		}
+		for _, ex := range experiments {
+			if ex.StartAt > now.Unix() {
+				// already calculated
+				continue
```
masaaania, line 125: I think you skip it because the experiment has been started yet, not because it has been already calculated.

pkg/calculator/experimentcalc/experiment_calculator.go
```diff
@@ -0,0 +1,650 @@
...
+	length := len(goalVariationCounts)
+	vids := make([]string, 0, length)
+	goalUc, evalUc := make([]int64, 0, length), make([]int64, 0, length)
+	vrs := make(map[string]*eventcounter.VariationResult, length)
+	valueMeans, valueVars := make([]float64, 0, length), make([]float64, 0, length)
+	baselineIdx, loopIdx := 0, 0
+
+	for vid, goalVariationCount := range goalVariationCounts {
+		if _, ok := evalVariationCounts[vid]; !ok {
+			return nil
```
masaaania, line 286: The return value of this function is to be accessed in [the next line](https://github.com/bucketeer-io/bucketeer/pull/395/files#diff-6b2e151d6e7453bb8a10c3b450be4ba51fc681fe35b1b485d43d452d0deb19b5R195-R196), so returning nil would result in a nil pointer exception.

Ubisoft-potato: Of cause. Thanks for pointing this out!

pkg/calculator/experimentcalc/experiment_calculator.go
Ubisoft-potato: When wrote this I was also confuse about the logic in Python Calculator, so I just keep it the same, maybe the orginal logic is not right :) :
[the logic in `run` function](https://github.com/bucketeer-io/bucketeer/blob/main/python/src/lib/calculator/job/calculate_experiments.py#L78)

Jump in:
[is_calculated](https://github.com/bucketeer-io/bucketeer/blob/main/python/src/lib/calculator/domain/experiment.py#L41)

masaaania: Right, I think so too.
I also confirmed with @cre8ivejp. 
Please fix it 🙏

Ubisoft-potato: Thanks for your confirmation!

So this will be changed to:
```go
for _, ex := range experiments {
			if ex.StartAt <= now.Unix() {
				// experiment not started yet 
				continue
```

pkg/calculator/experimentcalc/experiment_calculator.go
Ubisoft-potato: Fixed

pkg/calculator/experimentcalc/experiment_calculator.go
Ubisoft-potato: Fixed.

pkg/calculator/experimentcalc/experiment_calculator.go
```diff
@@ -0,0 +1,650 @@
...
+	}
+	for vid, vr := range cvrResult {
+		vrs[vid].CvrProb = copyDistributionSummary(vr.CvrProb)
+		vrs[vid].CvrProbBest = copyDistributionSummary(vr.CvrProbBest)
+		vrs[vid].CvrProbBeatBaseline = copyDistributionSummary(vr.CvrProbBeatBaseline)
+	}
+	// Skip the calculation if values are zero.
+	for i := 0; i < len(vids); i++ {
+		if goalUc[i] == 0 || valueMeans[i] == 0 || valueVars[i] == 0 {
+			return goalResult
```
cre8ivejp, line 333: Let's add a new metric to know that the calculation was skipped.

```go
skippedValuesAreZero
```

pkg/calculator/experimentcalc/experiment_calculator.go
```diff
@@ -0,0 +1,650 @@
...
+		vrs[vid] = vr
+		if vid == baseVid {
+			baselineIdx = loopIdx
+		}
+		loopIdx++
+	}
+	// Skip the calculation if evaluation count is less than goal count.
+	for i := 0; i < len(evalUc); i++ {
+		if evalUc[i] < goalUc[i] {
+			return goalResult
```
cre8ivejp, line 310: Let's add a new metric to know that the calculation was skipped.

```go
skippedEvaluationCountLessThanGoalEvent
```

Also, if we skip the calculation, should it return `&eventcounter.GoalResult{}` instead?
Because at [this line](https://github.com/bucketeer-io/bucketeer/pull/395/files#diff-6b2e151d6e7453bb8a10c3b450be4ba51fc681fe35b1b485d43d452d0deb19b5R300) it is happening results.

pkg/calculator/experimentcalc/experiment_calculator.go
```diff
@@ -0,0 +1,650 @@
...
+	goalResult := &eventcounter.GoalResult{}
+	length := len(goalVariationCounts)
+	vids := make([]string, 0, length)
+	goalUc, evalUc := make([]int64, 0, length), make([]int64, 0, length)
+	vrs := make(map[string]*eventcounter.VariationResult, length)
+	valueMeans, valueVars := make([]float64, 0, length), make([]float64, 0, length)
+	baselineIdx, loopIdx := 0, 0
+
+	for vid, goalVariationCount := range goalVariationCounts {
+		if _, ok := evalVariationCounts[vid]; !ok {
```
cre8ivejp, line 285: I think, at this point, the variation count should be on the map. If not, it should be treated as an error.
What do you think?

Either way, we should add a metric here so we can monitor those cases.

E.g.

```go
evalVariationCountNotFound
```

pkg/calculator/stan/stan.go
```diff
@@ -0,0 +1,376 @@
...
+	T0             int                    `json:"t0,omitempty"`
+	TermBuffer     int                    `json:"term_buffer,omitempty"`
+	Window         int                    `json:"window,omitempty"`
+}
+
+type CreateFitResp struct {
+	Name string `json:"name"`
+}
+
+type CrateFitErrResp struct {
```
cre8ivejp, line 87: nit: `CreateFitErrResp`

pkg/calculator/stan/stan.go
```diff
@@ -0,0 +1,376 @@
...
+// CreateFit Post request /v1/models/{model_id}/fits to start a long-running fit operation
+//
+// Status Codes:
+//
+//	201 Created – Identifier for completed Stan fit
+//	400 Bad Request – Error associated with request.
+//	404 Not Found – Fit not found.
+func (s Stan) CreateFit(ctx context.Context, modelID string, req CreateFitReq) (string, error) {
+	createFitResp := CreateFitResp{}
+	crateFitErrResp := CrateFitErrResp{}
```
cre8ivejp, line 190: nit: `createFitErrResp`

pkg/calculator/experimentcalc/experiment_calculator.go
```diff
@@ -0,0 +1,650 @@
...
+	cvrResult, sampleErr := e.binomialModelSample(ctx, vids, goalUc, evalUc, baselineIdx)
+	if sampleErr != nil {
+		calculationCounter.WithLabelValues(calculateFail).Inc()
+		e.logger.Error("BinomialModelSample error",
+			log.FieldsFromImcomingContext(ctx).AddFields(
+				zap.Error(sampleErr),
+				zap.String("baseVariationID", baseVid),
+			)...,
+		)
+		return goalResult
```
cre8ivejp, line 323: Same here.
https://github.com/bucketeer-io/bucketeer/pull/395#discussion_r1236424262

pkg/calculator/experimentcalc/metrics.go
```diff
@@ -0,0 +1,55 @@
...
+package experimentcalc
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+
+	"github.com/bucketeer-io/bucketeer/pkg/metrics"
+)
+
+const (
+	calculateFail    = "Fail"
```
cre8ivejp, line 25: nit: `calculationFail` & `calculationSuccess`

pkg/calculator/experimentcalc/experiment_calculator.go
Ubisoft-potato: I think so too. But I was not sure how to optimize this when developing and just keep this logic the same as Python code, so we can discuss about this here.

Anyway, like you said, we should ad a metric here to let us know when this case happens.

pkg/calculator/experimentcalc/experiment_calculator.go
Ubisoft-potato: > Also, if we skip the calculation, should it return `&eventcounter.GoalResult{}` instead? Because at [this line](https://github.com/bucketeer-io/bucketeer/pull/395/files#diff-6b2e151d6e7453bb8a10c3b450be4ba51fc681fe35b1b485d43d452d0deb19b5R300) it is happening results.

The original Python code logic was:
 [line 1](https://github.com/bucketeer-io/bucketeer/blob/main/python/src/lib/calculator/job/calculate_experiments.py#L459)
[line 2](https://github.com/bucketeer-io/bucketeer/blob/main/python/src/lib/calculator/job/calculate_experiments.py#L434)

I think we can return the empty `GoalResult{}` here, but also you can confirm this logic with the original writer.

Anyway, we should add this metric first.

pkg/calculator/experimentcalc/experiment_calculator.go
Ubisoft-potato: The original Python code logic [here](https://github.com/bucketeer-io/bucketeer/blob/main/python/src/lib/calculator/job/calculate_experiments.py#L472)

Same as  https://github.com/bucketeer-io/bucketeer/pull/395#discussion_r1236570075

pkg/calculator/experimentcalc/experiment_calculator.go
cre8ivejp: Next week the person who wrote the Python code will take some time to review this part.
So we can confirm with him the best approach for those cases. For now, let's just add the metrics.

kentakozuka: IMHO, the return values should be `goalResult` not empty struct because `goalResult` must contain the count values until here. The cause of this error can be a network error, client error, system error, or others, ending up the evaluation count is less than the goal count. This does not supposed to happen usually, but it can happen. I think we should return the real values to end-users rather than returning nothing. wdyt?

pkg/calculator/experimentcalc/experiment_calculator.go
```diff
@@ -0,0 +1,673 @@
...
+	var timestamps []int64
+	if endAt >= now {
+		endAt = now
+	}
+	for ts := startAt + day; ts < endAt; ts += day {
+		timestamps = append(timestamps, ts)
+	}
+	timestamps = append(timestamps, endAt)
+	return timestamps
+}
```
kentakozuka, line 614: I don't know if this function works well with UI.(I wanna ask myself back in time)
If the server runs in UCT and a user uses a browser in GMT, for example, how does the UI display the counts?
The server side here looks good to me, though the UI should be checked.

kentakozuka:
@Ubisoft-potato Thank you so much for doing this!
@cre8ivejp I checked the overall flow of the code, and it looks good to me. I left comments. I could not check all in detail so let me know the points that you want me to look at.

pkg/calculator/experimentcalc/experiment_calculator.go
Ubisoft-potato: I think we should use GMT timezone for all experiments in MySQL, the UI can show the real time according to the user's browser.

pkg/calculator/experimentcalc/experiment_calculator.go
cre8ivejp: @kentakozuka, I am concerned about returning the unfinished result, affecting the visualization on the UI, and introducing bugs that are hard to debug later.

Since the calculator will calculate the probability again in the next job, I thought it would be better to treat it as an error and not update the experiment result. Also, as you said, this usually won't happen, so implementing it not to update the experiment result sounds safer.

Let me know if I'm missing something.

pkg/calculator/experimentcalc/experiment_calculator.go
cre8ivejp: Reference.
https://github.com/bucketeer-io/bucketeer/pull/395#discussion_r1243356931

pkg/calculator/experimentcalc/experiment_calculator.go
cre8ivejp: @Ubisoft-potato 
For new services, we are adding a new config called `timezone` in the helm chart so that the user can set his timezone.
Can you implement the timezone flag in the calculator service?

Reference
https://github.com/bucketeer-io/bucketeer/blob/main/pkg/eventcounter/cmd/server/server.go#L114

https://github.com/bucketeer-io/bucketeer/blob/main/pkg/eventcounter/api/api.go#L624

.github/workflows/pr-python.yaml
```diff
@@ -47,3 +48,34 @@ jobs:
           make gen
           make test
 
+  proto-check:
```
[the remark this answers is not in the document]
Ubisoft-potato: OK

pkg/calculator/experimentcalc/experiment_calculator.go
Ubisoft-potato: >For new services, we are adding a new config called timezone in the helm chart so that the user can set his timezone.
>Can you implement the timezone flag in the calculator service?

@cre8ivejp No problem! Using timezone is a good solution for this.

Ubisoft-potato: > @Ubisoft-potato For new services, we are adding a new config called `timezone` in the helm chart so that the user can set his timezone. Can you implement the timezone flag in the calculator service?

Done! And Helm Chart PR also added timezone env.

Reference:
https://github.com/bucketeer-io/bucketeer/pull/420

pkg/calculator/experimentcalc/experiment_calculator.go
kentakozuka: I think that we should define what we treat as error and what not.
It can come to this line even if the Bucketeer works as we expected (and not sure it would not come again in the next job. In this line we can still return;
- evalUc
- goalUc
- valueMeans
- valueVars

Also, would it be better to return either?;
- `goalResult`, nil
- `nil, err`

cre8ivejp: @kentakozuka, the appended data [here](https://github.com/bucketeer-io/bucketeer/blob/e6180757e36ee6bb6558ed5d9ebea433b7dabed8/pkg/calculator/experimentcalc/experiment_calculator.go#L309) is the data display here?

If so, it should be ok to return the `goalResult`. 
If the goal count is higher than the evaluation count, the UI will only display the graph below, right?

Screenshot 2023-06-28 at 14 47 54

pkg/calculator/experimentcalc/experiment_calculator.go
```diff
@@ -0,0 +1,676 @@
...
+				)...,
+			)
+			return
+		}
+		for _, ex := range experiments {
+			if ex.StartAt <= now.Unix() {
+				// experiment not started yet
+				continue
+			}
+			experimentResult := e.createExperimentResult(ctx, env.Namespace, ex)
```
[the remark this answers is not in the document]
Ubisoft-potato: Fixed.

pkg/calculator/experimentcalc/experiment_calculator.go
kentakozuka: > he appended data here is the data displayed in the screenshot below?

I'm not sure but possibly yes.
https://github.com/bucketeer-io/bucketeer/blob/b3a0abc1385f9ec79d954aa4b7d8a7d11085a9c7/ui/web-v2/apps/admin/src/components/GoalResultDetail/index.tsx#L137

pkg/experimentcalculator/experimentcalc/experiment_calculator.go
```diff
@@ -129,6 +129,11 @@ func (e ExperimentCalculator) Run(ctx context.Context, request *calculator.Batch
 				// experiment not started yet
 				continue
 			}
+			if ex.Status == experiment.Experiment_STOPPED &&
+				now.Unix()-ex.StopAt > 2*day {
+				// experiment stopped and 2 days passed
```
cre8ivejp, line 134: Please add the following comment.

```
Because the evaluation and goal events may be sent with a delay for many reasons from the client side, we still calculate the results for two days after it stopped.
```

Ubisoft-potato: Done!

pkg/experimentcalculator/experimentcalc/experiment_calculator.go
```diff
@@ -0,0 +1,660 @@
...
+	ctx context.Context,
+	namespace string,
+) ([]*experiment.Experiment, error) {
+	req := &experiment.ListExperimentsRequest{
+		From:                 time.Now().In(e.location).Add(-2 * 24 * time.Hour).Unix(),
+		PageSize:             0,
+		Cursor:               "",
+		EnvironmentNamespace: namespace,
+		Statuses: []experiment.Experiment_Status{
+			experiment.Experiment_WAITING,
```
cre8ivejp, line 263: @Ubisoft-potato, I just noticed that because we will start the experiment via batch service, we no longer need to get the experiment in WAITING status.

Also, we can remove this condition.
https://github.com/bucketeer-io/bucketeer/pull/395/files#diff-b0f1a6e9d024e46f089e3bc6573a1c8f6dfaf821d6b3898e3c06794bbc4d0dd8R128-R131

Ubisoft-potato: Done!

cre8ivejp (approved):
@Ubisoft-potato, thank you for the amazing work!!

PR #397, opened 2023-06-06
chore: add httpstan container to experiment service pod

Ubisoft-potato:
## What I am doing?
* This pull request is part of imlementation of this [issue](https://github.com/bucketeer-io/bucketeer/issues/356).
* Implement HttpStan as a container in experiment service pod.

## Summary
This PR will add a container to the experiment service, so the architecture will look like the following:

architecture

### DIFF
--- a/manifests/bucketeer/charts/experiment/templates/deployment.yaml
+++ b/manifests/bucketeer/charts/experiment/templates/deployment.yaml
@@ -164,5 +164,26 @@ spec:
               scheme: HTTPS
           resources:
 {{ toYaml .Values.envoy.resources | indent 12 }}
+        - name: httpstan
+          image: "{{ .Values.httpstan.image.repository }}:{{ .Values.httpstan.image.tag }}"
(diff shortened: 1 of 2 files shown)


cre8ivejp:
@Ubisoft-potato, can you create another chart only for httpstan instead of adding it as a side-car?

The main reason is scalability. In some cases, the experiment service gets a lot of requests, but the calculator is not computing anything. So, when it scales out, it would also scale the httpstan wasting resources. Or vice-versa. The httpstan needs to scale out, and the experiment doesn't.

Ubisoft-potato:
> @Ubisoft-potato, can you create another chart only for httpstan instead of adding it as a side-car?
> 
> The main reason is scalability. In some cases, the experiment service gets a lot of requests, but the calculator is not computing anything. So, when it scales out, it would also scale the httpstan wasting resources. Or vice-versa. The httpstan needs to scale out, and the experiment doesn't.

Ok, This will look like what I have suggested during our conversation. I will create a new chart for httpstan so we can scale httpstan as needed.

PR #398, opened 2023-06-07
chore: add generated files from proto

ono-max:
As a best practice, we should commit generated files from proto.

Ref: https://go.dev/doc/articles/go_command


cre8ivejp:
We should also include the descriptor, python, and javascript generated proto files

```
.gitignore files

# /
/proto/**/proto_descriptor.pb
/proto/**/*.py

# /python
proto

# /ui/web-v2
**/proto/**/*.js
**/proto/**/*.ts
```

Also, because we are pushing the generated files, this `run: make proto-go` step in all the github action workflows can be removed.

The only step that we need to run `make proto-go` should be [here](https://github.com/bucketeer-io/bucketeer/blob/main/.github/workflows/pr-go.yaml#L81-L112).

@masaaania, can you confirm?

masaaania:
> The only step that we need to run make proto-go should be [here](https://github.com/bucketeer-io/bucketeer/blob/main/.github/workflows/pr-go.yaml#L81-L112).

I have the same opinion.

In addition to that, we should add a step to check the update of pb file to the pr-python and pr-ui-web workflows.

https://github.com/bucketeer-io/bucketeer/blob/9c7cb3461ee7f580b6cc6d42ed939fef8d2563be/.github/workflows/pr-python.yaml#L7-L9

https://github.com/bucketeer-io/bucketeer/blob/9c7cb3461ee7f580b6cc6d42ed939fef8d2563be/.github/workflows/pr-ui-web.yaml#L7-L9

PR #399, opened 2023-06-07
perf: generate the UEID from the features filtered by the tag

masaaania:
## problem
Currently, buckteer generates the UserEvaluationsID from all active(not archived) features.
This method results in unnecessary evaluations when features with other tags are updated.

e.g., Even when only the feature with the `ios` tag is updated, the features with other tags such as `web` or `android` will be evaluated.

## This PR
- This PR modifies the method of generating the UEID to generate them from the features filtered by the specified tag.


ono-max (approved):
👍

cre8ivejp (approved):
Thank you!

PR #272, opened 2023-04-03
feat(ui): show which flag the segment user is using when deleting or updating it

bimalgrg519:
Fix https://github.com/bucketeer-io/bucketeer/issues/276

PR #402, opened 2023-06-09
chore: add features field to list segment api

ono-max:
Currently, Bucketeer shows the segment user as follows:
Screen Shot 2023-06-09 at 14 11 58

If the segment user is used, it shows "In Use". If not, "Not In Use". Although it's useful, it would be better to show which feature uses this segment user. I implement this part of the backend.


pkg/feature/storage/v2/segment.go
```diff
@@ -256,32 +257,31 @@ func (s *segmentStorage) ListSegments(
...
-					FROM 
-						feature
-					WHERE
-						environment_namespace = ? AND
-						rules LIKE concat("%%", segment.id, "%%")
-				) > 0 THEN TRUE 
-				ELSE FALSE
-			END AS is_in_use_status
+			(
+				SELECT
```
cre8ivejp, line 261: Please also implement this logic in the `GetSegment` interface.

pkg/feature/api/segment.go
```diff
@@ -568,3 +622,62 @@ func (s *FeatureService) newSegmentListOrders(
...
 	return []*mysql.Order{mysql.NewOrder(column, direction)}, nil
 }
+
+func (s *FeatureService) injectFeaturesIntoSegments(
+	ctx context.Context,
+	segments []*featureproto.Segment,
+	featureIDsMap map[string][]string,
+	environmentNameSpace string,
+) error {
+	allFeatures, err := s.listAllFeatures(
```
cre8ivejp, line 632: For the `ListSegments` API, getting all the feature flags in advance makes sense, but for the `GetSegment` API would be better to use the `GetFeatures` API to get only the target flags. But it would need two implementations, so I think we can go as it is, and if we see performance issues, we can refactor this part.

cre8ivejp (dismissed):
Thanks!

cre8ivejp (approved):
👍

cre8ivejp:
Part of https://github.com/bucketeer-io/bucketeer/pull/272

masaaania (approved):
lgtm!

ono-max:
Sorry, conflicts occurred again. I just resolved it.

cre8ivejp:
@ono-max, can you take a look at the CI?

cre8ivejp (approved):
nit: `chore: add features field to list segment api`

ono-max:
Sorry to bother you again. I mistakenly commit the change. I reverted it.

cre8ivejp (approved):
Thanks!

masaaania (approved):

PR #404, opened 2023-06-12
chore: add backend service

masaaania:
Issue: https://github.com/bucketeer-io/bucketeer/issues/405

- This PR adds a new service named `backend,` which contains the functions of `auth` and `account.`
  - The `backend` service is supposed to implement other functions such as the `environment,` `feature,` and so on. 
  - Even if we merge this PR, no requests will be sent to the `backend` service. I will change it with the next PR.


cre8ivejp (approved):
Thank you!
Amazing work!

PR #414, opened 2023-06-14
chore: add gen-check command

ono-max:


.github/workflows/pr-python.yaml
```diff
@@ -47,3 +47,34 @@ jobs:
           make gen
           make test
```
masaaania, line 49: We should add the path `proto/**` to `on.pull_request.paths.`

masaaania (changes requested):

ono-max:
Rebased

.github/workflows/pr-python.yaml
ono-max: [d931295](https://github.com/bucketeer-io/bucketeer/pull/414/commits/d93129573ae4d9f1a10ddccc8eff56d45e38b81d)

cre8ivejp (approved):
👍

masaaania (approved):
LGTM!

PR #416, opened 2023-06-14
chore: add routing from other microservices to backend service

masaaania:
- This PR adds routing settings to the backend service from other microservices.
- After this PR is merged, requests will no longer be sent to the `auth` and `account` services.


cre8ivejp (approved):
Looks good!
Nice work!

PR #419, opened 2023-06-20
chore: add audit-log, auto-ops, environment service into the backend module

masaaania:
- This PR integrates the audit-log, the auto-ops, and the environment service into the backend module.
- This PR adds routing settings to the backend service from other microservices.
- After this PR is merged, requests will no longer be sent to the audit-log, auto-ops, and environment services.


cre8ivejp (approved):
Thank you!

PR #424, opened 2023-06-23
chore: add proto-all command in Makefile

ono-max:
Currently, generated proto files are checked in CI. If we update proto files, we must run all proto commands in each directory. Thus, it would be useful if we update all generated proto files in one command. I introduced proto-all command in this PR.


cre8ivejp (approved):
It is helpful!
Thanks!

masaaania (approved):
LGTM!

PR #426, opened 2023-06-26
chore: add migration-mysql, notification, push  service into the backend  module

masaaania:
- This PR integrates the migration-mysql, notification and push service into the backend module.
- Even if this PR is merged, requests will be sent to the migration-mysql, notification and push services from the web-gateway.


cre8ivejp (approved):
Thanks!

PR #429, opened 2023-06-27
chore: add routing to migration-mysql, notification, push in the backend module

masaaania:
- This PR adds the route to the migration-mysql, notification, and push services in the backend module.
- Once this PR is merged, requests to the migration-mysql, notification, and push module will no longer be sent.


cre8ivejp (approved):
Thanks!

PR #430, opened 2023-06-27
chore: remove unused services from helm chart

masaaania:
- This PR removes the services integrated into the backend service already.


cre8ivejp (approved):
🎉

PR #432, opened 2023-06-28
chore: adjust shutdown process

masaaania:
- This PR adjusts the process of a graceful shutdown.

### DIFF
--- a/pkg/backend/cmd/server/server.go
+++ b/pkg/backend/cmd/server/server.go
@@ -63,10 +63,13 @@ import (
 )
 
 const (
-	command            = "server"
-	gcp                = "gcp"
-	aws                = "aws"
-	autoOpsWebhookPath = "hook"
+	command               = "server"
+	gcp                   = "gcp"
+	aws                   = "aws"
+	autoOpsWebhookPath    = "hook"
+	healthCheckTimeout    = 1 * time.Second
+	clientDialTimeout     = 30 * time.Second
+	serverShutDownTimeout = 10 * time.Second
 )
 
 type server struct {
@@ -306,7 +309,7 @@ func (s *server) Run(ctx context.Context, metrics metrics.Metrics, logger *zap.L
 	// healthCheckService
 	restHealthChecker := health.NewRestChecker(
 		"", "",
-		health.WithTimeout(time.Second),
+		health.WithTimeout(healthCheckTimeout),
 		health.WithCheck("metrics", metrics.Check),
 	)
 	go restHealthChecker.Run(ctx)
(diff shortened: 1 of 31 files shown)


pkg/rpc/server.go
```diff
@@ -121,18 +123,12 @@ func (s *Server) Run() {
...
 
 func (s *Server) Stop(timeout time.Duration) {
-	s.logger.Info("Server is going to sleep 10 seconds before shutting down")
-	// When the sigterm signal is sent, sometimes the app could get the signal before envoy,
-	// when it does, the requests will fail because the app cannot receive any request after the shutdown.
-	// So we wait a bit in case there are still requests to be processed
-	// between the envoy and app after the signal.
-	time.Sleep(10 * time.Second)
-	s.logger.Info("Server is awakening from sleep, and going to shutdown")
+	s.logger.Info("Server is going to shutdown")
```
cre8ivejp, line 126: Can you add more logs to check how long the shutdown process takes and if the container didn't get killed during the shutdown process?

```suggestion
	s.logger.Info("Server is going to shut down")
	ctx, cancel := context.WithTimeout(context.Background(), timeout)
	defer cancel()
	err := s.httpServer.Shutdown(ctx)
	if err != nil {
		s.logger.Error("Server failed to shut down," zap.Error(err))
	}
	s.logger.Info("Server has shut down gracefully")
```

Or we could add the elapsed time in the log after the shutdown, making it easier to read the logs.
Whatever you decide works great for me.

pkg/rest/server.go
```diff
@@ -98,13 +98,13 @@ func (s *Server) Run() {
 }
 
 func (s *Server) Stop(timeout time.Duration) {
-	s.logger.Info("Server is going to sleep 10 seconds before shutting down")
+	s.logger.Info("Health check server is going to sleep 1 seconds before shutting down")
```
cre8ivejp, line 101: The rest server is a common package that can be used for any service. So by using the `Health check server` in the log, it seems this package is for health checking only. It could confuse us.

Maybe `Rest server is going to sleep 1 second before shutting down` would be better.

pkg/rest/server.go
```diff
@@ -98,13 +98,13 @@ func (s *Server) Run() {
...
 	// When the sigterm signal is sent, sometimes the app could get the signal before envoy,
 	// when it does, the requests will fail because the app cannot receive any request after the shutdown.
 	// So we wait a bit in case there are still requests to be processed
 	// between the envoy and app after the signal.
 	time.Sleep(time.Second)
```
cre8ivejp, line 106: I think it is safe to remove this as you did for the Stop in the app.
It will also speed up the failing process in the Kubernetes.

pkg/backend/cmd/server/server.go
```diff
@@ -628,14 +616,42 @@ func (s *server) Run(ctx context.Context, metrics metrics.Metrics, logger *zap.L
...
 	)
-	defer pushServer.Stop(10 * time.Second)
 	go pushServer.Run()
-	// other services...
+	// To detach this pod from Kubernetes Service before the app servers stop, we stop the health check service first.
+	// Then, after 10 seconds of sleep, the application server can be shut down, as no new requests are expected to be sent.
+	// In this case, the Readiness prove must fail within 10 seconds and the pod must be detached.
+	defer func() {
+		go healthcheckServer.Stop(10 * time.Second)
+		time.Sleep(10 * time.Second)
```
cre8ivejp, line 631: Since the timeout duration is the same for the other places, how about adding a constant for the timeout?

cre8ivejp (approved):
Awesome!!!

PR #438, opened 2023-06-30
chore: add variation name to get evaluations response

cre8ivejp:
I'm adding the variation name to make it easier to analyze the user evaluations outside of Bucketeer because the id is UUID, and it s hard to know which variation is which.

### Things done

- Added variation name to the proto message
- Improved unit tests to get evaluations

### Tested on the following SDKs

- Android 1.x and 2.x
- iOS 1.x and 2.x
- JavaScript 1.x and 2.x


masaaania (approved):
Looks Good To Me!

PR #451, opened 2023-07-03
docs: rfc for db migration

masaaania:
Part of https://github.com/bucketeer-io/bucketeer/issues/685


docs/rfcs/0043-db-migration-for-selfhost-users
```diff
@@ -0,0 +1,86 @@
...
+                args:
+                - -c
+                - |
+                  echo "START automatic DB migration."
+                  ENDPOINT="http://backend:9100/bucketeer.migration.MigrationMySQLService/MigrateAllMasterSchema
+                  echo ${ENDPOINT}
+                  RES=`curl -X POST -m 300 ${ENDPOINT} -H "authorization: bearer $token" -H "Content-Type: application/json" -s -o /dev/null -w '%{http_code}\\n'`
+                  echo "response = "${RES}
+                  echo "END automatic DB migration."
+```
```
masaaania, line 73: This manifest is incomplete. (For example, It doesn't have a notification function.)
I will modify it during my work.

kentakozuka:
The migration strategy is updated. This RFC is old now. See [here](https://github.com/bucketeer-io/bucketeer/issues/685).

PR #454, opened 2023-07-05
chore: precompile stan model

Ubisoft-potato:
## Summary
Golang calculator service(https://github.com/bucketeer-io/bucketeer/pull/395) need to call httpstan to compile the stan model code first. And it could take long time to finish compile progress.

So this PR will precompile the stan model that we use in Golang calculator service and save the precompiled model files in the HttpStan Docker image directly. Then we can skip compile model in the Golang calculator service pod to speed up service start.

### How to make it
* Implement a precompile model python script, the script will start a httpstan server "locally" then call the httpstan's compile model http endpoint.
* After httpstan finished to compile model, it will save the compiled model in the following directory:
 `/root/.cache/httpstan/${httpstan_version}/models/${model_id}`.
  NOTE: the `model_id` was calculated by using the stan model code string hash automaticly.
* Copy the script into the Docker builder then execute it, then we can get an image that contains compiled model.

### DIFF
--- a/tools/httpstan/Dockerfile
+++ b/tools/httpstan/Dockerfile
@@ -24,7 +24,7 @@ RUN python3 -m poetry build
 RUN python3 -m pip install dist/*.whl
 
 # Pre Compile the model
-copy ./model.stan ./model.stan
+COPY ./model.stan ./model.stan
 COPY ./stan_model_compile.py ./stan_model_compile.py
 RUN python3 stan_model_compile.py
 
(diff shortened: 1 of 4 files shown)


tools/httpstan/Dockerfile
```diff
@@ -23,6 +23,11 @@ RUN python3 -m poetry build
 # Install the wheel
 RUN python3 -m pip install dist/*.whl
 
+# Pre Compile the model
+copy ./model.stan ./model.stan
```
masaaania, line 27: question:
Is there a reason why you use lowercase?

masaaania (approved):
Thank you!
LGTM 👍

cre8ivejp (approved):
Thank you!

PR #459, opened 2023-07-10
chore: update envoy to v1.26.2

cre8ivejp:
Because the `envoyproxy/envoy-alpine` image is not maintained anymore, we need to move the `envoyproxy/envoy` or `envoyproxy/envoy-distroless`.

The Envoy image doesn't have wget or curl installed by default, so we need to create a custom image with it so we can execute the scripts in the `preStop` hook.

### Things done

- Set explicitly the typed_config including the version for missing filters
- Added a Makefile to build and push the custom envoy image


masaaania (approved):
LGTM! Thank you!

PR #460, opened 2023-07-10
chore: change list experiments condition in the event persister

cre8ivejp:
Because the evaluation and goal events may be sent with a delay for many reasons from the client side,
We should still calculate the results after the experiment stops.

E2E tests have passed.

### DIFF
--- a/pkg/eventpersisterdwh/persister/evaluation_event.go
+++ b/pkg/eventpersisterdwh/persister/evaluation_event.go
@@ -38,6 +38,7 @@ type evalEvtWriter struct {
 	writer           storage.EvalEventWriter
 	experimentClient ec.Client
 	flightgroup      singleflight.Group
+	location         *time.Location
 	logger           *zap.Logger
 }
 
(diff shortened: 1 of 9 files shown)


pkg/eventpersisterdwh/persister/evaluation_event.go
```diff
@@ -220,5 +224,16 @@ func (w *evalEvtWriter) listExperiments(
...
 		return nil, err
 	}
-	return exp.([]*exproto.Experiment), nil
+	// Filter the stopped experiments
+	// Because the evaluation and goal events may be sent with a delay for many reasons from the client side,
+	// we still calculate the results for two days after it stopped.
+	now := time.Now().In(w.location)
+	experiments := make([]*exproto.Experiment, 0, len(exp.([]*exproto.Experiment)))
+	for _, e := range exp.([]*exproto.Experiment) {
+		if e.Status == exproto.Experiment_STOPPED && now.Unix()-e.StopAt > 2*day {
```
cre8ivejp, line 233: I implemented the same logic used in the experiment calculator.

https://github.com/bucketeer-io/bucketeer/pull/395/files#diff-b0f1a6e9d024e46f089e3bc6573a1c8f6dfaf821d6b3898e3c06794bbc4d0dd8R128-R133

masaaania (approved):
LGTM! Thank you!

PR #461, opened 2023-07-12
fix: change GetEvaluations request field

masaaania:
- This PR adds the `user_evaluation_condition` field to GetEvaluationsRequest to determine which evaluation method should be used.
  - If the `user_evaluation_condition` field is not set to any value, Bucketeer evaluates the flags with the old evaluation method.


proto/gateway/service.proto
```diff
@@ -35,8 +35,8 @@ message GetEvaluationsRequest {
   string user_evaluations_id = 3;
   string feature_id = 4 [deprecated = true];  // instead, use GetEvaluation API
   bucketeer.event.client.SourceId source_id = 5;
-  int64 evaluated_at = 6;
-  bool user_attributes_updated = 7;
+  feature.UserEvaluationCondition user_evaluation_condition = 6;
+  string sdk_version = 7;
```
masaaania, line 39: Any SDK does not use the `evaluated_at` and `user_attributes_updated`.

proto/gateway/service.proto
```diff
@@ -30,13 +30,17 @@ message PingResponse {
...
+    bool user_attributes_updated = 2;
+  }
   string tag = 1;
   user.User user = 2;
   string user_evaluations_id = 3;
   string feature_id = 4 [deprecated = true];  // instead, use GetEvaluation API
   bucketeer.event.client.SourceId source_id = 5;
-  int64 evaluated_at = 6;
-  bool user_attributes_updated = 7;
+  UserEvaluationCondition user_evaluation_condition = 6;
```
masaaania, line 42: Any SDK does not use the `evaluated_at` and `user_attributes_updated`.

cre8ivejp (approved):
Thank you!

PR #462, opened 2023-07-12
chore: allow calculator image tag to be set individually

masaaania:
- This PR allows calculator image tag to be set individually.
  - The calculator(Python) will be replaced by the experiment-calculator(Go) in the future, so we decided to stop new calculator image builds from now on.


cre8ivejp (approved):
👍

PR #470, opened 2023-07-18
chore: upgrade protoc to 23.4

cre8ivejp:
I confirmed that the e2e test is passing without errors.


.github/workflows/pr-go.yaml
```diff
@@ -6,17 +6,18 @@ on:
       - main
     paths-ignore:
       - "ui/web-v2/**"
+      - "!ui/web-v2/apps/admin/src/proto/**"
```
cre8ivejp, line 9: Force to build the web project when the proto changes.

.github/workflows/pr-go.yaml
```diff
@@ -70,10 +70,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+      - uses: actions/setup-go@v4
```
cre8ivejp, line 73: Not sure why it worked until now.
We need this so it can find the binary in the go/bin path.

Error: `protolock: No such file or directory`

masaaania (approved):
Great! 👏

PR #472, opened 2023-07-19
chore: add stop function to be called when the app gets the sigterm signal

cre8ivejp:
Ensure to stop the health check before stopping the application so the Kubernetes Readiness can detect it faster and remove the pod from the service load balancer.


masaaania (approved):
Thank you!

PR #475, opened 2023-07-20
feat: add name and url_code to project

masaaania:
- This PR
  - adds `name` and `url_code` fields to the `project.proto` message.
  - modified the `CreateProject` and `CreateTrialProject` API so that they generate a UUID for their `id` field.
  - adds `name` and `url_code` fields to the `domain/event.proto` message too.
    - The `domain/event.proto` is addressed by the `auditlog_persister` service, but `name` and `url_code` are not used currently.

- The `UpdateProject` API is supposed to be modified in the next PR.


cre8ivejp (approved):
Thank you 🎉

PR #478, opened 2023-07-21
ci: compress build time 

masaaania:
- removes needless proto-compile processes.
- modifies go build process so that it works concurrently.
  - [Concurrent go build process took 1m4s](https://github.com/bucketeer-io/bucketeer/actions/runs/5620512977/job/15229596827). (Sequential build process takes about 2min.)


cre8ivejp (dismissed):
Nice!

cre8ivejp (approved):
👍

PR #481, opened 2023-07-25
chore: adjust features and segments cache TTL

masaaania:
- This PR sets to 0 the TTL of the features and user-segments caches.
  - We discussed the risk of MySQL's downtime and decided to change the TTL setting so that not to clear the cache.
  - As for refreshing the cache, we plan to do it more securely.


cre8ivejp (approved):
Thank you!

PR #484, opened 2023-07-28
chore: modify environment APIs to write v1 and v2 temporarily

masaaania:
- Temporarily, the environment API V1 persists objects to V1 and V2 tables until all V1 API will be replaced.


cre8ivejp (approved):
Thank you!

PR #498, opened 2023-08-02
feat: separete the mysql config for DDL and DML

masaaania:
- We need stronger permission to migrate the DB structure than usual manipulation such as SELECT, CREATE, and so on.
- So, I separate the MySQL user and pass configs.


cre8ivejp (approved):
Thank you!

PR #501, opened 2023-08-02
feat: enable Google Analytics4

masaaania:
- This PR enables Google Analytics 4.
  - The react project can only use environment variables at build time.
  - I prepare the window.env object so that the project reads GA4 tracking ID from that object.
  - The window.env object is created from the runtime environment variable.
- Since the `react-ga` library doesn't support GA4, I replace it with the `react-ga4` library.


cre8ivejp (approved):
Thank you!

cre8ivejp (approved):
Thank you!

PR #506, opened 2023-08-09
Inmemory cache

masaaania:


pkg/gateway/cmd/server.go
```diff
@@ -238,13 +238,16 @@ func (s *server) Run(ctx context.Context, metrics metrics.Metrics, logger *zap.L
 	defer redisV3Client.Close()
 	redisV3Cache := cachev3.NewRedisCache(redisV3Client)
 
+	inMemoryCache := cachev3.NewInMemoryCache(cachev3.WithEvictionInterval(cachev3.EnvironmentAPIKeyEvictionInterval))
```
cre8ivejp, line 241: It would be nice if we could configure the eviction interval via the command flag using the config file. But we can do this in another PR.

cre8ivejp (approved):
Nice work!
Thanks!

PR #507, opened 2023-08-10
chore: remove target store from batch service

cre8ivejp:
Because the batch server doesn't need a cache, and using a cache could have issues using data out of date or tests failing due to update timing, we should get the latest data from the DB.
Since the volume request is too low per minute, there is no problem with getting it from the DB directly.

### Things done

- Removed the target store from batch service
- Fixed the e2e TestGrpcGetEvaluationsWithoutTag failing sometimes
- Updated unit and e2e tests


pkg/batch/jobs/opsevent/datetime_watcher.go
```diff
@@ -49,45 +51,78 @@ func NewDatetimeWatcher(
...
 				lastErr = err
 			}
 		}
 	}
 	return
 }
 
+func (w *datetimeWatcher) listEnvironments(ctx context.Context) ([]*envproto.EnvironmentV2, error) {
+	resp, err := w.envClient.ListEnvironmentsV2(ctx, &envproto.ListEnvironmentsV2Request{
+		PageSize: 0,
```
masaaania, line 95: To make it behave like the V1 API, the following line needs to be added.
```suggestion
		PageSize: 0,
		Archived:  wrapperspb.Bool(false)
```

The V1 API always filtered out the deleted environments.
Since the V2 environment has the `archived` field instead of the `deleted` field, we need to specify where including or not `archived` environments.

pkg/batch/jobs/opsevent/event_count_watcher.go
```diff
@@ -76,43 +78,76 @@ func NewEventCountWatcher(
...
 				lastErr = err
 			}
 		}
 	}
 	return
 }
 
+func (w *eventCountWatcher) listEnvironments(ctx context.Context) ([]*envproto.EnvironmentV2, error) {
+	resp, err := w.envClient.ListEnvironmentsV2(ctx, &envproto.ListEnvironmentsV2Request{
+		PageSize: 0,
```
masaaania, line 111: Here is the same.

test/e2e/autoops/auto_ops_test.go
```diff
@@ -343,17 +333,7 @@ func TestGrpcOpsEventRateBatch(t *testing.T) {
...
-
-	feature = getFeature(t, featureClient, featureID)
-	if feature.Enabled {
-		t.Fatalf("feature must be disabled")
-	}
-	autoOpsRules = listAutoOpsRulesByFeatureID(t, autoOpsClient, featureID)
-	if autoOpsRules[0].TriggeredAt == 0 {
-		t.Fatalf("triggered at must not be zero")
-	}
+	checkIfAutoOpsRulesAreTriggered(t, featureID)
```
masaaania, line 336: 👏

pkg/batch/jobs/opsevent/datetime_watcher.go
cre8ivejp: I have added it.
Thanks!

masaaania (approved):
Thank you!!

PR #508, opened 2023-08-10
perf: modify api-gateway to fetch features cache without single flight

masaaania:
- This PR modifies the api-gateway behavior when it fetches features from Redis.
  - After merging this PR, the api-gateway accesses Redis without single flight.
  - When the features cache doesn't exist in Redis, the api-gateway fetches features from the feature server in the backend service with single flight.


cre8ivejp (approved):
Thank you!

PR #512, opened 2023-08-15
revert: modify api-gateway to fetch features cache without single flight

cre8ivejp:
Reverts bucketeer-io/bucketeer#508


masaaania (approved):
Thank you!

PR #513, opened 2023-08-17
chore: add tracing settings

masaaania:
- This PR adds tracing settings with OpenCensus.

- Although OpenCensus' Go library was archived on 1st August 2023, we continue to use it for now.
  - We plan to replace it with OpenTelemetry after [Pub/Sub library supports Otel](https://github.com/googleapis/google-cloud-go/issues/4665).


cre8ivejp (approved):
Nice!
Thanks!

PR #516, opened 2023-08-25
chore: add GoogleTagManager setting

masaaania:
- This PR
  - removes the Google Analytics4 setting
  - adds the Google Tag Manager setting
    - Google Tag Manager collects the events on our console and sends them to GA4.
  - adds the setting to send the event `feature_created` when a user creates a feature flag on the console.


cre8ivejp (approved):
Nice work!

PR #518, opened 2023-08-28
fix: getEnvironmentV2 fails to get env with empty ID

masaaania:
- This PR modifies the GetEnvironmentV2 API to retrieve the environment even with an empty ID.
- This PR fixes the batch server doesn't emit logs.


cre8ivejp (approved):
Thank you!

PR #519, opened 2023-08-31
fix: time location in unit test

Ubisoft-potato:
I noticed an error while executing the Github Action `pr-go`. This [line](https://github.com/bucketeer-io/bucketeer/blob/main/pkg/batch/jobs/notification/mau_count_watcher_test.go#L175)  has not been set to the same timzone as the [line](https://github.com/bucketeer-io/bucketeer/blob/main/pkg/batch/jobs/notification/mau_count_watcher.go#L75) in `mauCountWatcher.Run` function. 

Due to the timezone used in `mauCountWatcher` being `Japan`, when I ran this unit test in the `Beijing` timezone on August 31, 2023, at 11:56 PM, I encountered the following error:
image

So, we should set the same time zone in the unit test. :)

### DIFF
--- a/pkg/batch/jobs/notification/mau_count_watcher_test.go
+++ b/pkg/batch/jobs/notification/mau_count_watcher_test.go
@@ -172,7 +172,7 @@ func TestCreateMAUNotification(t *testing.T) {
 						Cursor:       "",
 					}, nil)
 				// get user count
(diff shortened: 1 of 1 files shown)


masaaania (approved):
Thank you!

cre8ivejp (approved):
👍

PR #530, opened 2023-09-27
chore: translate the localize file from Japanese to English

cre8ivejp:
### Things done

- Translated the status and error messages to English
- Fixed the context in the unit test not including the metadata


pkg/locale/locale.go
```diff
@@ -57,12 +57,11 @@ func GetLocation(timezone string) (*time.Location, error) {
 func getAcceptLang(ctx context.Context) string {
 	md, ok := metadata.FromIncomingContext(ctx)
 	if !ok {
-		return Ja
+		return En
```
cre8ivejp, line 60: I changed to return English when the `accept-language` is not set.

masaaania (approved):
Thank you!!

PR #545, opened 2023-10-03
chore: add none when the tag is empty for better readability

cre8ivejp:


pkg/eventpersisterdwh/persister/evaluation_event.go
```diff
@@ -158,7 +158,7 @@ func (w *evalEvtWriter) convToEvaluationEvent(
 	}
 	tag := e.Tag
 	if tag == "" {
-		// For requests with no tag, it will insert "none" instead, until all old SDK clients are updated
+		// Tag is optional, so we insert none when is empty.
 		tag = "none"
```
cre8ivejp, line 162: It was already adding none for Evaluation events when the tag was empty, so I changed the comment.

masaaania (approved):
Thank you!

PR #551, opened 2023-10-10
chore: remove update env v1 test

masaaania:
- Remove the update environment v1 test because it interferes with the v2 test and causes it to fail.


cre8ivejp (approved):
Thanks!

PR #552, opened 2023-10-11
feat: support rename function for project

masaaania:
- This PR
  - modifies the form UI to add project so that the user can specify the urlCode and name of the project.
  - modifies the form UI to update project so that the user can update the project name.
  - modifies the project list UI to show their name instead of ID.
  - modifies the sort function of the project list so that the user can sort via their name instead of ID.


pkg/environment/api/project.go
```diff
@@ -272,9 +264,8 @@ func validateCreateProjectRequest(req *environmentproto.CreateProjectRequest, lo
 		}
 		return dt.Err()
 	}
-	// TODO Once we support new create project API requiring name instead of id, we should validate name only.
 	name := strings.TrimSpace(req.Command.Name)
-	if !projectNameRegex.MatchString(name) && !projectIDRegex.MatchString(req.Command.Id) {
+	if !projectNameRegex.MatchString(name) {
```
cre8ivejp, line 268: Please remove the regex for the name. Otherwise, we can't use space, etc.

cre8ivejp: For the name field, please use this as a reference.

https://github.com/bucketeer-io/bucketeer/blob/fcacff8f1a28cca27f810ec16a3e670700d15c55/ui/web-v2/apps/admin/src/pages/feature/formSchema.ts#L33

masaaania: @cre8ivejp 
Thank you, I fixed it 👍

cre8ivejp (approved):
Thank you!

PR #553, opened 2023-10-15
feat: experiment calculator improvement

Ubisoft-potato:
### Summary
- This will add a new batch job that performs the following steps:
  1. Lists all environments (v2)
  2. Lists all experiments in each environment
  3. Calculates each experiment by calling the experiment-calculator.

- Additionally, this PR includes improvements and bug fixes for the experiment-calculator.


pkg/batch/cmd/server/server.go
```diff
@@ -267,6 +274,17 @@ func (s *server) Run(ctx context.Context, metrics metrics.Metrics, logger *zap.L
 		executor.WithLogger(logger),
 	)
 
+	experimentCalculatorClient, err := experimentcalculatorclient.NewClient(*s.experimentCalculatorService, *s.certPath,
```
masaaania, line 277: Please close the experimentalCalculatorClient in the defer function.

manifests/bucketeer/charts/experiment-calculator/values.yaml
```diff
@@ -80,11 +80,11 @@ health:
   # The duration is up to resources.
   # cf. CPU: 500m, MEM: 3Gi -> 280sec
   livenessProbe:
-    initialDelaySeconds: 300
```
cre8ivejp, line 83: The service takes around 4-5 minutes to get ready. Is there a reason to change it?

cre8ivejp:
@Ubisoft-potato can you describe the improvements that you made compared to the old implementation, please?

manifests/bucketeer/charts/experiment-calculator/values.yaml
Ubisoft-potato: Remember that I have create a PR for httpstan that precompile the stan model: https://github.com/bucketeer-io/bucketeer/pull/454?

For now, the service will be ready in around 10s. So if we still set this to 300s, we need to wait 300s for the pod to be ready.

Ubisoft-potato:
> @Ubisoft-potato can you describe the improvements that you made compared to the old implementation, please?

The improvement is the `experiment-calculator` initialization, we need to start `experiment-calculator` after httpstan.

But I didn't find a solution using k8s before. 

Here is the code: https://github.com/bucketeer-io/bucketeer/pull/553/files#diff-b0f1a6e9d024e46f089e3bc6573a1c8f6dfaf821d6b3898e3c06794bbc4d0dd8R79-R90

manifests/bucketeer/charts/experiment-calculator/values.yaml
cre8ivejp: Yes, I remembered. It improved a little, but it took more than 3 minutes in production to get ready. So, the health check restarts the container before it gets ready.

I will check it on the dev environment to ensure that it gets ready in 10 seconds.

pkg/experimentcalculator/experimentcalc/experiment_calculator.go
```diff
@@ -78,12 +76,18 @@ func NewExperimentCalculator(
...
 ) *ExperimentCalculator {
 	registerMetrics(metrics)
-	compiledModel, err := httpStan.CompileModel(context.TODO(), stan.ModelCode())
-	if err != nil {
-		logger.Error("Failed to compile model",
-			zap.Error(err),
-		)
-		return nil
+	var compiledModel stan.ModelCompileResp
+	for {
```
cre8ivejp, line 80: We should set a limit here. Otherwise, it could enter the infinity loop.
How frequently it could fail while compiling the model?

cre8ivejp:
> The improvement is the experiment-calculator initialization, we need to start experiment-calculator after httpstan.
> But I didn't find a solution using k8s before.

@Ubisoft-potato, my concern is if the httpstan doesn't compile fast enough, the batch service won't start until it does.

Ubisoft-potato:
> @Ubisoft-potato, are you going to remove the python calculator from the helm manifest in another PR?

Yes, I will delete it after you confirm this PR working without error in dev environment.

pkg/experimentcalculator/experimentcalc/experiment_calculator.go
Ubisoft-potato: Currently, we don't need to compile the model, we have already compiled it in httpstan image, means when the httpstan container is ready, we can directly use the stan model.

But httpstan container may start after experiment-calculator container.

Ubisoft-potato:
> > The improvement is the experiment-calculator initialization, we need to start experiment-calculator after httpstan.
> > But I didn't find a solution using k8s before.
> 
> @Ubisoft-potato, my concern is if the httpstan doesn't compile fast enough, the batch service won't start until it does.

same here: https://github.com/bucketeer-io/bucketeer/pull/553#discussion_r1360280429

pkg/experimentcalculator/experimentcalc/experiment_calculator.go
cre8ivejp: I understand, but this kind of implementation could lead to bugs in the future, so it's safer we set a limit in case it doesn't work as expected.

Ubisoft-potato: No problem, I will set it to retry 3 times to avoid this.
In my dev container, httpstan generally delay 1~2 second after experiment-calculator is ready.

Ubisoft-potato: Done!

pkg/batch/cmd/server/server.go
Ubisoft-potato: Done!

masaaania (approved):
Thank you very much 🎉

PR #562, opened 2023-10-24
chore: change not to save events after the experiment is ended

cre8ivejp:
To avoid saving events that won't be calculated by the experiment calculator, I changed the condition to ignore events older than the experiment's `stopAt`.


pkg/eventpersisterdwh/persister/persister_test.go
```diff
@@ -164,7 +164,7 @@ func TestConvToEvaluationEvent(t *testing.T) {
 	}
 	evaluationEvent := &eventproto.EvaluationEvent{
 		Tag:            "tag",
-		Timestamp:      t1.UnixMicro(),
```
cre8ivejp, line 167: The client sends it in seconds, not in micro, so I fixed the input data.

masaaania (approved):
Thank you!

PR #567, opened 2023-10-26
chore: change to return the object when a project or environment is created

cre8ivejp:
To make things easier when writing hack or e2e tests, I changed to return the object when a project or environment Is created.


masaaania (approved):
Thank you!

PR #568, opened 2023-10-26
fix: project hack script not creating environment correctly

cre8ivejp:
There is a bug while creating the environment. Instead of passing the project ID, it passes a random uuid as the project ID.

### Things done

- Fixed the project ID
- Changed the flag from `bool` to `strings` so the user can pass the environment he wants instead of always creating 3 environments as default.
- Added the flag `url-code` instead of using the name
- Updated the README

### DIFF
--- a/hack/create-project/command.go
+++ b/hack/create-project/command.go
@@ -27,7 +27,6 @@ import (
 	environmentclient "github.com/bucketeer-io/bucketeer/pkg/environment/client"
 	"github.com/bucketeer-io/bucketeer/pkg/metrics"
 	"github.com/bucketeer-io/bucketeer/pkg/rpc/client"
-	"github.com/bucketeer-io/bucketeer/pkg/uuid"
 	environmentproto "github.com/bucketeer-io/bucketeer/proto/environment"
 )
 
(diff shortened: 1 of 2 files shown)


hack/create-project/command.go
```diff
@@ -66,44 +72,37 @@ func (c *command) Run(ctx context.Context, metrics metrics.Metrics, logger *zap.
 	req := &environmentproto.CreateProjectRequest{
 		Command: &environmentproto.CreateProjectCommand{
 			Name:        *c.name,
-			UrlCode:     *c.name,
+			UrlCode:     strings.ToLower(strings.ReplaceAll(*c.urlCode, " ", "-")),
```
cre8ivejp, line 70: I added this line to lower the string and replace whitespace for hyphens. Same thing while creating environments.

hack/create-project/command.go
```diff
@@ -49,14 +49,20 @@ func registerCommand(r cli.CommandRegistry, p cli.ParentCommand) *command {
...
 		description:        cmd.Flag("description", "(optional) Description of an project.").String(),
-		createEnvironments: cmd.Flag("create-environments", "create environments or not").Bool(),
+		createEnvironments: cmd.Flag("create-environment", "A list of environment names to be created").Strings(),
 	}
 	r.RegisterCommand(command)
 	return command
 }
 
 func (c *command) Run(ctx context.Context, metrics metrics.Metrics, logger *zap.Logger) error {
+	logger.Info("Yeah babe",
```
masaaania, line 61: :sunglasses:

masaaania (approved):
Thank you! 🙏

Issue #314, opened 2023-05-08
feat: flag trigger


cre8ivejp:
Make it possible to turn on/off a flag via webhooks. It will make it easier for users who want to integrate actions in their metrics monitoring.

PR #586, opened 2023-11-14
docs: flag trigger RFC

Ubisoft-potato:
This PR will add the RPC documentation for the feature flag tigger functionality implementation: https://github.com/bucketeer-io/bucketeer/issues/314


docs/rfcs/autoops-flag-trigger.md
```diff
@@ -0,0 +1,193 @@
...
+```proto
+message ListFlagTriggersRequest {
+    string feature_id = 1;
+    string environment_namespace = 2;
+    int32 page = 3;
+    int32 page_size = 4;
+}
+```
+
+#### AutoOps Trigger: WebHook
```
cre8ivejp, line 132: Due to security reasons, we shouldn't show the URL more than once to the user on the console UI.
Can you add it to the RFC how are you going to deal with this when sending the response for the `ListFlagTriggersRequest`?

docs/rfcs/autoops-flag-trigger.md
```diff
@@ -0,0 +1,193 @@
...
+```
+
+- Update a flag trigger
+
+```Grpc
+rpc UpdateFlagTrigger(UpdateFlagTriggerRequest) returns (UpdateFlagTriggerResponse) {}
+```
+
+```proto
+message UpdateFlagTriggerRequest {
```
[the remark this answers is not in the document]
Ubisoft-potato: Yes, I believe so. I had planned on using the `Update` API to reset or update the trigger.
If we do not need to update the trigger, I suggest changing the `Update` API to only have a `Reset` URL.

docs/rfcs/autoops-flag-trigger.md
Ubisoft-potato: OK, I think I forget this use case.

And I have checked the `FeatBit` yesterday, they just show the url after user create the trigger for the first time. 

So I think it would be easy to achieve this:

- we will return the url in the `CreateFlagTrigger` API response
- return masked url in `ListFlagTrigger` and `GetFlagTrigger` API

docs/rfcs/autoops-flag-trigger.md
```diff
@@ -0,0 +1,228 @@
...
+
+Then return the new trigger url to the user.
+
+* Only show the trigger url to the user for the first time.
+
+    - When the user creates a new flag trigger, we will generate a new `uuid` and encrypt the `flagTriggerSecret` as the
+      above steps, and then return the trigger url to the user.
+    - When the user reset the trigger url, we will generate a new `uuid` and encrypt the `flagTriggerSecret` as the
+      above steps, and then return the trigger url to the user.
+    - In the `ListFlagTrigger` and `GetFlagTrigger` API response, we will return partial masked trigger url to the user.
```
cre8ivejp, line 228: How will you check if it should return a full or partial URL?

docs/rfcs/autoops-flag-trigger.md
```diff
@@ -0,0 +1,228 @@
...
+
+- Delete a flag trigger
+
+```Grpc
+rpc DeleteFlagTrigger(DeleteFlagTriggerRequest) returns (DeleteFlagTriggerResponse) {}
+```
+
+```proto
+message DeleteFlagTriggerRequest {
+    string id = 1;
```
cre8ivejp, line 128: Just to let you know, all requests contain a command.
But this minor detail doesn't need to be included in the RFC.
We want to ensure that the whole flow is correct. Which it is.

Reference.
https://github.com/bucketeer-io/bucketeer/blob/main/proto/experiment/service.proto#L191-L195

Ubisoft-potato: Yeah, I have noticed that the command patten in the project.

docs/rfcs/autoops-flag-trigger.md
Ubisoft-potato: I think we can return partial URL directly in the `ListFlagTrigger` and `GetFlagTrigger` API response, because we only return the full URL in the `CreateFlagTrigger` API response.

docs/rfcs/autoops-flag-trigger.md
```diff
@@ -0,0 +1,228 @@
...
+    `description`           text         NOT NULL default '',
+    `trigger_times`         int          NOT NULL,
+    `last_triggered_at`     bigint       NOT NULL,
+    `uuid`                  varchar(512) NOT NULL default '',
+    `disabled`              tinyint(1) NOT NULL DEFAULT '0',
+    `deleted`               tinyint(1) NOT NULL DEFAULT '0',
+    `created_at`            bigint       NOT NULL,
+    `updated_at`            bigint       NOT NULL,
+    PRIMARY KEY (`id`)
+)ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;
```
masaaania, line 30: For consistency with other tables, please can you add the Primary Key and Foreign Key like:
```
  PRIMARY KEY (`id`,`environment_namespace`),
  CONSTRAINT `foreign_flag_triggers_feature_id_environment_namespace`
  FOREIGN KEY (`feature_id`, `environment_namespace`)
  REFERENCES `feature` (`id`, `environment_namespace`)
  ON DELETE NO ACTION
  ON UPDATE NO ACTION
```

Ubisoft-potato: Thanks for your supplement!

cre8ivejp (approved):
Thank you!!

PR #589, opened 2023-11-22
feat: add organization proto

masaaania:
## This PR

- adds functions and messages for the organization to proto.
- implementation will be pushed in the next PR.


proto/environment/command.proto
```diff
@@ -63,3 +63,31 @@ message EnableProjectCommand {}
...
 
 message ConvertTrialProjectCommand {}
+
+message CreateOrganizationCommand {
+  string name = 1;
+  string url_code = 2;
+  string description = 3;  // optional
+}
+
+message CreateTrialOrganizationCommand {
```
cre8ivejp, line 73: What do you think about adding a flag in the `CreateOrganizationCommand` to create a trial project to reduce the APIs?

masaaania: Nice idea 😄.

masaaania:
@cre8ivejp 

I updated and fixed the PR. Please take a look when you have a time.

proto/environment/command.proto
```diff
@@ -63,3 +63,28 @@ message EnableProjectCommand {}
 message DisableProjectCommand {}
 
 message ConvertTrialProjectCommand {}
+
+message CreateOrganizationCommand {
+  string name = 1;
+  string url_code = 2;
+  string description = 3;  // optional
+  string is_trial = 4;
```
cre8ivejp, line 71: Please use boolean instead.

cre8ivejp (approved):
Thank you!!

proto/environment/command.proto
masaaania: @cre8ivejp 
Thank you, nice catch!! 🙏 🙏

PR #590, opened 2023-11-22
chore: add redis connection status to health check

cre8ivejp:
We already have an interface to check the Redis status, but we didn't implement it in the container health check.
By implementing this If for some reason the container gets a Redis connection failure, it will restart the container to recreate the connection.

### DIFF
--- a/pkg/gateway/cmd/server.go
+++ b/pkg/gateway/cmd/server.go
@@ -254,9 +254,8 @@ func (s *server) Run(ctx context.Context, metrics metrics.Metrics, logger *zap.L
 		api.WithLogger(logger),
 	)
 
-	// By checking the Redis in the API Gateway, if it fails, it won't get the features from the DB.
-	// But if we don't test it, the DB might fail due to high requests until the container restarts.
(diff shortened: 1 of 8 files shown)


pkg/gateway/cmd/server.go
```diff
@@ -257,6 +257,7 @@ func (s *server) Run(ctx context.Context, metrics metrics.Metrics, logger *zap.L
 	healthChecker := health.NewGrpcChecker(
 		health.WithTimeout(time.Second),
 		health.WithCheck("metrics", metrics.Check),
+		health.WithCheck("redis", redisV3Client.Check),
```
cre8ivejp, line 260: By checking the Redis in the API Gateway, if it fails, it won't get the features from the DB.
But if we don't test it, DB may fail due to high requests until the container restarts.

cre8ivejp: I removed it because we want to avoid downtime as much as possible.

cre8ivejp:
@masaaania, I have pushed the changes we discussed.

masaaania (approved):
Thank you!!

PR #593, opened 2023-11-27
feat: add experiments cache to event persister dwh

cre8ivejp:
I am implementing a cache for Experiments in the event persisted DWH to reduce the high load on MySQL.
I set the default TTL to 1 minute. We can change it in the future if needed.


pkg/eventpersisterdwh/persister/goal_event.go
```diff
@@ -319,28 +323,39 @@ func (w *goalEvtWriter) listExperiments(
...
-				if experimentSize == 0 || experimentSize < listRequestSize {
-					return experiments, nil
-				}
-				cursor = resp.Cursor
+			// Get the experiment cache
+			expList, err := w.cache.Get(environmentNamespace)
+			if err == nil {
+				return expList.Experiments, nil
+			}
+			w.logger.Info(
```
cre8ivejp, line 331: @masaaania, maybe it will be better if we remove this log because it will log every minute.

masaaania (approved):
Thank you!!

PR #594, opened 2023-11-28
chore: add temporary implementation to create organization with project

masaaania:
## This PR
- adds temporary implementation to create organizations at the same time as projects
- modifies the implementation to create env and project so that organization_id is saved.


cre8ivejp (approved):
Excellent work!
Thank you 🎉

PR #595, opened 2023-11-28
feat: add auto ops rules cache to event persister ops

cre8ivejp:
I am implementing a cache for Auto Ops Rules in the event persister OPS service to reduce the load on MySQL.
I set the default TTL to 1 minute. We can change it in the future if needed.

Redis was already implemented in the persister service and is a persistent-redis instance. Usually, we use the non-persistent-redis instance for caching, but since I'm setting the TTL in the key I don't think there will be a problem. Also, the persistent instance uses many more resources so we won't have an issue either.


masaaania (approved):
Thank you!

PR #613, opened 2023-12-07
fix: batch domain-event-informer

Ubisoft-potato:


pkg/batch/api/api_test.go
```diff
@@ -485,40 +446,6 @@ func TestProgressiveRolloutWatcher(t *testing.T) {
 	}, setupMock)
 }
 
-func TestDomainEventInformer(t *testing.T) {
```
Ubisoft-potato, line 488: Since we moved `createPuller` to informer, we are unable to mock puller. Therefore, let's delete this unit test here.

We will need to create an end-to-end test for `DomainEventInformer`.

pkg/batch/jobs/notification/domain_event_informer.go
```diff
@@ -104,36 +145,79 @@ func NewDomainEventInformer(
...
 	return &domainEventInformer{
 		environmentClient: environmentClient,
-		puller:            puller.NewRateLimitedPuller(p, options.maxMPS),
 		sender:            sender,
 		opts:              &options,
 		logger:            options.logger.Named("sender"),
 	}
 }
 
+func (i *domainEventInformer) createPuller(ctx context.Context) (puller.RateLimitedPuller, func(), error) {
```
cre8ivejp, line 154: It's not necessary, but you could move this function below the `Run` function.

masaaania (approved):
LGTM! Thank you!

cre8ivejp (approved):
Thank you!

PR #614, opened 2023-12-07
feat: remove role field from editor object

masaaania:
## This PR
- removes the role field from the editor object because it is unused.


cre8ivejp (approved):
Thank you!

PR #624, opened 2023-12-25
chore: shorten flag trigger url

Ubisoft-potato:
## Changes

- Use a hash function to generate the secret for the flag trigger. This function will generate a secret token consisting of 40-50 characters.
- Add a new column called `token` to the `flag_trigger` table.
- Remove `uuid` column for `flag_trigger` table.
- Validate the client by checking the secret token in the database while calling the flag trigger webhook.
- Display part of the flag triggers webhook url secret, for example: `https://dev.bucketeer.jp/webhook/AbCdE*****`.

### MySQL table update
```sql
CREATE TABLE IF NOT EXISTS `flag_trigger` (
  `id`                    VARCHAR(255) NOT NULL,
  `feature_id`            VARCHAR(255) NOT NULL,
  `type`                  INT          NOT NULL,
  `action`                TINYINT(1)   NOT NULL,
  `description`           TEXT         NOT NULL,
  `trigger_count`         INT          NOT NULL,
  `last_triggered_at`     BIGINT       NOT NULL,
  `token`                 VARCHAR(255) NOT NULL,
  `disabled`              TINYINT(1)   NOT NULL DEFAULT '0',
  `created_at`            BIGINT       NOT NULL,
  `updated_at`            BIGINT       NOT NULL,
  `environment_namespace` VARCHAR(255) NOT NULL,
  PRIMARY KEY (`id`, `environment_namespace`),
  CONSTRAINT `foreign_flag_trigger_feature_id_environment_namespace` 
    FOREIGN KEY (`feature_id`, `environment_namespace`)
    REFERENCES `feature` (`id`, `environment_namespace`)
    ON DELETE NO ACTION
    ON UPDATE NO ACTION
);
```


pkg/feature/api/flag_trigger.go
cre8ivejp, line 1: Please rename the field `secret` to `token` in the `FlagTriggerWebhookRequest` proto file and its references.

pkg/feature/domain/flag_trigger.go
```diff
@@ -88,3 +90,22 @@ func (ft *FlagTrigger) UpdateTriggerUsage() error {
 	ft.TriggerCount = ft.TriggerCount + 1
 	return nil
 }
+
+func (ft *FlagTrigger) GenerateToken() error {
+	secret := NewFlagTriggerSecret(
```
cre8ivejp, line 95: Because we now save the token, we no longer need to save the UUID in the DB. Also, we just need the UUID (One-time) to generate the token.

https://github.com/bucketeer-io/bucketeer/pull/624/files#diff-5559c5f4ce4b05e5cce1d1577c268d4987c108d1a6153348c8aa8ca76f377ed2R40-R43

Ubisoft-potato: So we can remove the `uuid` column for `flag_trigger` table.

cre8ivejp: Yes.

pkg/feature/storage/v2/sql/flag_trigger/insert_flag_trigger.sql
```diff
@@ -6,7 +6,6 @@ INSERT INTO flag_trigger(id,
                          description,
                          trigger_count,
                          last_triggered_at,
-                         uuid,
```
cre8ivejp, line 9: You also need to update the VALUES.

pkg/feature/domain/flag_trigger.go
cre8ivejp: I looked at your last commit removing the column from MySQL, but I couldn't find the changes removing it from the code.

pkg/feature/domain/flag_trigger.go
```diff
@@ -92,11 +77,15 @@ func (ft *FlagTrigger) UpdateTriggerUsage() error {
 }
 
 func (ft *FlagTrigger) GenerateToken() error {
+	newTriggerUuid, err := uuid.NewUUID()
+	if err != nil {
+		return err
+	}
 	secret := NewFlagTriggerSecret(
 		ft.GetId(),
```
cre8ivejp, line 85: We just need need the UUID. you can delete the `NewFlagTriggerSecret`.

Ubisoft-potato: You mean we can use `uuid` as the `token` directly?

cre8ivejp: I mean, you don't need to create a struct just for one field.
You can call it directly.

E.g.
```go
h.Write([]byte(newTriggerUuid.String()))
```

Ubisoft-potato: I understand what you mean. 
We can eliminate the `FlagTriggerSecret` struct and then hash the `UUID` to obtain the token.

pkg/feature/api/flag_trigger.go
cre8ivejp, line 1: Please delete the token data from the proto message before returning the response for the listing and getting APIs. Otherwise, anyone who has access to the console could see the token and generate the URL manually.

Ubisoft-potato: Sure.

pkg/feature/storage/v2/flag_trigger.go
```diff
@@ -169,7 +169,6 @@ func (f flagTriggerStorage) GetFlagTrigger(
 		&trigger.Description,
 		&trigger.TriggerCount,
 		&trigger.LastTriggeredAt,
-		&trigger.Token,
```
cre8ivejp: We can't remove the token from the query because the Update API calls the GET query. Otherwise, it will delete the current token data when the update query is committed.

https://github.com/bucketeer-io/bucketeer/blob/aa5e4e5ddc4acbaa333ed80d26ff2833f73895e4/pkg/feature/api/flag_trigger.go#L177-L181

It will be safer to remove it at the API layer to avoid bugs.

proto/feature/service.proto
```diff
@@ -502,7 +502,7 @@ service FeatureService {
   rpc FlagTriggerWebhook(FlagTriggerWebhookRequest)
       returns (FlagTriggerWebhookResponse) {
     option (google.api.http) = {
-      post: "/webhook/{token}"
+      get: "/webhook/{token}"
```
cre8ivejp, line 505: Please keep it as POST.

pkg/feature/storage/v2/flag_trigger.go
cre8ivejp: We need to delete the token from the List API response, too.
Also, as a default rule, remove the token from any API that returns the flag trigger. Such as create API, etc.

cre8ivejp:
@Ubisoft-potato, can you fix the conflict?

Ubisoft-potato:
> @Ubisoft-potato, can you fix the conflict?

Solved.

pkg/feature/api/flag_trigger.go
cre8ivejp, line 1: Before enabling or disabling the feature, we must check the current status. Otherwise, it will return an error if you try to enable a flag that is already enabled, and vice-versa.

https://github.com/bucketeer-io/bucketeer/blob/40b8e72ed552f61ed681af490969cd79d607c75b/pkg/feature/api/flag_trigger.go#L831

If no action is needed, we should skip and update the trigger status.

cre8ivejp:
@Ubisoft-potato, please fix the conflicts.

cre8ivejp (approved):
Thank you 🎉

PR #629, opened 2023-12-26
fix: validation while creating a progressive rollout

cre8ivejp:
It returns an error even if the operation is completed.


masaaania (approved):
LGTM!

PR #632, opened 2023-12-27
fix: batch server cannot execute ddl to mysql

masaaania:
- This PR adds the mysql connection settings for the admin user.
  - Only the admin user can execute ddl command.


cre8ivejp (approved):
Thank you!

PR #649, opened 2024-01-04
chore: add double write account v1 and v2

masaaania:
## This PR
- adds implementation for double write accounts v1 and v2.
  - This implementation is needed so that the users can create/update/delete account v2 at the same time as account v1.


cre8ivejp (approved):
Thank you!

PR #650, opened 2024-01-10
test: fix delete e2e data not working

cre8ivejp:
The delete e2e data workflow fails when deleting the feature flags because we must delete the trigger first.
There was one place that didn't set the test id in the description.


masaaania (approved):
Thank you!

PR #651, opened 2024-01-10
chore: replace account v1 to v2 when checking the role

masaaania:
This PR updates the `CheckRole()` function in the `/pkg/role/role.go` to use the accountV2 instead of the accountV1.

The following modules use the `CheckRole()`, so they are modified.

### DIFF
--- a/pkg/autoops/api/api_test.go
+++ b/pkg/autoops/api/api_test.go
@@ -704,7 +704,7 @@ func TestGetAutoOpsRuleMySQL(t *testing.T) {
 	}{
 		{
 			desc:        "err: ErrIDRequired",
-			req:         &autoopsproto.GetAutoOpsRuleRequest{},
+			req:         &autoopsproto.GetAutoOpsRuleRequest{EnvironmentNamespace: "ns0"},
 			expectedErr: createError(statusIDRequired, localizer.MustLocalizeWithTemplate(locale.RequiredFieldTemplate, "id")),
 		},
 		{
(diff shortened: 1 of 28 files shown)


pkg/autoops/api/api_test.go
```diff
@@ -921,13 +921,23 @@ func createAutoOpsService(c *gomock.Controller, db storage.Client) *AutoOpsServi
 	mysqlClientMock := mysqlmock.NewMockClient(c)
 	featureClientMock := featureclientmock.NewMockClient(c)
 	accountClientMock := accountclientmock.NewMockClient(c)
-	ar := &accountproto.GetAccountResponse{
-		Account: &accountproto.Account{
-			Email: "email",
-			Role:  accountproto.Account_VIEWER,
+	ar := &accountproto.GetAccountV2ByEnvironmentIDResponse{
```
cre8ivejp, line 924: Just confirming, is there another type of getting an account response?
If not, we could just use `GetAccountV2Response`.

masaaania: @cre8ivejp 

Product code uses the `GetAccountV2ByEnvironment` API because we don't receive the organizationID from the request. (We receive the environmentID.)

https://github.com/bucketeer-io/bucketeer/pull/651/files#diff-13112ed94680ecaec33373531ff5c65e48cecae142657203e04d967a185684d5R1459

cre8ivejp (approved):
Thank you!

Issue #652, opened 2024-01-10
test: deploy httpstan when running unit tests locally


kentakozuka:
### Describe the feature you'd like

Currently, when we run make test-go the unit tests for the experiment will fail because they make requests to the httpstan server.

In the GitHub Actions, we deploy it as a service.
https://github.com/bucketeer-io/bucketeer/blob/main/.github/workflows/pr-go.yaml#L26-L31

It would be nice if we automated the httpstan process before running the unit tests locally in the Makefile.

PR #653, opened 2024-01-10
chore: add make run-httpstan

kentakozuka:
Fix https://github.com/bucketeer-io/bucketeer/issues/652

I simply add a new make command rather than automatically run docker image.
I will add the instruction about it to https://docs.bucketeer.io/contribution-guide/contributing after this PR is merged.

WDYT?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -147,9 +147,14 @@ $(GO_APP_BUILD_TARGETS): build-%:
 .PHONY: build-go
 build-go: $(GO_APP_BUILD_TARGETS)
 
+# Make sure bucketeer-httpstan is already running. If not, run "make run-httpstan".
 .PHONY: test-go
 test-go:
 	TZ=UTC CGO_ENABLED=0 go test -v ./pkg/...
+
+.PHONY: run-httpstan
+run-httpstan:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -147,10 +147,15 @@ $(GO_APP_BUILD_TARGETS): build-%:
 .PHONY: build-go
 build-go: $(GO_APP_BUILD_TARGETS)
 
+# Make sure bucketeer-httpstan is already running. If not, run "make run-httpstan".
 .PHONY: test-go
```
cre8ivejp, line 151: I was thinking of automating the test process by starting the httpstan before the tests, and stopping it after. Either way, it would be nice to have a rule to stop the container, too.

Something like this.

```suggestion
.PHONY: test-go
test-go: start-httpstan
	TZ=UTC CGO_ENABLED=0 go test -v ./pkg/...
	make stop-httpstan
```

kentakozuka: Add stop-httpstan. Manually run command before make test-go.

kentakozuka:
@cre8ivejp Fixed 👍

cre8ivejp (approved):
Thank you!

PR #660, opened 2024-01-12
fix: change the api-gateway metrics label

kakcy:
- change the api-gateway metrics label to "urlCode" from "namespace"


kakcy:
I have resubmitted PR with #662, so I will close it.

PR #663, opened 2024-01-13
feat: add system_admin field to organization

masaaania:
- This PR adds the system_admin field to the organization object.
  - The `system_admin` column has already been added to the organization table in DB.
- System admin organization
  - System admin organization cannot be archived and disabled.
  - System admin organization cannot be created when it already exists in the table.


cre8ivejp (dismissed):
Thank you 🎉

PR #664, opened 2024-01-14
feat: add get my organizations api

masaaania:
- This PR adds the API to fetch user's organizations.
  - The organizations are filtered by the user's email in the IDToken.


cre8ivejp (approved):
Thank you!

PR #666, opened 2024-01-15
test: fix list account test

masaaania:
- This PR adds a process to create an account before executing ListAccountsV2 API.


cre8ivejp (approved):
Thanks!

PR #667, opened 2024-01-15
feat: add organization select page and console account object

masaaania:
- This PR adds an organization selection page.
  - Users can select the organization they wish to log in to and use the management console as before.
- This PR adds a console account object.
  - The management console will use the new `GetMe` API instead of the `GetMeV2` API to retrieve the console account object.
  - The console account object has information such as new environment roles and organization role.


cre8ivejp (approved):
Thank you! 🎉

PR #674, opened 2024-01-16
fix: unable to fetch all orgs when user is system admin

masaaania:
- This PR fixes the GetMyOrganizations API.
  - It could not fetch all organizations when the requesting user had a system admin account.


pkg/account/api/admin_account.go
```diff
@@ -944,12 +945,35 @@ func (s *AccountService) GetMyOrganizations(
...
-	myOrgs := make([]*accountproto.MyOrganization, 0, len(accountsWithOrg))
+	if s.containsSystemAdminOrganization(accountsWithOrg) {
+		resp, err := s.environmentClient.ListOrganizations(
+			ctx,
+			&environmentproto.ListOrganizationsRequest{
+				Disabled: wrapperspb.Bool(false),
+				Archived: wrapperspb.Bool(false),
+			})
+		if err != nil {
+			return nil, err
```
cre8ivejp, line 956: We must localize the error message before returning it.
Can you add a unit test for this case, too?

masaaania: @cre8ivejp 
Thank you, I fixed it!

cre8ivejp (approved):
Thanks!

PR #662, opened 2024-01-12
fix: change the api-gateway metrics label

kakcy:

PR #688, opened 2024-01-18
fix: change the api-gateway metrics label

kakcy:
Related fixes for #662.
It is correct to pass "UrlCode" instead of Environment's "ID".


kakcy:
Thank you!

Issue #692, opened 2024-01-18
push-image workflow fails due to a newline in commit messages


kentakozuka:
### Version

main

### Describe the bug

`dispatch` job in push-image workflow fails if commit message has multiple lines.

PR #693, opened 2024-01-18
fix: dispatch workflow fails if commit message contains multi-lines

kentakozuka:
Fixes https://github.com/bucketeer-io/bucketeer/issues/692


cre8ivejp (approved):
Thanks!

PR #709, opened 2024-01-22
feat: add environment id to list accounts filter options

masaaania:
- This PR modifies the ListAccountsV2 API so that we can get the accounts by environment ID.
  - This function is required to use accountV2 in the current management console.


cre8ivejp (dismissed):
Thank you!

Issue #308, opened 2023-05-08
feat: Progressive rollout


cre8ivejp:
### Describe the feature you'd like

This feature allows the user to progressively roll out a feature flag by scheduling it.
Basically, the user will schedule the rollout from 1 to 100% of a feature flag, and Bucketeer will change the rollout percentage based on the user's configuration.

PR #712, opened 2024-01-22
chore: change conditions when creating a progressive rollout

cre8ivejp:
Part of https://github.com/bucketeer-io/bucketeer/issues/308

Currently, it takes a lot of work to create a progressive rollout. To create, you must have everything on the targeting tab in the initial state. Also, you can't create auto operations or update the flag while a progressive rollout is in progress.
The user should be able to use the progressive rollout with the kill switch.

Since the progressive rollout only changes the default strategy rollout percentage and doesn't affect other services, I'm removing the unnecessary conditions to simplify the creation process.

#### Things done

- Remove the following conditions when creating a progressive rollout
  - if the flag is disabled
  - if the flag has a prerequisite
  - if the flag has an individual targeting
  - if the flag has rules
 
- Remove the following conditions when creating a Schedule or Kill Switch
  - check if there is a progressive rollout in waiting or progress state

- Added the following conditions when updating a feature flag
  - check if it has a progressive rollout when changing the default strategy
  - check if it has a progressive rollout when adding or removing a variation
  
 #### Next steps

- Enable the flag once when the progressive rollout starts if the flag is still disabled
- Implement the `StopProgressiveRollout` API to stop the rollout when the flag state changes to disabled
- Stop the progressive rollout when the flag is disabled by the user by the schedule or by the kill switch
- Remove the unnecessary conditions on the console when creating the progressive rollout
- Disable the Default Strategy when a progressive rollout is running on the console


pkg/feature/api/feature.go
```diff
@@ -875,7 +875,7 @@ func (s *FeatureService) existsRunningProgressiveRollout(
 	ctx context.Context,
 	featureID, environmentNamespace string,
 ) (bool, error) {
-	progressiveRollouts, err := s.listProgressiveRollouts(ctx, environmentNamespace, featureID)
+	progressiveRollouts, err := s.listProgressiveRollouts(ctx, featureID, environmentNamespace)
```
cre8ivejp, line 878: There was a bug here. I also added a unit test to ensure it's working.
I'm checking the values passed to the API.

https://github.com/bucketeer-io/bucketeer/pull/712/files#diff-fad41f54a190634386c37788c6f063b9d12ddb75b1928675e5b6ddd61b0d9663R3114-R3119

kentakozuka (approved):
LGTM 🎉

PR #716, opened 2024-01-23
ci: fix push-image triggering when pushing ci code

cre8ivejp:
Fixes https://github.com/bucketeer-io/bucketeer/issues/717

We didn't have the directory `.devcontainer` when this workflow was created. When we made changes to the `.devcontainer`, it triggered only the `push-image` workflow, causing errors on the PipeCD because the helm chart doesn't exist, which is expected. So, I'm fixing the condition to match the `push_chart` workflow.


masaaania (approved):
Thank you!!

PR #718, opened 2024-01-23
fix: progress rollout fails due to wrong conditions

cre8ivejp:
Part of https://github.com/bucketeer-io/bucketeer/issues/308

The progressive rollout fails while trying to update the flag default strategy due to the condition implemented [here](https://github.com/bucketeer-io/bucketeer/blob/main/pkg/feature/api/validation.go#L1158-L1160).
Because the user can't update the default strategy when the progress rollout is in progress, we must know from where the request is coming so we can handle it properly.

#### Things done

- Fixed rollout failing when calling the `UpdateFeatureTargeting` API due to wrong validation by adding the field `from` to the `UpdateFeatureTargeting` API so we can handle the request coming from
- Removed unnecessary implementation when the progress rollout is triggered (Removing flag rules)
- Added the `from` field to the feature module in the `ui/web-v2`
- Updated unit and e2e tests


pkg/autoops/api/progressive_rollout_operation.go
```diff
@@ -161,31 +157,7 @@ func updateFeatureTargeting(
 	return nil
 }
 
-func getResetFeatureCmds(
```
cre8ivejp, line 164: I'm removing this implementation because the progressive rollout must only change the default strategy.

ui/web-v2/apps/admin/src/modules/features.ts
```diff
@@ -288,6 +288,7 @@ export const updateFeatureTargeting = createAsyncThunk<
   request.setId(params.id);
   request.setComment(params.comment);
   request.setCommandsList(params.commands);
+  request.setFrom(UpdateFeatureTargetingRequest.From.USER);
```
cre8ivejp, line 291: All requests from the console and e2e test must have set the `from` field as `USER`.

pkg/autoops/api/progressive_rollout_operation.go
```diff
@@ -138,21 +138,17 @@ func updateFeatureTargeting(
...
-	}
-	cmds = append(cmds, c)
 	_, err = featureClient.UpdateFeatureTargeting(
 		ctx,
 		&featureproto.UpdateFeatureTargetingRequest{
 			EnvironmentNamespace: environmentNamespace,
 			Id:                   feature.Id,
-			Commands:             cmds,
+			Commands:             []*featureproto.Command{c},
+			From:                 featureproto.UpdateFeatureTargetingRequest_OPS,
```
cre8ivejp, line 151: All requests from the progressive rollout must have set the `from` field as `OPS`.

kentakozuka:
Just I want to ask;
How to stop progressive rollout manually?

cre8ivejp:
> Just I want to ask; How to stop progressive rollout manually?

@kentakozuka, the user will be able to stop it on the console.

masaaania (approved):
lgtm!

PR #720, opened 2024-01-23
refactor: move auto ops queries to sql files

cre8ivejp:
Fixes https://github.com/bucketeer-io/bucketeer/issues/719

#### Things done

- Moved the SQL query hard code to the SQL files in the auto ops service


masaaania (approved):
Thank you!

PR #724, opened 2024-01-23
chore: update account v2 api for compatibility

masaaania:
## Summary

- Update account V2 APIs (Create and Update).
- CreateAccountV2
  - Fixed to update environment roles when trying to create an existing account.
  - This fix is a temporary implementation until migration to the new console design.
- UpdateAccountV2
  - Added to the `WriteType(override or patch)` field to the `ChangeAccountV2EnvironmentRolesCommand` so that we can update the environment roles easily.
  - If override is selected, overwrite with the passed Environment Roles; if patch is selected, append to or partially overwrite existing Environment Roles.


cre8ivejp (approved):
👍

PR #725, opened 2024-01-24
feat: add GetMyOrganizationsByEmail API

masaaania:
## Summary

- `GetMyOrganizationsByEmail` API was added.
  - This API will be used by the other microservice instead of the console.
  - Especially, the auth service will use it in the authentication process.


cre8ivejp (dismissed):
Thank you!

PR #726, opened 2024-01-24
chore: update the process getting AdminAccount

masaaania:
## Summary

- The old getAdminAccount() function retrieves objects from the `admin_account` table.
- The new getAdminAccountV2() function retrieves the account belonging to the System Admin Organization.


cre8ivejp (approved):
👍

PR #727, opened 2024-01-24
chore: update generate token process

masaaania:
## Summary
- Change to use the GetMyOrganizationsByEmail() function instead of the old GetMeV2ByEmail() function when issuing IDToken.
- This change allows the GetMeV2ByEmail() function to be removed.


cre8ivejp (approved):
Thank you!

Issue #722, opened 2024-01-23
chore: enable the search feature for the filter options in the search filter


cre8ivejp:
Currently, the user can only select the tag from the filter options, but when there are too many options, we need to scroll down to find the tag we want, which takes time. So, it would be nice to have the search feature in the filter options.

PR #728, opened 2024-01-24
chore(ui): enable the search feature for the filter options in the search filter

bimalgrg519:
Fixes https://github.com/bucketeer-io/bucketeer/issues/722


cre8ivejp (approved):
Thank you!

PR #729, opened 2024-01-25
chore: remove webhook table from delete-e2e targets

masaaania:
## Summary

- We've already deleted the webhook table.


cre8ivejp (approved):
Sorry. I forgot that. Thank you!

PR #743, opened 2024-01-29
fix: unable create trial project

masaaania:
## Summary

- I replaced the old account API with the new API while creating a new org, project, and account.


cre8ivejp (approved):
Thanks!

PR #744, opened 2024-01-29
chore: remove old account api

masaaania:


cre8ivejp (approved):
Good bye V1!
Thank you 🎉

PR #755, opened 2024-01-31
fix: invalid mysql connection when using transaction

cre8ivejp:
Fixes https://github.com/bucketeer-io/bucketeer/issues/754

This fixes the account E2E tests failing randomly.


kentakozuka (approved):
lgtm 👍

PR #775, opened 2024-02-01
fix: omitempty not being removed in the proto subdirectories

cre8ivejp:
There are some proto directories with subdirectories. Those didn't match the condition and weren't removed.

#### Things done

- Fixed the remove code
- Updated the proto subdirectories


kentakozuka (approved):
👍

PR #776, opened 2024-02-01
chore: add system admin flag to idtoken

kentakozuka:
part of https://github.com/ca-dp/bucketeer-issue/issues/5

This pull request primarily focuses on refactoring the token generation process in the `hack/generate-service-token/command.go` file and updating the `IDToken` struct in the `pkg/token/idtoken.go` file. The key changes include replacing the `ioutil.WriteFile` function with `os.WriteFile`, adding an `IsSystemAdmin` field to the `IDToken` struct, and updating the token generation process to set `IsSystemAdmin` to `true`.

Here are the main changes:

* `hack/generate-service-token/command.go`: 
  * Replaced the `ioutil.WriteFile` function with `os.WriteFile` to write the signed ID token to a file. This change is in line with the deprecation of `ioutil` in Go 1.16. [[1]](diffhunk://#diff-cb32802d4e8e7750abbbc9e43742a3694197cb014d853d99a1724dc9cde38941L20-R20) [[2]](diffhunk://#diff-cb32802d4e8e7750abbbc9e43742a3694197cb014d853d99a1724dc9cde38941R78-R85)
  * Updated the token generation process to set `IsSystemAdmin` to `true` when creating a new ID token.

* `pkg/token/idtoken.go`: 
  * Added a new `IsSystemAdmin` field to the `IDToken` struct. This field is intended to replace the `AdminRole` field, which is now deprecated and will be removed in the future.

### DIFF
--- a/hack/generate-service-token/command.go
+++ b/hack/generate-service-token/command.go
@@ -17,7 +17,7 @@ package main
 import (
 	"context"
 	"errors"
-	"io/ioutil"
+	"os"
 	"time"
 
 	"go.uber.org/zap"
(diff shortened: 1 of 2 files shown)


cre8ivejp (approved):
Thank you!

PR #778, opened 2024-02-02
refactor: use system admin word correctly

kentakozuka:
Now we have two admins;
- system admin
- organizational admin

They should be explicitly used correctly to avoid confusion.

### DIFF
--- a/pkg/account/storage/v2/admin_account.go
+++ b/pkg/account/storage/v2/admin_account.go
@@ -34,7 +34,7 @@ var (
 	selectAdminAccountV2SQL string
 )
 
-func (s *accountStorage) GetAdminAccountV2(ctx context.Context, email string) (*domain.AccountV2, error) {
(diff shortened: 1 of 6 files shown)


pkg/account/storage/v2/admin_account.go
```diff
@@ -34,7 +34,7 @@ var (
 	selectAdminAccountV2SQL string
```
cre8ivejp, line 34: How about renaming the SQL file and the constant, too?

```suggestion
        //go:embed sql/account_v2/select_system_admin_account_v2.sql
        selectSystemAdminAccountV2SQL string
```

kentakozuka: I'm fixing in the next PR. There are some other places to be fixed too.

cre8ivejp (approved):
👍

PR #787, opened 2024-02-03
chore: remove tx field in accountStorage

Ubisoft-potato:
We don't need `tx` field in `accountStorage`, we can remove it.

```go
type accountStorage struct {
	client mysql.Client
	tx     mysql.Transaction
}
```


cre8ivejp (approved):
Thank you!

Issue #795, opened 2024-02-06
System admin does not have access to resources


kentakozuka:
### Version

main

### Describe the bug

`generateToken` should return `is_system_admin: true` if the account is a system admin.
https://github.com/bucketeer-io/bucketeer/blob/62b04f247b129a1ca573f379adb244f6156050a4/pkg/auth/api/api.go#L390-L397

PR #796, opened 2024-02-06
fix: system admin does not have access to resources

kentakozuka:
Fixes https://github.com/bucketeer-io/bucketeer/issues/795


cre8ivejp (approved):
Thank you!

PR #817, opened 2024-02-14
fix: stopped_by value being reported incorrectly

cre8ivejp:
The `stoppedBy` is being reported as `OPS_KILL_SWITCH` when the flag is disabled by a Scheduled operation.


kentakozuka (approved):
lgtm 👍

PR #832, opened 2024-02-20
fix: event persister subscribing to pubsub when running scheduled operations

cre8ivejp:
Because the Scheduled operations are not based on events, they must not be included in the query results. Otherwise, the `event-persister-evaluation-events-ops` will subscribe to PubSUb, pulling messages unnecessarily.

Because we no longer support multiple clauses in the same rule, I will refactor the proto structure so we can add a new column to know the operations, simplifying the query, too.

### Query  result

Screenshot 2024-02-20 at 12 57 45 PM


kentakozuka (approved):
LGTM 👍

PR #836, opened 2024-02-20
fix(ui): admin update pages using using url code instead of id

cre8ivejp:
The update page must use the ID instead of the URL code.


kentakozuka (approved):
LGTM 👍

PR #837, opened 2024-02-21
fix(ui): open update overlay when click on the url link

bimalgrg519:
Fixes https://github.com/bucketeer-io/bucketeer/issues/838


cre8ivejp (approved):
Thanks!

PR #852, opened 2024-02-29
fix: pubsub subscription check stops working when an error occurs

cre8ivejp:
Many places are using `continue` when an error occurs. By placing the reset interval in the last when an error occurs, it stops checking whether to subscribe or unsubscribe to PubSub.

This also could be related to flaky e2e tests using the event persisters.


Ubisoft-potato:
It could stop subscribing if it execute `continue` before reset the `Timer`.

This PR will fix this issue. LGTM 👍!

PR #853, opened 2024-02-29
fix(ui): kill switch operation's start status is incorrect

### LINKED ISSUE #847: fix(ui): kill switch operation's start status is incorrect
It should be FROM ON -> OFF, not ON -> OFF.

Screenshot 2024-02-28 at 4 21 17 PM

bimalgrg519:
fix #847


cre8ivejp (approved):
Thank you!

PR #854, opened 2024-02-29
test: fix flaky e2e test in the api-gateway

cre8ivejp:
Depending on the time this test runs, there might not be any feature flag in the cache
So, we wait until the caches are updated. Usually, one minute.

### DIFF
--- a/test/e2e/gateway/api_grpc_test.go
+++ b/test/e2e/gateway/api_grpc_test.go
@@ -353,7 +353,11 @@ func TestGrpcGetEvaluationsWithPreviousEvaluation31daysAgo(t *testing.T) {
 	uuid := newUUID(t)
 	userID := newUserID(t, uuid)
 	prevEvalAt := time.Now().Add(-31 * 24 * time.Hour).Unix()
-	response := grpcGetEvaluationsByEvaluatedAt(t, "", userID, "userEvaluationsID", prevEvalAt, false)
+	tag := fmt.Sprintf("%s-tag-%s", prefixTestName, uuid)
+	featureID := newFeatureID(t, uuid)
+	createFeatureWithTag(t, tag, featureID)
+	updateFeatueFlagCache(t)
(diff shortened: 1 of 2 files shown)


test/e2e/gateway/api_grpc_test.go
```diff
@@ -353,7 +353,11 @@ func TestGrpcGetEvaluationsWithPreviousEvaluation31daysAgo(t *testing.T) {
 	uuid := newUUID(t)
 	userID := newUserID(t, uuid)
 	prevEvalAt := time.Now().Add(-31 * 24 * time.Hour).Unix()
-	response := grpcGetEvaluationsByEvaluatedAt(t, "", userID, "userEvaluationsID", prevEvalAt, false)
+	tag := fmt.Sprintf("%s-tag-%s", prefixTestName, uuid)
+	featureID := newFeatureID(t, uuid)
+	createFeatureWithTag(t, tag, featureID)
+	updateFeatueFlagCache(t)
+	response := grpcGetEvaluationsByEvaluatedAt(t, tag, userID, "userEvaluationsID", prevEvalAt, false)
```
cre8ivejp, line 360: Because the cache can be empty depending on the time the test runs, we must ensure that there is at least one flag.

kentakozuka:
@cre8ivejp which line does actually fix the flaky tests?

cre8ivejp:
> @cre8ivejp which line does actually fix the flaky tests?

@kentakozuka, the `forceUpdate` must be true. Also, I left a comment [here](https://github.com/bucketeer-io/bucketeer/pull/854#discussion_r1510564445).

https://github.com/bucketeer-io/bucketeer/blob/6da157a991fcd8a9fb9f1d4cd816947d83bde4c1/test/e2e/gateway/api_grpc_test.go#L364

kentakozuka (approved):
LGTM 👍

PR #868, opened 2024-03-01
refactor: remove incorrect comment

kentakozuka:


pkg/role/role.go
```diff
@@ -55,20 +55,19 @@ func CheckEnvironmentRole(
...
-	// TODO remove this condition after migration to AccountV2
-	if !token.IsSystemAdmin {
-		// get account for the environment namespace
-		account, err := getAccountFunc(token.Email)
-		if err != nil {
-			if code := status.Code(err); code == codes.NotFound {
-				return nil, ErrUnauthenticated
-			}
-			return nil, ErrInternal
+	if token.IsSystemAdmin {
```
kentakozuka, line 58: This condition is still necessary, is it?

cre8ivejp: I think it is the same here.
https://github.com/bucketeer-io/bucketeer/blob/23da5c06eff4734f5f57fec5747c70b2029a16a6/pkg/role/role.go#L105-L110

It probably returns immediately if it is a system admin to speed up the authentication process. Otherwise, it would make a request to the account service.

cre8ivejp (approved):
👍

PR #870, opened 2024-03-06
test: fix retry logic in the event counter tests

cre8ivejp:
There was a bug in the retry logic, making it retry less than it was supposed to. In that case, it was 30 times.
Also, I made some improvements around the timeout and the retry logic.
Now, after retrying, it won't fail. Instead, it will return the error the caller handles and proceed accordingly.

### DIFF
--- a/test/e2e/eventcounter/eventcounter_test.go
+++ b/test/e2e/eventcounter/eventcounter_test.go
@@ -2196,11 +2196,12 @@ func getExperimentResult(t *testing.T, c ecclient.Client, experimentID string) (
 			break
 		}
 		st, _ := status.FromError(err)
-		if st.Code() != codes.Unavailable {
-			return nil, err
+		if st.Code() == codes.Unavailable || st.Code() == codes.Internal {
+			fmt.Printf("Failed to get experiment result. Experiment ID: %s. Error code: %d. Retrying in 5 seconds.\n", experimentID, st.Code())
+			time.Sleep(5 * time.Second)
+			continue
 		}
-		fmt.Printf("Failed to get experiment result. Experiment ID: %s. Error code: %d. Retrying in 5 seconds.\n", experimentID, st.Code())
-		time.Sleep(5 * time.Second)
+		return nil, err
 	}
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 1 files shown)


test/e2e/eventcounter/eventcounter_test.go
```diff
@@ -537,7 +591,17 @@ func TestExperimentResult(t *testing.T) {
 			t.Fatalf("retry timeout")
 		}
 		time.Sleep(10 * time.Second)
-		resp := getExperimentResult(t, ecClient, experiment.Id)
+
+		resp, err := getExperimentResult(t, ecClient, experiment.Id)
+		if err != nil {
+			st, _ := status.FromError(err)
+			if st.Code() != codes.NotFound {
```
cre8ivejp, line 598: We will only retry if the experiment is not ready.

test/e2e/eventcounter/eventcounter_test.go
```diff
@@ -2099,67 +2291,84 @@ func getFeature(t *testing.T, client featureclient.Client, featureID string) *fe
...
+	for i := 0; i < numRetries; i++ {
+		response, err = c.GetEvaluationTimeseriesCount(
+			ctx,
+			req,
+		)
+		if err == nil {
+			break
+		}
+		st, _ := status.FromError(err)
+		if st.Code() != codes.Unavailable {
```
cre8ivejp, line 2336: It only retries for all the requests if the server is unavailable due to pods rescheduling, etc.

test/e2e/eventcounter/eventcounter_test.go
```diff
@@ -53,8 +53,8 @@ import (
 
 const (
 	prefixTestName = "e2e-test"
-	timeout        = 20 * time.Second
-	retryTimes     = 30
+	timeout        = 60 * time.Second
```
cre8ivejp, line 56: During the tests, I saw a few times a context deadline was exceeded while inserting data in BigQuery. So, I changed it to 60 seconds as the dev environment is configured.

test/e2e/eventcounter/eventcounter_test.go
```diff
@@ -2001,26 +2189,28 @@ func getExperimentResult(t *testing.T, c ecclient.Client, experimentID string) *
...
+	numRetries := 3
 	for i := 0; i < numRetries; i++ {
 		response, err = c.GetExperimentResult(ctx, req)
 		if err == nil {
 			break
 		}
 		st, _ := status.FromError(err)
-		if st.Code() == codes.Internal {
-			t.Fatal(err)
+		if st.Code() != codes.Unavailable {
```
Ubisoft-potato, line 2199: I think we can also retry if the `st.Code()` is **Internal**. For example, we can return `InternalServerError` in the `GetExperimentResult` [function](https://github.com/bucketeer-io/bucketeer/blob/main/pkg/eventcounter/api/api.go#L761-L764), and this `InternalServerError` could also caused by downstream unavailable.

So this condition can be `st.Code() != codes.Unavailable && st.Code() != codes.Internal`.

`codes.Internal` Doc:
```golang
        // Internal errors. Means some invariants expected by underlying
	// system has been broken. If you see one of these errors,
	// something is very broken.
	//
	// This error code will be generated by the gRPC framework in several
	// internal error conditions.
        Internal Code = 13

```

Ubisoft-potato:
LGTM👍!

Issue #866, opened 2024-03-01
feat: be able to make the comment feature optional


cre8ivejp:
By default, we require the user to write a comment on every change he makes to the Feature Flags, which is good when the environment is running in production and you want to ensure that all changes are logged in detail.
But, for the development stage, it is a bit annoying to write comments that are unnecessary all the time.

We can add a setting in the Environment page so the user can decide whether or not to require the comment when making changes.

As a reference:
https://github.com/bucketeer-io/bucketeer/discussions/848

PR #875, opened 2024-03-12
chore: add environment settings proto

cre8ivejp:
Part of https://github.com/bucketeer-io/bucketeer/issues/866


proto/environment/environment.proto
```diff
@@ -18,6 +18,10 @@ package bucketeer.environment;
 option go_package = "github.com/bucketeer-io/bucketeer/proto/environment";
 
 message EnvironmentV2 {
+  message Settings {
+    bool require_comment = 1;
+  }
```
kentakozuka, line 23: Why is this nested?

cre8ivejp: To maintain consistency with other proto messages.
Also, this settings message belongs to Environment only. I don't see a specific reason not to be in the Environment message.

Reference.
https://github.com/bucketeer-io/bucketeer/blob/main/proto/account/account.proto#L42-L69

kentakozuka: I wanted to ask why `Settings` is necessary. We can simply add `require_comment` to `EnvironmentV2`.
If `Settings` is introduced, `name` and `description` can be also settings.

cre8ivejp: We might need to add more settings in the new UI or if some users ask us. So it will be easier to manage it when we add more settings.

> If Settings is introduced, name and description can be also settings.

Name and description are general information, not exactly a setting IMHO. It's used for reference only different from the settings that will be used in validations, etc.

kentakozuka (approved):
LGTM 👍

PR #876, opened 2024-03-14
chore: use go-sql-driver in the mysql client

cre8ivejp:
I'm not sure why we are using the `go-sql-driver` to convert the errors, but we are not using the driver in the client. I think we forgot to import it.

Because the `go-sql-driver` is also lightweight and faster than the standard drivers, I'm implementing the use of it.
The e2e tests also have passed.

Reference.
https://github.com/bucketeer-io/bucketeer/blob/main/pkg/storage/v2/mysql/error.go#L21

### DIFF
--- a/pkg/storage/v2/mysql/client.go
+++ b/pkg/storage/v2/mysql/client.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"time"
 
+	_ "github.com/go-sql-driver/mysql"
 	"go.uber.org/zap"
 
 	"github.com/bucketeer-io/bucketeer/pkg/metrics"
@@ -29,6 +30,7 @@ import (
 const dsnParams = "collation=utf8mb4_bin"
 
 type options struct {
+	connMaxLifetime time.Duration
 	connMaxIdleTime time.Duration
 	maxOpenConns    int
 	maxIdleConns    int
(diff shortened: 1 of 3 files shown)


pkg/storage/v2/mysql/client.go
```diff
@@ -128,6 +137,7 @@ func NewClient(
 		logger.Error("Failed to open db", zap.Error(err))
 		return nil, err
 	}
+	db.SetConnMaxLifetime(dopts.connMaxLifetime)
```
cre8ivejp, line 140: I'm adding the setting because it is required to avoid issues where the SQL server could close the connection before the driver.

https://github.com/go-sql-driver/mysql?tab=readme-ov-file#important-settings

pkg/storage/v2/mysql/client.go
```diff
@@ -38,15 +40,22 @@ type options struct {
 
 func defaultOptions() *options {
 	return &options{
-		connMaxIdleTime: 300 * time.Second,
+		connMaxLifetime: 10 * time.Minute,
```
cre8ivejp, line 43: Because the `connMaxIdleTime` is supposed to be less than the `connMaxLifetime`, I implemented the default value to 10 minutes.

Ubisoft-potato:
LGTM! 👍

PR #880, opened 2024-03-20
feat: multi subscriber arch in batch

Ubisoft-potato:


pkg/batch/subscriber/processor/domain_event_informer.go
```diff
@@ -44,161 +40,31 @@ var (
...
-	runningDurationPerBatch      time.Duration
-	project                      string
-	domainSubscription           string
-	domainTopic                  string
-	pullerNumGoroutines          int
-	pullerMaxOutstandingMessages int
-	pullerMaxOutstandingBytes    int
-	opts                         *options
-	logger                       *zap.Logger
+type DomainEventInformer struct {
```
cre8ivejp, line 43: How about creating an interface to return `Process(msg *puller.Message)` and make the struct private?
Wdyt?

Ubisoft-potato: Great idea! Let me implement this change.

pkg/batch/subscriber/components.go
cre8ivejp, line 1: nit: `components.go` -> `subscriber.go`

Ubisoft-potato: Done.

pkg/batch/cmd/server/server.go
```diff
@@ -560,6 +531,62 @@ func (s *server) createMySQLClient(
...
+			name, config, p,
+			subscriber.WithLogger(logger),
+		))
+	}
+	multiSubscriber.Start(ctx)
+	return multiSubscriber
+}
+
+func (s *server) registerProcessorMap(
+	ctx context.Context,
```
cre8ivejp, line 575: We can remove the context since it's being used.

Ubisoft-potato: Done.

pkg/batch/cmd/server/server.go
```diff
@@ -560,6 +530,61 @@ func (s *server) createMySQLClient(
...
+		subscriber.WithLogger(logger),
+	)
+	for name, config := range configMap {
+		p, err := processors.GetProcessorByName(name)
+		if err != nil {
+			logger.Fatal("subscriber: processor not found",
+				zap.String("name", name),
+				zap.Error(err),
+			)
+			return nil
```
cre8ivejp, line 562: It would be safer if we return an error here. Otherwise, we won't notice that the subscription is not working when the name doesn't match the processor name.

pkg/batch/cmd/server/server.go
```diff
@@ -560,6 +530,61 @@ func (s *server) createMySQLClient(
...
 }
 
+func (s *server) startMultiPubSub(
+	ctx context.Context,
+	processors *processor.Processors,
+	logger *zap.Logger,
+) *subscriber.MultiSubscriber {
+	bytes, err := os.ReadFile(*s.subscriberConfig)
+	if err != nil {
+		logger.Fatal("subscriber: failed to read subscriber config",
```
cre8ivejp, line 540: Please use `logger.Error` instead.
Same for all other places.

manifests/bucketeer/charts/batch/values.yaml
```diff
@@ -155,3 +145,13 @@ cronjob:
     - name: auto-ops-rules-cacher
       jobId: AutoOpsRulesCacher
       schedule: "* * * * *"
+
+subscribers:
+  domainEventInformer:
```
cre8ivejp, line 150: Please add a comment here.

```go
// This is the processor's name. It must match the same name defined in the
// pkg/batch/subscriber/processor/processors.go
```

cre8ivejp (approved):
Thank you!

PR #928, opened 2024-04-23
chore: change cronjob error code condition to ignore 503 and 000 errors

cre8ivejp:
Because the jobs run every minute, if they fail due to network or unavailable errors, we can ignore them because they are temporary.

000 = Network issues (Host not found, etc)
503 = Can happen when the pods are being terminated


Ubisoft-potato (approved):
LGTM! We should disregard network errors when invoking the `batch` job API.

Issue #921, opened 2024-04-15
feat: evaluate the user in the go sdk to improve response latency


cre8ivejp:
## Summary

The current architecture when the get variation interface is called. It requests the evaluation directly from the server.
This approach is more straightforward but may introduce latency. Also, this could lead to infra-cost issues because it could evaluate the same user many times in a short time, even if the flag didn't change.

We should consider implementing the evaluation logic in the SDK to implement a cache similar to the client SDK.
This would improve the latency issues and drastically reduce the requests to the server, reducing the infra-cost.

I was thinking of creating a new common repository for the evaluation and proto files, but it would be better if we kept this as a mono repo for this case. We can move it to another repository in the future if needed.

## TODO

- [x] Create the proto dir as a go module
- [x] Move the evaluation code to a go module
- [x] Implement evaluation module in the API Gateway
- [x] Implement server-side API key type (Backend & Frontend)
- [x] Implement API to get all flags and segment users per environment using the server-side API key
- [x] Implement cache in the Go SDK (In-memory)
- [x] Implement the configuration to select to evaluate locally or remotely
- [x] Update documentation

## API Gateway

It will evaluate the user from the evaluation module. There are no changes in the evaluation code.

## API Key

We will implement a new key type, which will have access to all the flags and segment users per environment.
I will add a new column called `key_type` in the `api_key` table to check what the key can retrieve.

- **Server-side**:  Access to All flags and segment users per environment (Read-only)
- **Client-side**: Access to a single user's evaluation (Read-only)

## SDK cache

The cacher will cache all the flags and segment users per environment in memory.
It will poll the latest data from the Bucketeer server and override the current in-memory data every 30 seconds.
This implementation will be similar to the other client SDKs.

## SDK get variation interface

We won't change the current interfaces. The changes will be made internally to evaluate the user in the SDK, just like the Bucketeer app does on the server side.

PR #931, opened 2024-04-24
chore: implement created_at in the feature and segment users cache

cre8ivejp:
Part of https://github.com/bucketeer-io/bucketeer/issues/921

To check if the cache in the SDK is the latest, we need something to compare with the server cache.
For features, we will generate an ID based on all the feature IDs and versions.
We will also use the segment's `UpdatedAt` value for the segment users' cache.


pkg/batch/jobs/cacher/featureflagcacher.go
```diff
@@ -71,7 +72,11 @@ func (c *featureFlagCacher) Run(ctx context.Context) error {
 			c.logger.Error("Failed to list features", zap.String("environmentId", env.Id))
 			return err
 		}
-		if err := c.cache.Put(&ftproto.Features{Features: features}, env.Id); err != nil {
+		fts := &ftproto.Features{
+			Id:       evaluation.GenerateFeaturesID(features),
```
cre8ivejp, line 76: Here, we generate an ID based on all the features per environment.
The same ID will be returned to the SDK so it can be checked on the server when the SDK requests the features from the cache.

pkg/batch/jobs/cacher/segmentusercacher.go
```diff
@@ -85,6 +85,7 @@ func (c *segmentUserCacher) Run(ctx context.Context) error {
 			su := &ftproto.SegmentUsers{
 				SegmentId: seg.Id,
 				Users:     users,
+				UpdatedAt: seg.UpdatedAt,
```
cre8ivejp, line 88: Here, we set the segment's `UpdatedAt`.
The same value will be returned to the SDK so it can be checked on the server when the SDK requests the segment users from the cache.

evaluation/user_evaluations.go
```diff
@@ -69,3 +69,14 @@ func sortMapKeys(data map[string]string) []string {
 	sort.Strings(keys)
 	return keys
 }
+
+func GenerateFeaturesID(features []*ftproto.Feature) string {
```
cre8ivejp, line 73: I'm using the same logic used to generate the [UserEvaluationsID](https://github.com/bucketeer-io/bucketeer/blob/main/evaluation/user_evaluations.go#L47).

Ubisoft-potato (commented):
The comparison and check logic will be implemented in another PR, right?

cre8ivejp:
> The comparison and check logic will be implemented in another PR, right?

@Ubisoft-potato, yes. When implementing the API to get the flags and the segment users, I will check those values.

Ubisoft-potato:
LGTM! 👍

PR #934, opened 2024-05-07
chore: change to use redis to cache environment api key

cre8ivejp:
Fixes https://github.com/bucketeer-io/bucketeer/issues/935

We have been experiencing instability issues when the in-memory cache is gone and trying to get the environment API from MySQL. Since the Memorystore has a higher SLA than CloudSQL, I have changed it to Redis.
The batch service already updates the environment API key cache every minute.

This will increase the response latency by a few milliseconds, but it will bring more stability.

### DIFF
--- a/pkg/cache/v3/environment_api_key.go
+++ b/pkg/cache/v3/environment_api_key.go
@@ -26,9 +26,8 @@ import (
 )
 
 const (
-	environmentAPIKeyKind             = "environment_apikey"
-	environmentAPIKeyTTL              = 1 * time.Minute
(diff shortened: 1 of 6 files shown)


Ubisoft-potato:
> This will increase the response latency by a few milliseconds, but it will bring more instability.

instability -> stability?

cre8ivejp:
> > This will increase the response latency by a few milliseconds, but it will bring more instability.
> 
> I think instability -> stability?

Yes. I updated the description.

Ubisoft-potato (approved):
LGTM! 🚀

Issue #885, opened 2024-03-22
feat: maintenance window operation


cre8ivejp:
Currently, we support scheduled operations. But there are a few cases where you want to show a feature (survey, campaign) or hide it for maintenance. We need to create 2 operations (ON/OFF) for those cases. This feature will allow users to configure multiple schedules in one operation, improving the UX and management.

PR #937, opened 2024-05-08
docs: add rfc for multi schedule

kakcy:
This PR adds RPC documentation for implementing the multi schedule. : #885

### DIFF
--- a/docs/rfcs/multi-schedules.md
+++ b/docs/rfcs/multi-schedules.md
@@ -1,0 +1,74 @@
+# Summary
+
+The Multi Schedule function allows you to register multiple schedules at once.
+It is managed together like a staged "progressive rollout".
+
+## Function
+### Create
+It is possible to specify ON or OFF for multiple dates at the same time.
+* Validation
+  * Dates must be set in ascending order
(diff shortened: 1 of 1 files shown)


docs/rfcs/multi-schedules.md
```diff
@@ -0,0 +1,80 @@
...
+
+### Delete
+It is possible to delete uncompleted schedules.
+* Validation
+  * Schedules that have already been completed cannot be deleted.
+
+### Stop
+It is possible to stop a schedule that has already started.
+* Validation
+  * Schedules that have already been completed cannot be stopped.
```
kentakozuka, line 30: Do you think that stop is necessary?
We can just delete a schedule instead of stopping it.

cre8ivejp: On the operations page, there are two tabs: one for active operations and the other for finished or stopped operations.

When using multiple schedules, the user might want to stop the operation to keep the history of the other schedules executed earlier.
The Progressive Rollout has the same behavior.

kentakozuka (approved):
Thanks!

PR #939, opened 2024-05-08
test: add server sdk api key flag to e2e tests

cre8ivejp:
Part of https://github.com/bucketeer-io/bucketeer/issues/921

To add E2E tests for the new Server SDK, we will need to use a different API Key because it has a different role.
So, I'm changing it to pass the `api-key-server` flag.


Makefile
```diff
@@ -218,7 +218,7 @@ create-api-key:
 		--web-gateway=${WEB_GATEWAY_URL}:443 \
 		--service-token=${SERVICE_TOKEN_PATH} \
 		--name=$$(date +%s) \
-		--role=SDK_CLIENT \
+		--role=${API_KEY_ROLE} \
```
cre8ivejp, line 221: The E2E test will pass this value to create client and server API keys.

Ubisoft-potato (approved):
LGTM! 👍

PR #943, opened 2024-05-09
fix: nil pointer when logging

Ubisoft-potato:
This pull request will fix the issue of the `logger` field not being set, which has caused the user event persister to stop subscription.


cre8ivejp (approved):
Thank you!

PR #945, opened 2024-05-10
chore: add mysql open connection configuration to batch server

cre8ivejp:
After we moved the user persister to the batch service, many requests got stuck when upserting the values in the DB. It was getting context deadline errors.
The reason was the MySQL client, by default, uses 10 connections, and the user persister uses 10 workers.
Also, there are other services that use the same client instance, which causes delays in the operations in MySQL, causing errors.

For now, I'm setting the default value to 50 connections. But we should be careful when moving or adding new services that use MySQL in the future.


Ubisoft-potato (approved):
LGTM! 👍

PR #953, opened 2024-05-21
fix: get segment users api condition

cre8ivejp:


pkg/gateway/api/api_grpc.go
```diff
@@ -864,7 +864,7 @@ func (s *grpcGatewayService) GetSegmentUsers(
 	updatedSegments := make([]*featureproto.SegmentUsers, 0, len(targetSegmentUsers))
 	adjustedRequestedAt := req.RequestedAt - secondsForAdjustment
 	for _, su := range targetSegmentUsers {
-		if su.UpdatedAt < adjustedRequestedAt {
+		if su.UpdatedAt > adjustedRequestedAt {
```
cre8ivejp, line 867: The `UpdatedAt` must be higher than the `requestedAt`, so I can return only the updated segments.

Ubisoft-potato (approved):
LGTM, thank you!

PR #954, opened 2024-05-22
chore: change to allow server api role in the register events api

cre8ivejp:
Part of #921 

The register events API must also accept requests using the `SERVER_SDK` API role so the new server SDK can report events correctly.

### DIFF
--- a/pkg/gateway/api/api_grpc.go
+++ b/pkg/gateway/api/api_grpc.go
@@ -306,7 +306,7 @@ func (s *grpcGatewayService) GetEvaluations(
 ) (*gwproto.GetEvaluationsResponse, error) {
 	ctx, span := trace.StartSpan(ctx, "bucketeerGRPCGatewayService.GetEvaluations")
 	defer span.End()
-	envAPIKey, err := s.checkRequest(ctx, accountproto.APIKey_SDK_CLIENT)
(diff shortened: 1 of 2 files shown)


pkg/gateway/api/api_grpc.go
```diff
@@ -1567,6 +1563,15 @@ func checkEnvironmentAPIKey(environmentAPIKey *accountproto.EnvironmentAPIKey, r
 	return nil
 }
 
+func contains[T comparable](s []T, e T) bool {
```
cre8ivejp, line 1566: Go supports generics from 1.18. Because other implementations use the contains function, I unify it using generics.

pkg/gateway/api/api_grpc.go
```diff
@@ -1554,8 +1546,12 @@ func getEnvironmentAPIKeyFromCache(
 	return nil, err
 }
 
-func checkEnvironmentAPIKey(environmentAPIKey *accountproto.EnvironmentAPIKey, role accountproto.APIKey_Role) error {
-	if environmentAPIKey.ApiKey.Role != role {
+func checkEnvironmentAPIKey(
+	environmentAPIKey *accountproto.EnvironmentAPIKey,
+	roles []accountproto.APIKey_Role,
```
cre8ivejp, line 1551: I changed here to support multiple roles.

Ubisoft-potato (approved):
LGTM! 👍

PR #967, opened 2024-05-30
fix: evaluation fails

kentakozuka:
fixes https://github.com/bucketeer-io/bucketeer/issues/968

### DIFF
--- a/evaluation/evaluation.go
+++ b/evaluation/evaluation.go
@@ -19,6 +19,8 @@ import (
 	"fmt"
 	"time"
 
+	"golang.org/x/exp/maps"
+
 	ftproto "github.com/bucketeer-io/bucketeer/proto/feature"
 	userproto "github.com/bucketeer-io/bucketeer/proto/user"
 )
@@ -58,7 +60,6 @@ func NewEvaluator() *evaluator {
 	return &evaluator{}
 }
 
-// Deprecated: use EvaluateFeaturesByEvaluatedAt instead.
 // This function will be removed once all the SDK clients are updated.
 func (e *evaluator) EvaluateFeatures(
 	fs []*ftproto.Feature,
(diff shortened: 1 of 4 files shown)


evaluation/evaluation_test.go
```diff
@@ -712,8 +770,11 @@ func TestEvaluateFeaturesByEvaluatedAt(t *testing.T) {
 				p.tag,
 			)
 			assert.Equal(t, p.expectedError, err)
+			if p.expectedError != nil {
+				return
+			}
 			assert.Equal(t, p.expectedEvals.UserEvaluations.ForceUpdate, actual.ForceUpdate)
-			assert.Equal(t, p.expectedEvals.UserEvaluations.ArchivedFeatureIds, actual.ArchivedFeatureIds)
+			assert.ElementsMatch(t, p.expectedEvals.UserEvaluations.ArchivedFeatureIds, actual.ArchivedFeatureIds)
```
kentakozuka, line 777: Fixes a flaky test.

evaluation/evaluation.go
```diff
@@ -101,8 +102,10 @@ func (e *evaluator) EvaluateFeaturesByEvaluatedAt(
 	if len(updatedFeatures) == 0 {
 		return e.evaluate(fs, user, mapSegmentUsers, true, targetTag)
 	}
-	featuresHavePrerequisite := e.getFeaturesHavePrerequisite(fs)
-	evalTargets := e.getPrerequisiteUpwards(updatedFeatures, featuresHavePrerequisite)
+	evalTargets, err := e.getEvalFeatures(updatedFeatures, fs)
```
cre8ivejp, line 105: We need to add this before calling `evaluate` here, too.

https://github.com/bucketeer-io/bucketeer/blob/16f9c2fc693eba6749ce753cab8bf78a44b475c8/evaluation/evaluation.go#L70

cre8ivejp: The `GetEvaluation` API call this [getTargetFeatures](https://github.com/bucketeer-io/bucketeer/blob/16f9c2fc693eba6749ce753cab8bf78a44b475c8/pkg/gateway/api/api_grpc.go#L924-L934), which calls the `GetPrerequisiteDownwards`.

cre8ivejp: For this case, don't we also need to call the `getFeaturesDependedOnTargets`?

kentakozuka: It is not necessary to be added here because `fs` includes all features to be evaluated 👍

kentakozuka: I modified `GetPrerequisiteDownwards()` to call `getFeaturesDependedByTargets()`.

cre8ivejp (approved):
Thank you!

PR #969, opened 2024-05-30
chore(proto): add get feature flags and segment users to api ID

cre8ivejp:
Part of #921 

I'm adding new API IDs for the metric events that will be used in the Go server SDK.


Ubisoft-potato (dismissed):
LGTM 👍

proto/event/client/event.proto
```diff
@@ -94,6 +94,9 @@ enum ApiId {
   GET_EVALUATION = 1;
   GET_EVALUATIONS = 2;
   REGISTER_EVENTS = 3;
+  GET_FEATURE_FLAGS = 4;
+  GET_SEGMENT_USERS = 5;
+  SDK_GET_VARIATION = 100;
```
cre8ivejp, line 99: I'm adding an ID to report the evaluation latency when running locally on the server SDK.
To keep things clean, we will use the `SDK` as a prefix for all new IDs referencing the SDK internal metrics.

kentakozuka (approved):
Nice 👍

Issue #963, opened 2024-05-29
feat: implement prerequisite condition in the custom rule


cre8ivejp:
Currently, when using the Prerequisite feature, the user is limited to 3 variations when testing Flag B.
Ideally, the user should be able to work with Prerequisite without any variation number restrictions.

We could achieve this without restrictions if we also implement the Prerequisite in the custom rule.

### Now

Flag A | Flag B | Assigning
--- | --- | ---
A | A | 25%
 　 | B | 25% 
B | B (Off variation) | 50%

### Ideal

Flag A | Flag B | Assigning
--- | --- | ---
A | A | 25%
 　 | B | 25% 
B | A | 25%
 　 | B | 25%

PR #983, opened 2024-06-03
feat: evaluate flag dependency of targeting rule

kentakozuka:
part of https://github.com/bucketeer-io/bucketeer/issues/963

### DIFF
--- a/evaluation/feature_dependency_evaluator.go
+++ b/evaluation/feature_dependency_evaluator.go
@@ -1,0 +1,34 @@
+// Copyright 2024 The Bucketeer Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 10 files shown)


evaluation/feature_dependency_evaluator.go
```diff
@@ -0,0 +1,34 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package evaluation
+
+type featureDependencyEvaluator struct {
+}
+
+func (e *featureDependencyEvaluator) Evaluate(
```
cre8ivejp: Because the feature name is the Prerequisite, using the same name will be better to avoid confusion.
 Same for the file name. E.g. `prerequisite_evaluator.go`
 
 ```suggestion
func (e *prerequisiteEvaluator) Evaluate(
```

cre8ivejp: Please fix the other file names.

kentakozuka: IMHO, it is different from prerequisite.
Prerequisite is the setting to check before the entire targeting, while this is one of the targeting rules.
WDYT?

cre8ivejp: Agree.

cre8ivejp (approved):
Thank you!

PR #952, opened 2024-05-20
feat: move on-demand subscriber to batch service

Ubisoft-potato:
Part of https://github.com/bucketeer-io/bucketeer/issues/405

This PR will upgrade multi PubSub architecture to support on-demand subscriber: 

- event-persister-evaluation-events-dwh
- event-persister-evaluation-events-ops
- event-persister-goal-events-dwh
- event-persister-goal-events-ops

---
**Key Changes**

- Add on-demand subscriber architecture. [[1]](https://github.com/bucketeer-io/bucketeer/pull/952/files#diff-20026b354097df62938c1ca4fc1c9ac8b10397d6ba01a5f278568f2ea1aea9eb) [[2]](https://github.com/bucketeer-io/bucketeer/pull/952/files#diff-6e1312d8eb4c761a6bed38a73030372d365031c859b3de0eb331c01633f01e47R61-R65)
- Manage on-demand subscriber json configuration seperately. [[1]](https://github.com/bucketeer-io/bucketeer/pull/952/files#diff-da142715fb447fa09cd1b8854c0f9caa732a8d7d0555e25ae2bd0d8cb52d4b4dR207-R247) [[2]](https://github.com/bucketeer-io/bucketeer/pull/952/files#diff-da142715fb447fa09cd1b8854c0f9caa732a8d7d0555e25ae2bd0d8cb52d4b4dR267-R291) [[3]](https://github.com/bucketeer-io/bucketeer/pull/952/files#diff-186ceb07184707bd9312a6f564dde8b86c7025ad9a3fd7d74e2688860f64c12a)
- Register on-demand subscribers. [[1]](https://github.com/bucketeer-io/bucketeer/pull/952/files#diff-9f3bffbf9b659d0228f7f25aeb18a14d9fa79a72e2a8100e7371b1d112faa00cR766-R831) [[2]](https://github.com/bucketeer-io/bucketeer/pull/952/files#diff-9f3bffbf9b659d0228f7f25aeb18a14d9fa79a72e2a8100e7371b1d112faa00cR702-R715)

**Refactor event persisters**

- [Event dwh persister](https://github.com/bucketeer-io/bucketeer/pull/952/files#diff-7f9de3b99af1557121449dacd5e2bfc2ddc055006382d22c8ba3a0fa0df0fb40)
[...]

PR #992, opened 2024-06-05
chore: ignore configuration not found error

Ubisoft-potato:
This PR will ignore the batch subscriber json config not found error rather than crash batch service while user is not configure subscriber json config.

### DIFF
--- a/manifests/bucketeer/charts/batch/templates/deployment.yaml
+++ b/manifests/bucketeer/charts/batch/templates/deployment.yaml
@@ -79,13 +79,13 @@ spec:
               value: "{{ .Values.env.experimentCalculatorService }}"
             - name: BUCKETEER_BATCH_BATCH_SERVICE
               value: "{{ .Values.env.batchService }}"
-            - name: BUCKETEER_OPS_EVENT_SCHEDULE_COUNT_WATCHER
+            - name: BUCKETEER_BATCH_EVENT_SCHEDULE_COUNT_WATCHER
               value: "{{ .Values.env.scheduleCountWatcher }}"
-            - name: BUCKETEER_OPS_EVENT_SCHEDULE_DATETIME_WATCHER
+            - name: BUCKETEER_BATCH_EVENT_SCHEDULE_DATETIME_WATCHER
               value: "{{ .Values.env.scheduleDatetimeWatcher }}"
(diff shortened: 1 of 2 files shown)


manifests/bucketeer/charts/batch/templates/deployment.yaml
```diff
@@ -79,13 +79,13 @@ spec:
               value: "{{ .Values.env.experimentCalculatorService }}"
             - name: BUCKETEER_BATCH_BATCH_SERVICE
               value: "{{ .Values.env.batchService }}"
-            - name: BUCKETEER_OPS_EVENT_SCHEDULE_COUNT_WATCHER
+            - name: BUCKETEER_BATCH_EVENT_SCHEDULE_COUNT_WATCHER
```
Ubisoft-potato, line 82: This env value name was wrong, fix it here.

pkg/batch/cmd/server/server.go
```diff
@@ -594,37 +594,36 @@ func (s *server) startMultiPubSub(
...
-			subscriber.WithLogger(logger),
-		))
+		for name, config := range configMap {
+			p, err := processors.GetProcessorByName(name)
+			if err != nil {
+				logger.Error("subscriber: processor not found",
+					zap.String("name", name),
+					zap.Error(err),
+				)
+				return nil, err
```
cre8ivejp, line 618: We should also ignore it here.
It will fail if we add a new configuration, but it's not in the processors.
The idea is not to process persisters who don't have the configuration, or it's not specified in the processors.

Ubisoft-potato: Typically, when we want to add a subscriber then we will add the subscriber config first. At the same time, the corresponding processor must be implemented and add in processor config.

Conclusion: all the subscribers `configMap`'s subscriber must have a processor, otherwise, we shouldn't start the batch service, because we or developer made a mistake that forgot to implement or add processor for the subscriber.

cre8ivejp: If the container is deployed before the config map is updated, it will crash, and vice-versa.
We can't deploy both at the same time. So, this will be an issue, especially running canary releases and scaling out the pods simultaneously.

Ubisoft-potato: In canary releases, it make sense.
Since we have this change in PR #952, I will update it there and close this PR.

Issue #904, opened 2024-04-04
feat: public APIs


cre8ivejp:
To be able to manage the flags and other features outside of the console.

kentakozuka:
# What we want

- Giving up command
- Diff b/w versions
- We can still go with gRPC
    - Simplifying APIs is necessary in any case (OpenAPI／gRPC).
- Start from Feature object

# milestones

1. Simplify gRPC API
    1. Start from Feature
        1. Create new APIs
            1. Create/Get/Update/Delete
    2. Give up commands
2. (if necessary) implement HTTP with OpenAPI
3. Do the same for all entities

kentakozuka:
## System design

```mermaid
sequenceDiagram
    3rdParty->>+WebGateway: GetFeature (gPRC)
    WebGateway->>+Backend (Public API): GetFeature (gRPC)
    Backend (Public API)->>+Feature: GetFeature()
    3rdParty->>+WebGateway: GET (REST)
    WebGateway->>+Backend (Public API): GetFeature (gPRC)
    Backend (Public API)->>+Feature: GetFeature()
```

kentakozuka:
todos
1. Create a new API w/o command
2. Create a new API in GTW
3. Replace an old API with a new API in UI
4. Deprecate/Remove an old API

hvn2k1:
Check list:
- [x] push
- [x] notification
- [x] account
- [x] api key
- [x] tag
- [ ] feature
- [ ] user segment
- [ ] auto ops
- [ ] flag trigger
- [ ] audit log
- [ ] environment
- [ ] event counter
- [ ] experiment

PR #998, opened 2024-06-07
chore: add api key roles for public api

kentakozuka:


proto/account/api_key.proto
```diff
@@ -23,6 +23,8 @@ message APIKey {
     UNKNOWN = 0;
     SDK_CLIENT = 1;
     SDK_SERVER = 2;
+    API_READ_ONLY = 3;
+    API_READ_WRITE = 4;
```
kentakozuka, line 27: A good prefix did not come out 😢 
Do you folks have any idea?

cre8ivejp, line 27: I think we should have at least 3 roles for public APIs, just like the account roles.
How about this?

```suggestion
    PUBLIC_API_READ_ONLY = 3;
    PUBLIC_API_WRITE = 4;
    PUBLIC_API_ADMIN = 5; // For sensitive data
```

or

```suggestion
    OPEN_API_READ_ONLY = 3;
    OPEN_API_WRITE = 4;
    OPEN_API_ADMIN = 5; // For sensitive data
```

Ubisoft-potato: I think that utilizing `OPEN_API_` would be more appropriate. 

Many open-source projects that support REST API access adhere to the [OpenAPI Specification](https://swagger.io/specification/).

kentakozuka: I prefer just `API_XXX` in these three options because it already has the same sense as the other two.
OpenAPI explains how we implement the API. We need to know what implementation we use. It could be gRPC or another. It's unlikely, but it could also be possible that we will add an implementation of GraphQL to meet the user's demands.

cre8ivejp: Only API for me it's abstract.
It could be anything.
Server API, SDK API, Client API.
If the implementation it's not decided, I still prefer PUBLIC as prefix because it's not related to the implementation.
Or if you guys have a better idea.

kentakozuka: OK, let's go with `PUBLIC_API_` 🚀

kentakozuka:
@cre8ivejp Fixed 👍

cre8ivejp (approved):
👍

PR #1000, opened 2024-06-10
chore: proto-all command does everything

kentakozuka:


cre8ivejp (approved):
nit: `chore: change proto-all to run all the proto commands`

PR #1003, opened 2024-06-10
ci: add mockgen job

kentakozuka:


.github/workflows/pr-go.yaml
```diff
@@ -18,7 +18,7 @@ env:
   GOLANGCI_LINT: v1.51.2
   PROTOLOCK_VERSION: v0.16.0
   PROTOC_GEN_GO_VERSION: v1.5.2
-  MOCKGEN_VERSION: v1.6.0
+  MOCKGEN_VERSION: v0.1.0
```
cre8ivejp, line 21: Is there a reason not to use the latest version 0.4.0?

kentakozuka: The version is specified here.
https://github.com/bucketeer-io/bucketeer/blob/411fafe37d19ba086485acbeb3e240b81aad6cdc/Makefile#L50

We need to upgrade it anyway, but this is not the scope of this PR.
Let me do that in another PR 👍

cre8ivejp (approved):
👍

PR #1006, opened 2024-06-11
ci: update tools

kentakozuka:


.golangci.yml
```diff
@@ -8,18 +8,12 @@ linters-settings:
...
     - gosimple
     - govet
     - ineffassign
-    # Since we got the following error, staticcheck is disabled. After solving it, we'll enable it.
-    # pkg/account/apikeycacher/apikeycacher.go:284:12: SA1019: ptypes.UnmarshalAny is deprecated: Call the any.UnmarshalTo method instead. (staticcheck)
-    # - staticcheck
-    - structcheck
     - typecheck
     - unused
-    - varcheck
```
kentakozuka, line 22: Removed deprecated linters

.github/workflows/pr-go.yaml
```diff
@@ -152,6 +160,6 @@ jobs:
           go-version: ${{ env.GO_VERSION }}
           cache: true
       - name: Install golangci-lint
-        run: go install github.com/golangci/golangci-lint/cmd/golangci-lint@${{ env.GOLANGCI_LINT }}
+        run: go install github.com/golangci/golangci-lint/cmd/golangci-lint@latest
```
cre8ivejp, line 163: Is there a reason to use the latest version?
This could introduce instabilities to the CI.

kentakozuka: I set latest here as goimport.
https://github.com/bucketeer-io/bucketeer/blob/c2a78137cf9bdbcd8dd5b01442b7e21b2a622d39/.github/workflows/pr-go.yaml#L65

Also, golangci-lint does not affect codes while mockgen and protoc do.

cre8ivejp: I asked because last year, an older version broke the CI due to new linters, deprecation, etc.

cre8ivejp (approved):
👍

PR #1014, opened 2024-06-12
ci: set clang-format version

cre8ivejp:
The `ubuntu-latest` uses clang-format 14, which is very old. 
Sometimes, it fails due to a newer version running on the local machine when developing.
To avoid this, I'm setting a specific version to avoid errors.


kakcy (approved):
Thank you 👍

PR #1015, opened 2024-06-12
fix(ui): segment cannot be set in the targeting rule

kentakozuka:


cre8ivejp (approved):
👍 

nit: fix(ui): segment cannot be set in the targeting rule

Issue #405, opened 2023-06-12
chore: modular monolithic architecture


masaaania:
## Summary
Bucketeer chose a microservices architecture in the early stages of development, but as the number of services grew, the following problems emerged:

- Increasing infra costs
- Increasing learning costs for contribution and self-host

These disadvantages could intimidate future users from adopting Bucketeer.

While the benefits gained by microservices architecture are also hard to dismiss.
Microservices architecture has allowed us to implement properly separated services quickly.

We think the modular monolithic architecture will allow us to enjoy the benefits of microservices architecture while reducing its disadvantages.

![Architecture202302 (1)](https://github.com/bucketeer-io/bucketeer/assets/2755429/1e2a4993-bb89-4b54-a42c-28c1aa909db8)

## TODO

- [x] backend service integration
- [x] dex
- [x] web
- [x] consumer service integration
- [x] improve implementation and remove services
  - [x] feature-recorder
  - [x] feature-segment-persister
  - [x] account-apikey-cacher
- [x] replace old sdk and remove services
  - [x] metrics-event-persister  

## Status

We have integrated eleven microservices into a service named `backend` with the following PRs:
- https://github.com/bucketeer-io/bucketeer/pull/404
- https://github.com/bucketeer-io/bucketeer/pull/416
- https://github.com/bucketeer-io/bucketeer/pull/419
- https://github.com/bucketeer-io/bucketeer/pull/422
- https://github.com/bucketeer-io/bucketeer/pull/425
- https://github.com/bucketeer-io/bucketeer/pull/426
- https://github.com/bucketeer-io/bucketeer/pull/427
- https://github.com/bucketeer-io/bucketeer/pull/429
- https://github.com/bucketeer-io/bucketeer/pull/430
- https://github.com/bucketeer-io/bucketeer/pull/431
- https://github.com/bucketeer-io/bucketeer/pull/432

![Architecture202307](https://github.com/bucketeer-io/bucketeer/assets/2755429/e0063fc1-4117-43f6-be39-a33e828dc8b0)

PR #1020, opened 2024-06-13
feat: move metrics event to batch service

Ubisoft-potato:
Part of #405 

This PR will move metrics event persister to batch service and delete old metrics event persister code and manifest files.


cre8ivejp (approved):
Thank you!

PR #1027, opened 2024-06-17
feat(ui): add api key options of public api to form pages

kentakozuka:


cre8ivejp:
The PR looks good, but shouldn't we merge this after the public API implementation is done?
Otherwise, it will confuse the user when creating an API Key.

kentakozuka:
@cre8ivejp 
IMHO, it is ok to ship it because it's only merging the PR, not released with the specific version. Even though some users use main branch, `GetFeature` is already in main (and it is what main branch is running).

cre8ivejp:
> @cre8ivejp IMHO, it is ok to ship it because it's only merging the PR, not released with the specific version. Even though some users use main branch, `GetFeature` is already in main (and it is what main branch is running).

What I meant was that the current users who access the console. They can create an API Key, which might confuse them into thinking that we already support it because there is no documentation.
When I developed the Server SDK cache, I left the changes to the console in the last step of implementation after everything was tested and the documentation was published.

kentakozuka:
@cre8ivejp 
For me, merging it to the main is okay unless it has backward compatibility(This does not mean releasing a version). But I understand your opinion. I'll merge it after everything is ready to go.

kentakozuka:
@cre8ivejp It's ready to go 🚀

cre8ivejp (approved):
Thank you!

PR #1032, opened 2024-06-18
ci: update devcontainer building conditions

cre8ivejp:
I'm configuring the dev container building conditions to build it automatically so we don't forget.

### DIFF
--- a/.github/workflows/devcontainer-image-build.yaml
+++ b/.github/workflows/devcontainer-image-build.yaml
@@ -2,6 +2,11 @@ name: Bucketeer Dev Container Build and Push Image
 
 on:
   workflow_dispatch:
+  push:
+    branches:
+      - main
+    paths:
+      - ".github/.devcontainer"
 
 jobs:
   build-and-push:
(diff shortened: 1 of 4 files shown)


.github/workflows/devcontainer-image-build.yaml
```diff
@@ -2,6 +2,11 @@ name: Bucketeer Dev Container Build and Push Image
 
 on:
   workflow_dispatch:
+  push:
+    branches:
+      - main
+    paths:
+      - ".github/.devcontainer"
```
cre8ivejp, line 9: Any changes in this directory will trigger the workflow when merging to the main branch.

cre8ivejp: If the other files in this directory don't affect the docker image when they change, we can restrict the trigger to the `Dockerfile` file only.

Ubisoft-potato: The other files in this directory is the key point to change, `.github/.devcontainer` is fine.

Generally we need to add tools in `devcontainer.json` then build the image.

Ubisoft-potato (commented):
👍, LGTM!

Ubisoft-potato (approved):

PR #1039, opened 2024-06-20
fix(ui): incorrect and unmet peer dependency warnings

kentakozuka:
Fixes incorrect peer dependencies.

1. removes `resolutions` section
2. fixes incorrect peer dependencies.

before
```sh
warning " > @material-ui/core@4.11.1" has incorrect peer dependency "react@^16.8.0".
warning " > @material-ui/core@4.11.1" has incorrect peer dependency "react-dom@^16.8.0".
warning "@material-ui/core > @material-ui/styles@4.11.5" has incorrect peer dependency "react@^16.8.0 || ^17.0.0".
warning "@material-ui/core > @material-ui/styles@4.11.5" has incorrect peer dependency "react-dom@^16.8.0 || ^17.0.0".
warning "@material-ui/core > @material-ui/system@4.12.2" has incorrect peer dependency "react@^16.8.0 || ^17.0.0".
warning "@material-ui/core > @material-ui/system@4.12.2" has incorrect peer dependency "react-dom@^16.8.0 || ^17.0.0".
warning "@material-ui/core > @material-ui/utils@4.11.3" has incorrect peer dependency "react@^16.8.0 || ^17.0.0".
warning "@material-ui/core > @material-ui/utils@4.11.3" has incorrect peer dependency "react-dom@^16.8.0 || ^17.0.0".
warning " > @material-ui/icons@4.11.3" has incorrect peer dependency "react@^16.8.0 || ^17.0.0".
warning " > @material-ui/icons@4.11.3" has incorrect peer dependency "react-dom@^16.8.0 || ^17.0.0".
warning " > @material-ui/lab@4.0.0-alpha.56" has incorrect peer dependency "react@^16.8.0".
warning " > @material-ui/lab@4.0.0-alpha.56" has incorrect peer dependency "react-dom@^16.8.0".
warning " > @material-ui/pickers@3.2.10" has unmet peer dependency "@date-io/core@^1.3.6".
warning " > @material-ui/pickers@3.2.10" has unmet peer dependency "prop-types@^15.6.0".
warning " > @material-ui/pickers@3.2.10" has incorrect peer dependency "react@^16.8.4".
warning " > @material-ui/pickers@3.2.10" has incorrect peer dependency "react-dom@^16.8.4".
warning " > react-chartjs-2@2.11.1" has incorrect peer dependency "react@^0.14.0 || ^15.0.0 || ^16.0.0 || ^17.0.0".
[...]

PR #1031, opened 2024-06-17
refactor: remove nx 

kentakozuka:
Fixes https://github.com/bucketeer-io/bucketeer/issues/1026

- Remove Nx from the build process
- Remove unused files
  - babel
  - jest
- Update lint/format
  - eslint
  - prettier
- Move files from `src/app/admin/` to `src/`
- Add `.env` file for development (See [diff of README.md](https://github.com/bucketeer-io/bucketeer/pull/1031/files#diff-3a4eeffd219895c01de8021a5b6a48d2604401429429829b2661d2479e670e48))
- Set up devcontainer

PR #1040, opened 2024-06-20
ci: update npm packages with dependabot

kentakozuka:
The UI's packages were recently cleaned up (https://github.com/bucketeer-io/bucketeer/pull/1031, https://github.com/bucketeer-io/bucketeer/pull/1039), so I suppose they're ready to be updated with Dependabot.


cre8ivejp (approved):
👍

PR #1045, opened 2024-06-20
chore: enable eslint recommended plugins 

kentakozuka:
- Removes unused files
- Adds recommended eslint settings.
- Fixes errors that eslint automatically can fix.
- Treats errors as warnings (because they are too many to fix all in this PR)


cre8ivejp (approved):
Nice!

PR #1047, opened 2024-06-21
chore: add generate all command to makefile

kentakozuka:
For convenience...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,6 +37,9 @@ docker-run:
 # Go
 #############################
 
+.PHONY: generate
+generate: proto-all mockgen
+
 .PHONY: check-all
 check-all: proto-check mockgen update-repos diff-check lint build-go test-go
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -37,6 +37,9 @@ docker-run:
 # Go
 #############################
 
+.PHONY: generate
+generate: proto-all mockgen
```
cre8ivejp, line 41: The rule name is too abstract. It can be anything.
Please specify what it generates.

kentakozuka: That's what I expect this command to run all generate commands.
Do you have any idea?

cre8ivejp: If you expect to generate all the commands, how about `generate-all` or `generate-all-commands`?

kentakozuka: `generate-all` sounds look to me 👍

kentakozuka:
@cre8ivejp Fixed 👍

cre8ivejp (approved):
👍

Issue #804, opened 2024-02-06
feat: be able to check the audit log details


cre8ivejp:
Currently, we can check who made the change, what entity changed, and what time it changed, but we can't see what values changed. It's impossible to track the changes, and there is no `diff` feature to compare the before/after changes, which makes it hard to manage the logs.

kentakozuka:
Related https://github.com/bucketeer-io/bucketeer/issues/904
To add public API, we need to simplify web API, which fixes this issue, too.

kentakozuka:
- [x] Feature
  - [ ] remove any commands
- [ ] Segment
- [ ] ...

PR #1054, opened 2024-06-23
chore: run push sender when feature updated event is created

kentakozuka:
part of https://github.com/bucketeer-io/bucketeer/issues/804


cre8ivejp (approved):
👍 

nit: `chore: run push sender when feature updated event is created`

PR #1059, opened 2024-06-24
chore: update feature returns updated object

kentakozuka:

The update method should return the updated object because the user wants to see what is updated and use it for further operations.


cre8ivejp (approved):
👍

PR #1071, opened 2024-06-26
chore: remove github token

Ubisoft-potato:
Since we already merged new database migration, we can remove the github token dependency of old migration.


cre8ivejp (approved):
Thank you!

PR #1053, opened 2024-06-21
chore: enable typescript-eslint/no-explicit-any

kentakozuka:
part of https://github.com/bucketeer-io/bucketeer/issues/1017

PR #1072, opened 2024-06-26
fix: variation value input is now shown in add page

kentakozuka:
This bug comes from [this PR](https://github.com/bucketeer-io/bucketeer/pull/1053) because [VariationInput](https://github.com/bucketeer-io/bucketeer/blob/main/ui/web-v2/src/components/VariationInput/index.tsx) is used by both add and update form.

The above PR introduced form being typed by update form schema, which is not working add form.
Instead of reverting the PR to bring non-type codes, I create a new variation input for the add form for type safe.


cre8ivejp (approved):
Thank you!

PR #1074, opened 2024-06-26
chore: add tags,enabled,archived to update feature api

kentakozuka:


Makefile
```diff
@@ -26,7 +26,7 @@ LDFLAGS_BUILDDATE := $(LDFLAGS_PACKAGE).BuildDate
 #############################
 
 .PHONY: generate-all
-generate: proto-all mockgen
+generate-all: proto-all mockgen
```
kentakozuka, line 29: fix

pkg/feature/domain/feature.go
```diff
@@ -953,6 +962,38 @@ func (f *Feature) Update(
...
+				return nil, err
+			}
+			incVersion = true
+		} else {
+			if err := updated.Disable(); err != nil {
+				return nil, err
+			}
+			incVersion = true
+		}
+	}
```
cre8ivejp, line 983:
```suggestion
	if enabled != nil {
		if enabled.Value {
			if err := updated.Enable(); err != nil {
				return nil, err
			}
		} else {
			if err := updated.Disable(); err != nil {
				return nil, err
			}
		}
		incVersion = true
	}
```

pkg/feature/domain/feature.go
```diff
@@ -953,6 +962,38 @@ func (f *Feature) Update(
...
+				return nil, err
+			}
+			incVersion = true
+		} else {
+			if err := updated.Unarchive(); err != nil {
+				return nil, err
+			}
+			incVersion = true
+		}
+	}
```
cre8ivejp, line 996:
```suggestion
	if archived != nil {
		if archived.Value {
			if err := updated.Archive(); err != nil {
				return nil, err
			}
		} else {
			if err := updated.Unarchive(); err != nil {
				return nil, err
			}
		}
		incVersion = true
	}
```

cre8ivejp (commented):
I  left a few comments. The rest looks good!

kentakozuka:
@cre8ivejp Thanks!
Fixed 👍

cre8ivejp (approved):
Thank you!

PR #1075, opened 2024-06-27
chore: move atlas installation to dockerfile

cre8ivejp:


Makefile
```diff
@@ -287,9 +286,16 @@ check-apply-migration:
...
+.PHONY: build-devcontainer
 build-devcontainer:
 	devcontainer build --workspace-folder=.github --push=false --image-name="ghcr.io/bucketeer-io/bucketeer-devcontainer:latest"
 
+.PHONY: push-devcontainer
+push-devcontainer:
+	@echo $(PAT) | docker login ghcr.io -u $(GITHUB_USER_NAME) --password-stdin
+	docker tag ghcr.io/bucketeer-io/bucketeer-devcontainer:latest ghcr.io/bucketeer-io/bucketeer-devcontainer:latest
+	docker push ghcr.io/bucketeer-io/bucketeer-devcontainer:latest
+
```
kentakozuka, line 298: Do we need to push the image?
We can run a container that is locally built.

Makefile
```diff
@@ -287,9 +286,16 @@ check-apply-migration:
 #############################
 
 # build devcontainer locally
+.PHONY: build-devcontainer
 build-devcontainer:
 	devcontainer build --workspace-folder=.github --push=false --image-name="ghcr.io/bucketeer-io/bucketeer-devcontainer:latest"
 
+.PHONY: push-devcontainer
+push-devcontainer:
```
Ubisoft-potato, line 294: I think we shouldn't push our local dev container image to ghcr.
Because we only build our local machine arch image, it'll influence other user.

For local testing purpose, we don't need to push it to ghcr, it can use local image directly.

cre8ivejp: I have removed it.

Makefile
cre8ivejp: I have removed it.

Ubisoft-potato (commented):
LGTM!

Ubisoft-potato (approved):

PR #1080, opened 2024-06-29
fix: proto clone changs empty slice to nil

kentakozuka:
proto built-in Clone() changes an empty slice to nil, which creates unexpected diffs.

Screenshot 2024-06-29 at 13 12 20


kakcy (approved):
LGTM 👍

PR #1095, opened 2024-07-04
feat: fix Update API for Multi Schedule

kakcy:
In this pull request, I modified the AutoOps Update API to support Multi Schedule.

Part of https://github.com/bucketeer-io/bucketeer/issues/885.

Key Changes:
- Fixed to reject update requests if AutoOps is stopped or completed.
- The current UI requests a Clause update without giving ActionType, so it has been modified to add ActionType according to OpsType.

Also added some validations:
- Update to a different OpsType
- Requests where the number of operations becomes 0 after updating


pkg/autoops/api/api.go
```diff
@@ -632,6 +632,55 @@ func (s *AutoOpsService) UpdateAutoOpsRule(
...
+			})
+			if err != nil {
+				return statusInternal.Err()
+			}
+			return dt.Err()
+		}
+
+		// Changes to a different opsType are not allowed
+		if req.ChangeAutoOpsRuleOpsTypeCommand != nil &&
+			req.ChangeAutoOpsRuleOpsTypeCommand.OpsType != autoOpsRule.OpsType {
```
kentakozuka, line 671: Should it be the following?
`ChangeAutoOpsRuleOpsTypeCommand` is deprecated, so the command should not be used anymore, shouldn't it?

```suggestion
		if req.ChangeAutoOpsRuleOpsTypeCommand != nil {
```

pkg/autoops/api/api.go
```diff
@@ -632,6 +632,55 @@ func (s *AutoOpsService) UpdateAutoOpsRule(
...
+			})
+			if err != nil {
+				return statusInternal.Err()
+			}
+			return dt.Err()
+		}
+
 		if req.ChangeAutoOpsRuleOpsTypeCommand != nil {
 			if req.ChangeAutoOpsRuleOpsTypeCommand.OpsType == autoopsproto.OpsType_ENABLE_FEATURE &&
 				len(req.AddOpsEventRateClauseCommands) > 0 {
```
kentakozuka, line 686: If the comment above is right, these lines should be removed.

pkg/autoops/api/api.go
```diff
@@ -825,19 +889,50 @@ func (s *AutoOpsService) isNoUpdateAutoOpsRuleCommand(req *autoopsproto.UpdateAu
...
 	}
+
 	for _, c := range req.AddOpsEventRateClauseCommands {
+		if opsType != nil {
+			if *opsType == autoopsproto.OpsType_DISABLE_FEATURE {
+				c.OpsEventRateClause.ActionType = autoopsproto.ActionType_DISABLE
+			} else if *opsType == autoopsproto.OpsType_ENABLE_FEATURE {
+				c.OpsEventRateClause.ActionType = autoopsproto.ActionType_ENABLE
+			}
+		}
```
kentakozuka, line 905: I'm not sure what this part is trying to do.
`ChangeAutoOpsRuleOpsTypeCommand` is deprecated but `req.ChangeAutoOpsRuleOpsTypeCommand.OpsType` is referenced.

pkg/autoops/api/api.go
kakcy: @kentakozuka 
Currently, it seems that this command is sent even if the OpsType is not changed when updating from the management screen.
https://github.com/bucketeer-io/bucketeer/blob/cc1466b8607a15323c6d06a829ef77dc2d4ebd8c/ui/web-v2/src/components/OperationAddUpdateForm/index.tsx#L255
Therefore, if you fix it, AutoOps will probably not be able to update.
How about not making the correction in this PR, but making the correction after the correction is made on the management screen?

pkg/autoops/api/api.go
kakcy: To support multi-scheduling, Enable and Disable have been changed from being managed by OpsType to being managed by ActionType held within Clause.

As mentioned in this comment, if you change the management screen, this code will no longer be needed and will be deleted.

kakcy:
I added the necessary comments and revised some processing in [this commit](https://github.com/bucketeer-io/bucketeer/pull/1095/commits/fc2210d4c4a061d6b9774e424b33b1ddf4f1044f).

pkg/autoops/api/api.go
```diff
@@ -879,19 +949,55 @@ func (s *AutoOpsService) isNoUpdateAutoOpsRuleCommand(req *autoopsproto.UpdateAu
...
+
+	// The current web console uses ChangeAutoOpsRuleOpsTypeCommand to update OpsTypes and enable or disable them.
+	// So ActionType is not set and you need to update ActionType according to OpsType.
+	// ToDo: If ChangeAutoOpsRuleOpsTypeCommand is no longer used in the web console,
+	// the following code is no longer needed and should be removed.
+	var opsType *autoopsproto.OpsType = nil
 	if req.ChangeAutoOpsRuleOpsTypeCommand != nil {
 		commands = append(commands, req.ChangeAutoOpsRuleOpsTypeCommand)
+		opsType = &req.ChangeAutoOpsRuleOpsTypeCommand.OpsType
 	}
```
kentakozuka, line 961: How about the following?
```suggestion
	actionType := autoopsproto.ActionType_UNKNOWN
	if req.ChangeAutoOpsRuleOpsTypeCommand != nil {
		commands = append(commands, req.ChangeAutoOpsRuleOpsTypeCommand)
		if req.ChangeAutoOpsRuleOpsTypeCommand.OpsType == autoopsproto.OpsType_DISABLE_FEATURE {
			actionType = autoopsproto.ActionType_DISABLE
		} else if req.ChangeAutoOpsRuleOpsTypeCommand.OpsType == autoopsproto.OpsType_ENABLE_FEATURE {
			actionType = autoopsproto.ActionType_DISABLE
		}
	}
```

Then use `actionType` for ActionType of each commands. We can remove if block in each for loop bellow.

pkg/autoops/api/api.go
```diff
@@ -879,19 +949,55 @@ func (s *AutoOpsService) isNoUpdateAutoOpsRuleCommand(req *autoopsproto.UpdateAu
...
+
 	for _, c := range req.AddOpsEventRateClauseCommands {
+		if c.OpsEventRateClause.ActionType == autoopsproto.ActionType_UNKNOWN {
+			if opsType != nil && *opsType == autoopsproto.OpsType_DISABLE_FEATURE {
+				c.OpsEventRateClause.ActionType = autoopsproto.ActionType_DISABLE
+			}
```
kentakozuka, line 967: What if opsType is autoopsproto.OpsType_ENABLE_FEATURE? c.OpsEventRateClause.ActionType can remain unknown, so should it return an error?

kakcy: Validation is performed below.
https://github.com/bucketeer-io/bucketeer/blob/main/pkg/autoops/api/api.go#L395
The reason why we haven't validated Unknown is because the console doesn't support ActionType, so if we validate it, an error will occur, so we can't do it. Added ToDo comment.
https://github.com/bucketeer-io/bucketeer/pull/1095/commits/6fde42bc4c1879ff2e55f5002cea395130ab9778

pkg/autoops/api/api.go
```diff
@@ -879,19 +951,43 @@ func (s *AutoOpsService) isNoUpdateAutoOpsRuleCommand(req *autoopsproto.UpdateAu
...
+			c.OpsEventRateClause.ActionType = actionType
+		}
 		commands = append(commands, c)
 	}
 	for _, c := range req.AddDatetimeClauseCommands {
-		commands = append(commands, c)
+		if c.DatetimeClause.ActionType == autoopsproto.ActionType_UNKNOWN {
+			c.DatetimeClause.ActionType = actionType
+			commands = append(commands, c)
+		}
```
kentakozuka, line 985: Maybe this?
```suggestion
		if c.DatetimeClause.ActionType == autoopsproto.ActionType_UNKNOWN {
			c.DatetimeClause.ActionType = actionType
		}
		commands = append(commands, c)
```

kakcy: Thank you！
I Fixed https://github.com/bucketeer-io/bucketeer/pull/1095/commits/10702ed311855a7d0c77e32de0dd0bbada19045c .

kentakozuka (dismissed):
LGTM ❤️

kentakozuka (approved):

kakcy:
Thank you!

PR #1097, opened 2024-07-04
chore: remove dex

cre8ivejp:
Part of https://github.com/bucketeer-io/bucketeer/issues/405

### Things done

- Removed dex
- Refactor Auth API to reuse the refresh and token generation when using multiple Oauth services
- Removed unused fields in the Authentication

----

This pull request includes several changes to the codebase, primarily focusing on the removal of certain OAuth configurations and dependencies from the backend and batch services, as well as the removal of the `dex` service. These changes are part of a larger effort to simplify the codebase and remove unnecessary dependencies.

Removal of OAuth configurations:

* [`manifests/bucketeer/charts/backend/templates/deployment.yaml`](diffhunk://#diff-f3b6d027d73efe3a994465eb68c12ae044c7065a5c8fd20d38674fb416c0aef8L149-L156): Removed several OAuth-related environment variables from the backend service configuration. [[1]](diffhunk://#diff-f3b6d027d73efe3a994465eb68c12ae044c7065a5c8fd20d38674fb416c0aef8L149-L156) [[2]](diffhunk://#diff-f3b6d027d73efe3a994465eb68c12ae044c7065a5c8fd20d38674fb416c0aef8L165-L180)
* [`manifests/bucketeer/charts/backend/values.dev.yaml`](diffhunk://#diff-4c42fb2cdeb0ff66a5f05236d2ad2b6464824fd5a122618ededbf6ff460c2bf3L48-R52): Removed several OAuth-related configurations from the values file for the backend service.
* [`manifests/bucketeer/charts/batch/templates/deployment.yaml`](diffhunk://#diff-299980ca0f3ebbce9876b6dabecf46bed633a07f1206ad49a93eb68379924c7cL134-R136): Replaced the `BUCKETEER_BATCH_OAUTH_CLIENT_ID` environment variable with `BUCKETEER_BATCH_OAUTH_AUDIENCE` in the batch service configuration.
* [`manifests/bucketeer/charts/batch/values.dev.yaml`](diffhunk://#diff-e22de46129c31ba83d35848cbda70471be471b32ce7f95022d4d5fb971bbf086L60-R61): Replaced the `clientId` configuration with `audience` in the values file for the batch service.

Removal of dependencies:

[...]

PR #1104, opened 2024-07-08
chore: change to verify issuer and audience in the access token

cre8ivejp:
Part of https://github.com/bucketeer-io/bucketeer/pull/1097

Because the issuer was different when using Dex, we had to stop verifying the issuer and audience until we updated the new issuer and audience using the new authenticator.


kentakozuka (approved):
🚀

PR #1113, opened 2024-07-09
fix: cannot update archived features

kentakozuka:
This PR enables UpdateFeature API to update archived features.

### DIFF
--- a/pkg/feature/api/feature.go
+++ b/pkg/feature/api/feature.go
@@ -761,9 +761,9 @@ func (s *FeatureService) UpdateFeature(
 		if err != nil {
 			return err
 		}
-		// The feature is removed from the list to be evaluated if it is
-		// archived or deleted. If it is disabled features are still evaluated.
-		evaluateds := []*featureproto.Feature{}
+		// To check if the flag to be updated is a dependency of other flags, we must validate it before updating.
+		// Exclude all the archived and deleted flags from the list.
+		tgts := []*featureproto.Feature{}
 		for _, f := range features {
 			if f.Id == updated.Id {
 				f = updated.Feature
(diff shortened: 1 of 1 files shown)


pkg/feature/api/feature.go
```diff
@@ -762,17 +761,16 @@ func (s *FeatureService) UpdateFeature(
...
-		// The feature is removed from the list to be evaluated, if it is
+		// The feature is removed from the list to be evaluated if it is
 		// archived or deleted. If it is disabled features are still evaluated.
-		if updated.Archived || updated.Deleted {
-			for i, f := range features {
-				if f.Id == updated.Id {
-					features = append(features[:i], features[i+1:]...)
-					break
-				}
+		evaluateds := []*featureproto.Feature{}
```
cre8ivejp, line 766:
```suggestion
		filteredFlags := []*featureproto.Feature{}
```

pkg/feature/api/feature.go
```diff
@@ -762,17 +761,16 @@ func (s *FeatureService) UpdateFeature(
 		if err != nil {
 			return err
 		}
-		// The feature is removed from the list to be evaluated, if it is
+		// The feature is removed from the list to be evaluated if it is
 		// archived or deleted. If it is disabled features are still evaluated.
```
cre8ivejp, line 765: The comment is confusing to me.
It describes that it is preparing a list to be evaluated, but there is no evaluation in the updating process.

```suggestion
                 // To check if the flag to be updated is a dependency of other flags, we must validate it before updating.
		// Exclude all the archived and deleted flags from the list.
```

cre8ivejp (approved):
Thank you!

Issue #1070, opened 2024-06-26
chore: publish feature updated event when progressive rollout is triggered


kentakozuka:
### Describe the feature you'd like

Publish feature updated event when progressive rollout is triggered

PR #1114, opened 2024-07-09
chore: call update feature api at progressive rollout execution

kentakozuka:
Fixes https://github.com/bucketeer-io/bucketeer/issues/1070

### DIFF
--- a/pkg/autoops/api/progressive_rollout.go
+++ b/pkg/autoops/api/progressive_rollout.go
@@ -21,6 +21,7 @@ import (
 
 	"go.uber.org/zap"
 	"google.golang.org/genproto/googleapis/rpc/errdetails"
+	"google.golang.org/protobuf/types/known/wrapperspb"
 
 	"github.com/bucketeer-io/bucketeer/pkg/autoops/command"
 	"github.com/bucketeer-io/bucketeer/pkg/autoops/domain"
@@ -458,10 +459,12 @@ func (s *AutoOpsService) ExecuteProgressiveRollout(
 			return nil
 		}
 		// Enable the flag if it is disabled and it is the first rollout execution
+		var enabled *wrapperspb.BoolValue
 		if !feature.Enabled && progressiveRollout.IsWaiting() {
 			if err := feature.Enable(); err != nil {
 				return err
 			}
+			enabled = &wrapperspb.BoolValue{Value: true}
 		}
 		handler, err := command.NewProgressiveRolloutCommandHandler(
 			editor,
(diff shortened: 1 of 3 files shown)


pkg/autoops/api/progressive_rollout.go
```diff
@@ -475,7 +478,13 @@ func (s *AutoOpsService) ExecuteProgressiveRollout(
 		); err != nil {
 			return err
 		}
-		if err := ftStorage.UpdateFeature(ctx, feature, req.EnvironmentNamespace); err != nil {
+		if _, err := s.featureClient.UpdateFeature(ctx, &featureproto.UpdateFeatureRequest{
```
cre8ivejp, line 481: Calling an API inside of a transaction MySQL is not a good idea.
If we need to increment the version, we should implement it here.
This could easily get into the wrong state where the API succeeds, but MySQL doesn't.

kentakozuka: I agree with that. It is not a good idea to call API inside the transaction. But I don't think updating the feature directly and create a domain event here is a good way. So I will change this to calling UpdateFeature outside transaction (the way in which the autoops does).

cre8ivejp: If the auto ops service changes the feature entity, I don't see why we can't create the event to publish it here.
The issue remains even after moving the API calls outside of the transaction. The Progressive Rollout will be updated as triggered, and if the API call fails, the Flag won't change its state, causing bugs and making it harder to debug.

To avoid other issues, it will be safer to update the flag in the same transaction and create the domain event for it here.

kentakozuka:
@cre8ivejp Fixed.
I tried to move lines of code that are supposed to be commonly used, but there are few lines that I can move.
So, I just changed it to publish a domain event inside the API method.

pkg/autoops/api/progressive_rollout.go
```diff
@@ -528,6 +558,23 @@ func (s *AutoOpsService) ExecuteProgressiveRollout(
...
+			)...,
+		)
+		dt, err := statusInternal.WithDetails(&errdetails.LocalizedMessage{
+			Locale:  localizer.GetLocale(),
+			Message: localizer.MustLocalize(locale.InternalServerError),
+		})
+		if err != nil {
+			return nil, statusInternal.Err()
+		}
+		return nil, dt.Err()
```
cre8ivejp, line 576: It's just a note, but because this is a background operation, we don't need to localize the error.

cre8ivejp (approved):
Nice! Thank you!

PR #1117, opened 2024-07-11
chore: make the `ttlSecondsAfterFinished` optional in the migration job

cre8ivejp:
The current setting is 24 hours. After the Job is deleted, PipeCD keeps showing an out-of-sync warning because the job is no longer there. So, I'm changing the `ttlSecondsAfterFinished` to be optional so we can keep the job after completion and avoid warnings.


Ubisoft-potato (approved):
LGTM!

PR #1118, opened 2024-07-11
chore: dev container performance

Ubisoft-potato:
This pull request will increase the resource limits upon starting Minikube. 

Additionally, it will verify if the user has initiated Minikube more than twice; if is, the start process will be skipped.


cre8ivejp (approved):
Thanks!

PR #1120, opened 2024-07-12
fix: auditlogs have empty entity data

kentakozuka:


pkg/auditlog/domain/auditlog.go
```diff
@@ -27,14 +27,16 @@ type AuditLog struct {
 func NewAuditLog(event *domainevent.Event, envirronmentNamespace string) *AuditLog {
```
cre8ivejp, line 27:
```suggestion
func NewAuditLog(event *domainevent.Event, environmentNamespace string) *AuditLog {
```

pkg/auditlog/domain/auditlog.go
cre8ivejp, line 1: Please add a unit test.

Reference.
https://github.com/bucketeer-io/bucketeer/blob/7b5395d39c53932a2369b7519312046cde381bea/pkg/environment/domain/environment_test.go#L24-L42

cre8ivejp (approved):
Thank you!

PR #1126, opened 2024-07-17
chore: improve security in the sign-in api

cre8ivejp:
If we disabled the demo sign-in but the demo account is still enabled, someone could get the access token by calling the SignIn API directly.
So, I changed it to check the config file to ensure we won't access it if it is disabled.


Ubisoft-potato (approved):
Looks great! 👍

PR #1127, opened 2024-07-17
fix: demo environment not being created

cre8ivejp:
I found an error when creating the demo environment in the dev environment.
Because we already have a project with the `demo` ID, it fails with an `Already Exists` error, not proceeding with the environment and account creation.

I fixed it only to stop the process if it gets an error different than not found or already exists errors.


Ubisoft-potato (approved):
Great improvement!

PR #1141, opened 2024-07-19
chore: return unauthenticated error if account disabled

kentakozuka:


cre8ivejp:
@kentakozuka, I think we should also check [here](https://github.com/bucketeer-io/bucketeer/pull/1141/files#diff-0461d777e3a3b676a8900a9b57e31c5f5126a69848cbef75d72d4e3db1059ed3R114-R123) just in case.

kentakozuka:
@cre8ivejp Thanks! Fixed 👍

cre8ivejp (approved):
Thank you!

PR #1146, opened 2024-07-22
chore: remove not found log from check account status

cre8ivejp:
System admin accounts have access to all organizations but are registered only in the system admin organization.
So, to avoid false errors, we ignore them if the account wasn't found in non-system admin organizations.


Ubisoft-potato (approved):
👍

PR #1153, opened 2024-07-23
chore: change envoy load balancer to least_request

cre8ivejp:
Since `LEAST_REQUEST` outperforms `ROUND_ROBIN` in many cases. I'm changing the default value to see how it performs on Bucketeer.

Reference:
`
Updated Istio’s default load balancing algorithm from ROUND_ROBIN to LEAST_REQUEST. The ROUND_ROBIN algorithm can lead to overburdened endpoints, especially when weights are used. The LEAST_REQUEST algorithm distributes the load more evenly across and is far less likely to overburden endpoints. A number of experiments (by both the Istio and Envoy teams) have shown that LEAST_REQUEST outperforms ROUND_ROBIN in virtually all cases, with little/no downsides. It’s generally considered a drop-in replacement for ROUND_ROBIN.
`
https://istio.io/latest/news/releases/1.14.x/announcing-1.14/change-notes/


Ubisoft-potato (approved):
`LEAST_REQUEST` looks great! Let's try it.

PR #1158, opened 2024-07-24
chore: delete docker building cache when open dev container

cre8ivejp:
When building many times a day, it will increase the storage usage and fail with no space left.
So, I added the `docker system prune -f` command to delete the docker building cache to avoid that error.


Ubisoft-potato (approved):
🚀!

PR #1162, opened 2024-07-26
fix(ui): auto operation action type not being set

cre8ivejp:
The action type wasn't set when creating an event rate operation. Also, it wasn't set when creating a schedule operation on the feature flags list page.


kakcy (approved):
Thank you 👍

bimalgrg519 (approved):

PR #1164, opened 2024-07-26
fix: ops event rate not being processed

cre8ivejp:
### Things done

- Fixed count query
- Updated e2e tests to use the correct `ops_type and` and `action_type` when testing event rate operations.

### DIFF
--- a/pkg/batch/storage/v2/auto_ops_rule.go
+++ b/pkg/batch/storage/v2/auto_ops_rule.go
@@ -28,7 +28,7 @@ var (
 )
 
 type AutoOpsRuleStorage interface {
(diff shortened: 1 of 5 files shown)


pkg/batch/storage/v2/auto_ops_rule.go
```diff
@@ -28,7 +28,7 @@ var (
 )
 
 type AutoOpsRuleStorage interface {
-	CountNotTriggeredAutoOpsRules(ctx context.Context) (int, error)
+	CountOpsEventRate(ctx context.Context) (int, error)
```
cre8ivejp, line 31: I renamed the interface because the meaning was incorrect.

kakcy (approved):
Thank you 👍

PR #1165, opened 2024-07-26
chore: ensure that all kubectl commands for the dev container run in the minikube only

cre8ivejp:
To avoid the kubectl commands for the dev container being executed on the host by accident, I changed to always executing it in the `minikube` context.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -324,6 +324,7 @@ modify-hosts:
 
 # enable vault transit secret engine
 enable-vault-transit:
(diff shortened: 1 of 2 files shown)


ui/web-v2/dist/DONT-EDIT-FILES-IN-THIS-DIRECTORY
Ubisoft-potato, line 1: We can delete this file, I think we should add a check here

PR #1166, opened 2024-07-26
build(ui): add post build script to recreate the file after building web project

cre8ivejp:
When we build the web project, it deletes the `DONT-EDIT-FILES-IN-THIS-DIRECTORY` in the dist directory.
Because the go embed building needs that file, we must ensure it is not accidentally deleted and pushed to the main branch.


Ubisoft-potato (approved):
👍

Issue #721, opened 2024-01-23
feat: save the search filter options


cre8ivejp:
Every time the user accesses the console, he needs to configure the filters again. It would be great if there were a feature to save the last filter options and reuse them when accessing the console.

- [ ] Backend
- [ ] Frontend

PR #1167, opened 2024-07-29
docs: add rfc for Save Search Filter

kakcy:
This PR adds RPC documentation for implementing the Save Search Filter. : #721

### DIFF
--- a/docs/rfcs/save-search-filter.md
+++ b/docs/rfcs/save-search-filter.md
@@ -2,23 +2,33 @@
 
 "Save Filters" saves the filters set for each feature list in the console.
 
-This feature allows you to give default settings to saved filters.
+This feature allows to give default settings to saved filters.
 
 This feature should be designed to store filter information for the various information displayed in the console.
 
-## package
-packageName: savefilter
-
-## Function
 Filter and search information is based on the string in the query part of the URL.
 
 ex) https://dev.bucketeer.jp/default/features?page=1&q=test&sort=-createdAt&tagIds=iOS
 
 Filter and Search Info: ```q=test&sort=-createdAt&tagIds=iOS```
 
+**Concerns**
+- Since the string we save is a query parameter, the saved query may not work if the filter is changed.
+   - ex: Save Query String [&sort=-createdAt&tagIds=iOS]
+     If there is no longer a filter for Tag, saved queries will not work.
+
+I have considered the following two methods, but considering the advantages and disadvantages, we will propose ***Part2***.
+
+## Part1: Create a new search_filter table and manage one record for each filter.
+
+### Package
+packageName: savefilter
+
+### API
+
 In common with all functions, instead of receiving an email address from the user to identify the user, we use the [AccessToken.Email](https://github.com/bucketeer-io/bucketeer/blob/main/pkg/token/token.go#L26) obtained with [GetAccessToken](https://github.com/bucketeer-io/bucketeer/blob/41da1916100bf29cc925010b629f74c17e014d5f/pkg/rpc/auth.go#L70).
 
-### Create
+#### Create
 Receives and saves requests for filter titles, query strings, and default settings flags.
 * Request
   * Name: Filter Name String
@@ -33,7 +43,7 @@ Receives and saves requests for filter titles, query strings, and default settin
   * TargetType: Reject Unknown Type
   * EnvironmentId: check EnvironmentRole
 
-### Update
+#### Update
 Receives and updates requests for filter titles, query strings, and default settings flags.
 * Request
   * ID: Filter ID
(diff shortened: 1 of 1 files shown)


docs/rfcs/save-search-filter.md
```diff
@@ -0,0 +1,81 @@
...
+  * TargetType: Reject Unknown Type
+
+## Implementation
+
+### Storage
+
+* Create `search_filter` table definition.
+
+```sql
+      CREATE TABLE `search_filter` (
```
cre8ivejp, line 66: How do you plan to connect the filter to the user logged in to the console?

docs/rfcs/save-search-filter.md
```diff
@@ -0,0 +1,81 @@
...
+ex) https://dev.bucketeer.jp/default/features?page=1&q=test&sort=-createdAt&tagIds=iOS
+
+Filter and Search Info: ```q=test&sort=-createdAt&tagIds=iOS```
+
+### Create
+Receives and saves requests for filter titles, query strings, and default settings flags.
+* Request
+  * Name: Filter Name String
+  * Query: Query Parameters String
+  * DefaultFlag: Filter flags to set by default
```
cre8ivejp, line 24: Can you give an example of using this field?

docs/rfcs/save-search-filter.md
```diff
@@ -0,0 +1,81 @@
...
+* Create `search_filter` table definition.
+
+```sql
+      CREATE TABLE `search_filter` (
+        `id` varchar(255) CHARACTER SET utf8mb4 COLLATE utf8mb4_bin NOT NULL,
+        `name` sting NOT NULL,
+        `query` sting NOT NULL,
+        `target_type` int NOT NULL,
+        `default` tinyint(1) NOT NULL DEFAULT '0',
+        `environment_namespace` sting NOT NULL,
```
cre8ivejp, line 72: For new tables, please use `environment_id` instead.

docs/rfcs/save-search-filter.md
kakcy: I have a few questions regarding this comment.

Since this comment only has `Create`, does it mean that a `SELECT` type query is also required to obtain the search list?

Or is the intent of how it's supposed to be managed since the Create request doesn't include any information about the user?
I was thinking of saving per environment, not per user. (I just realized it's better to run it per project instead of per environment.)
I thought it would be useful to be able to share saved search results with your team.What do you think?
Or is it better to run it for each user? (It might not be that difficult if you share the URL with each user.)

docs/rfcs/save-search-filter.md
kakcy: I know there is a "Set as preferred" in the menu.
This field is intended to be used when you want to prioritize search conditions for saved search information.

docs/rfcs/save-search-filter.md
cre8ivejp: Since each user has his own preference, the feature is more useful per user not per environment or project.

docs/rfcs/save-search-filter.md
cre8ivejp: I see.
It's better to use `DefaultFilter` as the name.

cre8ivejp:
@kakcy, have you considered creating one record per user in the DB? The current schema creates 1 record per filter, so if the user has 10 filters, it will have 10 records in the DB.

docs/rfcs/save-search-filter.md
cre8ivejp: Since the filter is per user, have you considered saving the filters in the account table?

kakcy:
@cre8ivejp 

> have you considered creating one record per user in the DB? The current schema creates 1 record per filter, so if the user has 10 filters, it will have 10 records in the DB.

As you pointed out, it is designed so that one record exists in one filter.
Is this a problem with this design? (Will the number of records be huge?)

Or is it better to add a column like `filters` to the `account_v2` table without creating a new table?
In that case, it is assumed that it will be stored in a Json array like `environment_roles` column.

kakcy:
> Since the filter is per user, have you considered saving the filters in the account table?

Added Part2 proposal to add a filter information column to the `account` table.
Please review when you have time.

kentakozuka:
@kakcy Why don't you save queries in the local storage? It's easier to implement.

cre8ivejp:
> @kakcy Why don't you save queries in the local storage? It's easier to implement.

@kentakozuka, as discussed before, if the user accesses from another PC or the cache is deleted, we will lose all his filters.
Also, the other services that save on the server have a better user experience.

kakcy:
@kentakozuka 
Proposed Thank you 👍 
Similar to @cre8ivejp's opinion, I think it would be better to save it to the DB in order to provide stable functionality.

kentakozuka:
@kakcy @cre8ivejp 
Understood. But I still think this feature would be too much. How often do we change our browsers to access Bucketeer? How often is the local storage deleted? Would it be worth to bring the complexity?

PR #1169, opened 2024-07-30
fix: pubsub message not being acknowledged when the reason is nil

cre8ivejp:
Fix https://github.com/bucketeer-io/bucketeer/issues/1170


Ubisoft-potato (approved):
👍👍

PR #1171, opened 2024-07-31
fix: validated the same date for schedule Auto Ops

kakcy:
Fixed to reject requests that have the same date or already have the same date when scheduling AutoOps.

### DIFF
--- a/pkg/autoops/api/api.go
+++ b/pkg/autoops/api/api.go
@@ -19,6 +19,12 @@ import (
 	"strconv"
 	"time"
 
+	"go.uber.org/zap"
+	"google.golang.org/genproto/googleapis/rpc/errdetails"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
 	accountclient "github.com/bucketeer-io/bucketeer/pkg/account/client"
 	authclient "github.com/bucketeer-io/bucketeer/pkg/auth/client"
 	"github.com/bucketeer-io/bucketeer/pkg/autoops/command"
@@ -38,11 +44,6 @@ import (
 	autoopsproto "github.com/bucketeer-io/bucketeer/proto/autoops"
 	eventproto "github.com/bucketeer-io/bucketeer/proto/event/domain"
 	experimentproto "github.com/bucketeer-io/bucketeer/proto/experiment"
-	"go.uber.org/zap"
-	"google.golang.org/genproto/googleapis/rpc/errdetails"
-	"google.golang.org/grpc"
-	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/status"
 )
 
 type options struct {
--- a/pkg/autoops/api/api.go
+++ b/pkg/autoops/api/api.go
@@ -412,22 +412,9 @@ func (s *AutoOpsService) validateDatetimeClauses(
 	clauses []*autoopsproto.DatetimeClause,
 	localizer locale.Localizer,
 ) error {
-	var checkClauses []*autoopsproto.DatetimeClause
+	checkTimes := make(map[int64]bool)
 	for _, c := range clauses {
-		if err := s.validateDatetimeClause(c, checkClauses, localizer); err != nil {
-			return err
-		}
-		checkClauses = append(checkClauses, c)
-	}
-	return nil
-}
-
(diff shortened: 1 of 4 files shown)


pkg/autoops/api/api.go
```diff
@@ -412,15 +412,32 @@ func (s *AutoOpsService) validateDatetimeClauses(
 	clauses []*autoopsproto.DatetimeClause,
 	localizer locale.Localizer,
 ) error {
+	var checkClauses []*autoopsproto.DatetimeClause
```
kentakozuka, line 415: We could use map for checkClauses, which is faster and easier to understand code.

pkg/autoops/domain/auto_ops_rule.go
```diff
@@ -337,6 +337,21 @@ func (a *AutoOpsRule) ExtractDatetimeClauses() (map[string]*proto.DatetimeClause
...
+		return nil, err
+	}
+	for id, c := range datetimeClauses {
+		if id == clauseId {
+			datetimeClause = c
+			break
+		}
+	}
+	return datetimeClause, nil
+}
```
kentakozuka, line 353: Is this function necessary?
You can simply callctDatetimeClauses and then get one item by clause ID(as key of map).

pkg/autoops/api/api.go
```diff
@@ -727,6 +744,52 @@ func (s *AutoOpsService) UpdateAutoOpsRule(
 				}
 				return dt.Err()
 			}
+
+			// Extract each schedule time (however, omit the time scheduled for deletion)
+			var scheduleTimes []int64
```
kentakozuka, line 749: Please use map here. It would be faster and easier.

kentakozuka (commented):
Thanks! Left comments. PTAL.

kakcy (commented):
Removed `ExtractDatetimeClause()` and reviewed the processing of the pointed part.

pkg/autoops/api/api.go
```diff
@@ -727,6 +741,42 @@ func (s *AutoOpsService) UpdateAutoOpsRule(
...
+
+			// Delete a deletion schedule from the currently held schedules
+			extractDateTimeClauses, _ := autoOpsRule.ExtractDatetimeClauses()
+			for _, deleteClause := range req.DeleteClauseCommands {
+				delete(extractDateTimeClauses, deleteClause.Id)
+			}
+
+			// Check if there is a schedule with the same date and time.
+			for _, c := range req.AddDatetimeClauseCommands {
+				for _, extractDateTimeClause := range extractDateTimeClauses {
```
kentakozuka, line 753: Iterating `extractDateTimeClauses` in each validation is inefficient. How about creating a map with a time key at first and checking if the same time exists in each loop?

pkg/autoops/api/api.go
```diff
@@ -930,10 +982,12 @@ func (s *AutoOpsService) validateUpdateAutoOpsRuleRequest(
 			}
 			return dt.Err()
 		}
-		if err := s.validateDatetimeClause(c.DatetimeClause, localizer); err != nil {
-			return err
-		}
+		checkDatetimeClauses = append(checkDatetimeClauses, c.DatetimeClause)
+	}
+	if err := s.validateDatetimeClauses(checkDatetimeClauses, localizer); err != nil {
+		return err
```
kentakozuka, line 988: These lines seem to check the same as validation inside the transaction. Is it necessary?

kentakozuka (commented):
Thanks! Left comments. PTAL 🙏

pkg/autoops/api/api.go
kakcy: This validation process checks if the same date exists in the `AddDatetimeClauseCommands`.

Validation within a transaction is a check to see if there is a date that is the same as the date that has already been set, so validation here is necessary.

pkg/autoops/api/api.go
kakcy: I fixed https://github.com/bucketeer-io/bucketeer/pull/1171/commits/5f05b90543c7115fc2a7c6700de6e43e5508b044.

pkg/autoops/api/api.go
kentakozuka: I see. Thanks!

kentakozuka (approved):
LGTM ❤️

kakcy:
Thank you!

PR #1179, opened 2024-08-01
chore: delete triggeredAt for AutoOps

kakcy:


proto/event/domain/event.proto
```diff
@@ -789,7 +789,7 @@ message AutoOpsRuleCreatedEvent {
   string feature_id = 1;
   bucketeer.autoops.OpsType ops_type = 2;
   repeated bucketeer.autoops.Clause clauses = 3;
-  int64 triggered_at = 4 [deprecated = true];
+  reserved 4;
```
kentakozuka, line 792:
```suggestion
  reserved 4; // int64 triggered_at = 4 [deprecated = true];
```

proto/autoops/service.proto
```diff
@@ -85,8 +85,7 @@ message UpdateAutoOpsRuleResponse {}
 message ExecuteAutoOpsRequest {
   string environment_namespace = 1;
   string id = 2;
-  ChangeAutoOpsRuleTriggeredAtCommand
-      change_auto_ops_rule_triggered_at_command = 3 [deprecated = true];
+  reserved 3;
```
kentakozuka, line 88:
```suggestion
  reserved 3; // ChangeAutoOpsRuleTriggeredAtCommand
      change_auto_ops_rule_triggered_at_command = 3 [deprecated = true];
```

proto/autoops/auto_ops_rule.proto
```diff
@@ -24,7 +24,7 @@ message AutoOpsRule {
   string feature_id = 2;
   OpsType ops_type = 3;
   repeated Clause clauses = 4;
-  int64 triggered_at = 6 [deprecated = true];
+  reserved 6;
```
kentakozuka, line 27:
```suggestion
  reserved 6; // int64 triggered_at = 6 [deprecated = true];
```

pkg/autoops/api/api.go
```diff
@@ -706,7 +706,7 @@ func (s *AutoOpsService) UpdateAutoOpsRule(
 			return err
 		}
 
-		if autoOpsRule.AlreadyTriggered() || autoOpsRule.IsFinished() || autoOpsRule.IsStopped() {
+		if autoOpsRule.IsFinished() || autoOpsRule.IsStopped() {
```
kentakozuka, line 709: How about adding a new method `IsCompleted()` to the domain?

kentakozuka (commented):
Thanks! left comments. ptal.

pkg/autoops/api/api.go
cre8ivejp: The `isCompleted` will be confusing IMHO.
Because the complete and finish have the same meaning in this case.
It will be easier to understand and leave it as it is. WDYT?

pkg/autoops/storage/v2/auto_ops_rule.go
```diff
@@ -179,7 +182,7 @@ func (s *autoOpsRuleStorage) ListAutoOpsRules(
 			&autoOpsRule.FeatureId,
 			&opsType,
 			&mysql.JSONObject{Val: &autoOpsRule.Clauses},
-			&autoOpsRule.TriggeredAt,
+			&tmpTriggeredAt, // ToDo: TriggeredAt is deprecated. Remove this line after migration.
```
cre8ivejp, line 185: By removing this [line](https://github.com/bucketeer-io/bucketeer/blob/main/pkg/autoops/storage/v2/sql/auto_ops_rule/select_auto_ops_rules.sql#L6), we don't need to wait for the migration.

pkg/autoops/storage/v2/sql/auto_ops_rule/update_auto_ops_rule.sql
```diff
@@ -4,7 +4,7 @@ SET
     feature_id = ?,
     ops_type = ?,
     clauses = ?,
-    triggered_at = ?,
+    triggered_at = 0,
```
cre8ivejp, line 7: There is no need to set this to zero.
You can remove this line from the query because the field is not required when updating.

pkg/autoops/api/api.go
kakcy: @cre8ivejp @kentakozuka 

It may be confusing how to use it.
Add deleted to the condition,
How about changing it to `IsInactive()`?

cre8ivejp: @kakcy, the issue is not the function name. By creating `IsInactive` or `isCompleted`, we are creating a new state that doesn't exist in the auto_ops_rule domain layer, which is confusing because  `completed` and `finished` words have the same meaning in this case.
So, my suggestion is to leave it as it is.

kakcy: @cre8ivejp @kentakozuka 
Since there were multiple locations that made the same judgment, we decided that it would be better to add a new one to improve the readability of the code.

However, as you pointed out, `IsCompleted()` does not exist in AutoOpsStatus, and having a function that determines a similar status causes confusion when used, so I will discontinue it.

kakcy (commented):
I have corrected the points pointed out.
Please check it when you have time.

cre8ivejp (approved):
Thank you!

PR #1180, opened 2024-08-02
ci: fix dependabot cannot find package.json for ui/web-v2

kentakozuka:


cre8ivejp (approved):
Thanks!

nit: `ci: fix dependabot cannot find package.json for ui/web-v2`

PR #1185, opened 2024-08-05
chore: ignore context canceled errors when publishing to pubsub

cre8ivejp:
The log cost is increasing significantly due to context canceled errors.
Since the request can be canceled by a timeout or the client canceling it, I'm changing it to ignore it.
We can still monitor the context canceled request on the Grafana.


kentakozuka (approved):
Thanks!

PR #1186, opened 2024-08-06
fix: subscriber metric name

cre8ivejp:
Because the Subsystem is a subscriber, the metric name should not include the `subscriber` prefix.


Ubisoft-potato (approved):
👍

PR #1191, opened 2024-08-07
fix: fixed to also check ActionType when validating the same date

kakcy:
When updating an AutoOps schedule, updating to a different ActionType while keeping the same date caused an error.


kentakozuka (approved):
LGTM 👍

kakcy:
Thank you!

PR #1202, opened 2024-08-19
fix: unimplemented api errror in the subscriber svc

cre8ivejp:
Fix https://github.com/bucketeer-io/bucketeer/issues/1203


Ubisoft-potato (approved):
Thanks! 👍

PR #1205, opened 2024-08-20
feat: add support for storage as SearchFilter

kakcy:
- add `UpdateSearchFilters` (a function that updates only SearchFilter)
- change so that `SearchFilters` is also got when getting an account


cre8ivejp (approved):
Thank you!

PR #1207, opened 2024-08-22
fix: missing rule path when deploying

kakcy:


kakcy:
I deleted a necessary setting value carelessly, so I put it back.
I have confirmed that the deployment is successful in the dev container.

cre8ivejp (approved):
Thanks.

nit: `fix: missing rule path when deploying`

PR #1209, opened 2024-08-23
chore: add environment_id column for tables

Ubisoft-potato:
We will repalce `environment_namespace` column to `environment_id` column for all table, this PR will add `environment_id` column first.

Here is the migration process:

1. Add new column
   Add the environment_id column to all relevant tables, allowing NULL initially
   ```sql
   ALTER TABLE `table_name` ADD COLUMN `environment_id` varchar(255) CHARACTER SET utf8mb4 COLLATE utf8mb4_bin NULL;
   ```

2. Populate new column
   Populate the new environment_id column with data from environment_namespace:
   ```sql
   UPDATE `table_name` SET `environment_id` = `environment_namespace`;
   ```
3. Update backend service (Phase 1)
Update service to write to both `environment_namespace` and `environment_id` columns. Read primarily from `environment_id`, falling back to `environment_namespace` if `environment_id` is NULL. This ensures compatibility with both old and new data.

4. Backfill any missing data
   If any data was written during the deployment of the updated service:
   ```sql
   UPDATE `table_name` SET `environment_id` = `environment_namespace` WHERE `environment_id` IS NULL;
   ``` 
5. Make new column NOT NULL
   After verifying all data is populated:
   ```sql
   ALTER TABLE `table_name` MODIFY COLUMN `environment_id` varchar(255) CHARACTER SET utf8mb4 COLLATE utf8mb4_bin NOT NULL;
   ```
6. Update backend service (Phase 2)
Update service to only write to and read from environment_id. Remove all references to environment_namespace.


cre8ivejp (approved):
Thank you!

PR #1211, opened 2024-08-23
chore: add documentation for web account amd environment services

cre8ivejp:
I'm adding account and environment services documentation so the web developer can work on the new UI APIs.


Ubisoft-potato (approved):
LGTM👍!

PR #1213, opened 2024-08-26
fix: sql argument when getting account

cre8ivejp:
Fix the error `sql: expected 19 destination arguments in Scan, not 20"` when getting the user account.


kakcy (approved):
Thank you 👍

PR #1229, opened 2024-09-05
feat(ui/dashboard): nav change project environment

steveninhle:
- Switch environment belongs to the projects list 
- Update the new navigation list 
- Add setting animation

image


cre8ivejp (approved):
Thank you!

PR #1234, opened 2024-09-11
fix: column 'id' in where clause is ambiguous when searching flag

kakcy:
If the experience was filtered by criteria, an Unknown Error occurred when searching by ID or name.


cre8ivejp (approved):
Thank you!

nit: `fix: column 'id' in where clause is ambiguous when searching flag`

PR #1235, opened 2024-09-12
fix: system admin search filter not being updated

cre8ivejp:
Fix https://github.com/bucketeer-io/bucketeer/issues/1236

When a system admin account logs in to a non-system admin organization, its account doesn't exist in that organization, so when the server tries to create/update/delete, it fails because it doesn't find the account.

So, I fixed it to check if it is a system admin. If so, it will replace the organization ID with the system admin organization ID to save it correctly.
When getting the console account via **[GetMe](https://github.com/bucketeer-io/bucketeer/blob/main/pkg/account/api/admin_account.go#L140)** API, it already replaces the organization ID, so no need to change anything.

### DIFF
--- a/pkg/account/api/search_filter_test.go
+++ b/pkg/account/api/search_filter_test.go
@@ -30,6 +30,7 @@ import (
 	v2as "github.com/bucketeer-io/bucketeer/pkg/account/storage/v2"
 	accstoragemock "github.com/bucketeer-io/bucketeer/pkg/account/storage/v2/mock"
 	"github.com/bucketeer-io/bucketeer/pkg/locale"
+	"github.com/bucketeer-io/bucketeer/pkg/pubsub/publisher/mock"
 	accountproto "github.com/bucketeer-io/bucketeer/proto/account"
 )
 
@@ -75,7 +76,7 @@ func TestCreateSearchFilter(t *testing.T) {
 							},
 						},
 					},
-				}, nil).AnyTimes()
+				}, nil)
 			},
 			req: &accountproto.CreateSearchFilterRequest{
 				OrganizationId: "org0",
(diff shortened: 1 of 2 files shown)


pkg/account/api/search_filter_test.go
```diff
@@ -376,6 +389,64 @@ func TestCreateSearchFilter(t *testing.T) {
...
+
+				s.accountStorage.(*accstoragemock.MockAccountStorage).EXPECT().UpdateAccountV2(
+					gomock.Any(), gomock.Any(),
+				).Return(nil)
+
+				s.accountStorage.(*accstoragemock.MockAccountStorage).EXPECT().RunInTransaction(
+					gomock.Any(), gomock.Any(),
+				).Do(func(ctx context.Context, fn func() error) {
+					_ = fn()
+				}).Return(nil)
```
cre8ivejp, line 434: I also added tests to check inside of `RunInTransaction`.

pkg/account/api/search_filter_test.go
```diff
@@ -376,6 +389,64 @@ func TestCreateSearchFilter(t *testing.T) {
...
+				s.accountStorage.(*accstoragemock.MockAccountStorage).EXPECT().GetAccountV2ByEnvironmentID(
+					gomock.Any(), "email", gomock.Any(),
+				).Return(acc, nil)
+
+				s.accountStorage.(*accstoragemock.MockAccountStorage).EXPECT().GetSystemAdminAccountV2(
+					gomock.Any(), email,
+				).Return(acc, nil)
+
+				s.accountStorage.(*accstoragemock.MockAccountStorage).EXPECT().GetAccountV2(
+					gomock.Any(), email, orgID,
```
cre8ivejp, line 419: Here I ensure that the organization ID is the system admin organization and not the ID from the request.

Ubisoft-potato (approved):
👍👍

PR #1237, opened 2024-09-13
chore: using configured stan modelId

Ubisoft-potato:
This PR will eliminate the request for httpstan during the batch service startup, as httpstan requires additional time to be ready after the batch service has started. This will impact the health check of the batch service pod.

We already precompiled the stan model during building httpstan Docker image, so we can use the stan model id directly, we don't need to call httpstan compile API during batch service starting.


manifests/bucketeer/charts/batch/values.yaml
```diff
@@ -97,6 +97,7 @@ httpstan:
     pullPolicy: IfNotPresent
   config:
   port: 8080
+  modelId: y3qsnd7m
```
cre8ivejp, line 100:
```suggestion
  httpStan:
    # The model ID is generated when the HTTP Stan model is compiled.
    # We must update the ID every time we compile the model.
    # Since we don't change the model so often, this ID won't change.
    modelId: y3qsnd7m
```

pkg/batch/jobs/calculator/experiment_calculate.go
```diff
@@ -52,6 +52,7 @@ type experimentCalculate struct {
 
 func NewExperimentCalculate(
 	httpStan *stan.Stan,
+	modelID string,
```
cre8ivejp, line 55: Please rename `modelID` -> `stanModelID` here and in the other places.
```suggestion
	stanModelID string,
```

pkg/batch/jobs/calculator/experiment_calculate.go
```diff
@@ -154,8 +156,7 @@ func (e *experimentCalculate) calculateExperimentWithLock(ctx context.Context,
 		)
 		return nil
 	}
-
-	defer func() {
+	if calcErr := e.calculateExperiment(ctx, env, experiment); calcErr != nil {
```
cre8ivejp, line 159:
```suggestion
	if calcErr := e.calculateExperiment(ctx, env, experiment); calcErr != nil {
	    // To prevent calculating the same experiment multiple times in a short time,
	    // we set the TTL for the lock key and only unlock it when an error occurs.
```

cre8ivejp (approved):
Thank you!

Issue #1252, opened 2024-09-20
enhancement: remove mysql dependency from api layer


cre8ivejp:

kakcy:
**StorageList**
- [ ]  account
- [ ]  auditlog
- [ ]  autoops
- [ ]  environment
- [ ]  eventcounter
- [ ]  experiment
- [ ]  feature
- [ ]  notification
- [ ]  push
- [ ]   experimentcalculator
- [ ]   mau
- [ ]   opsevent
- [ ]   subscriber

PR #1266, opened 2024-10-02
test: refactor account storage test

kakcy:
I introduced [go-sqlmock](https://github.com/DATA-DOG/go-sqlmock) to perform tests from the following perspectives for storage layer unit tests.
- Test whether the content of the query is what you intended.
- Test whether the Select Queries returns the acquired information correctly.
- Test whether the intended value is successfully passed to Query.


kakcy:
I will close this for now as it will need to be considered after #1252 is addressed.

PR #1279, opened 2024-10-14
fix: audit log pubg message not being acknowledged

cre8ivejp:
When the subscriber tries to insert an audit log that already exists in the DB, it doesn't acknowledge the PubSub message and keeps trying in an infinite loop until the message expires in PubSub.

I'm changing it to insert a single audit log so we can acknowledge the correct message when it happens and not all the messages in the chunk, which could lead to data loss.

### DIFF
--- a/pkg/auditlog/storage/v2/admin_audit_log.go
+++ b/pkg/auditlog/storage/v2/admin_audit_log.go
@@ -31,6 +31,8 @@ import (
 var (
 	ErrAdminAuditLogAlreadyExists = errors.New("auditlog: admin auditlog already exists")
 
+	//go:embed sql/adminauditlog/insert_admin_audit_logs_v2.sql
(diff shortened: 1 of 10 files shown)


pkg/auditlog/storage/v2/admin_audit_log.go
```diff
@@ -41,6 +43,7 @@ var (
 
 type AdminAuditLogStorage interface {
 	CreateAdminAuditLogs(ctx context.Context, auditLogs []*domain.AuditLog) error
```
Ubisoft-potato, line 45: The `CreateAdminAuditLogs` will not be used after apply this PR, maybe we can remove it?

cre8ivejp: I'm not sure yet. So, I'm keeping it for now.

PR #1289, opened 2024-10-21
feat: support non-command for web create Push API

hvn2k1:
Related to issue: https://github.com/bucketeer-io/bucketeer/issues/904
- Support non-command input for web create Push API
- Return Push object after create Push


pkg/push/api/api.go
```diff
@@ -236,20 +241,149 @@ func (s *PushService) CreatePush(
...
+	return &pushproto.CreatePushResponse{
+		Push: push.Push,
+	}, nil
 }
 
-func (s *PushService) validateCreatePushRequest(req *pushproto.CreatePushRequest, localizer locale.Localizer) error {
-	if req.Command == nil {
-		dt, err := statusNoCommand.WithDetails(&errdetails.LocalizedMessage{
+// CreatePushV2 implement logic without command
+func (s *PushService) CreatePushV2(
```
cre8ivejp, line 250: Is this a new API?
If so, instead of creating a new API, can you implement this using the current `CreatePush` API?

proto/push/service.proto
cre8ivejp, line 1: We need to configure the API method and URL for the rest of the APIs.
Also, we need to write the documentation.

### Reference

URL Configuration
https://github.com/bucketeer-io/bucketeer/blob/main/proto/environment/service.proto#L302-L305

Documentation:
https://github.com/bucketeer-io/bucketeer/blob/main/proto/environment/service.proto#L306-L345

pkg/push/api/api.go
hvn2k1: Oh no, this is still the old API, I should have named it as createPushV2 or createPushNoCommand, this function is called in https://github.com/bucketeer-io/bucketeer/pull/1289/files#diff-927c02bd692b9ee775cb3720ede88aa16612d46cf46213f45274a0ddd792c782R114

hvn2k1: I have rename the function to avoid misunderstanding 😄

pkg/push/api/api.go
```diff
@@ -236,20 +241,149 @@ func (s *PushService) CreatePush(
 		}
 		return nil, dt.Err()
 	}
-	return &pushproto.CreatePushResponse{}, nil
+	return &pushproto.CreatePushResponse{
+		Push: push.Push,
```
cre8ivejp, line 245: For security reasons, we must remove the FCM service account from the response

Reference:
https://github.com/bucketeer-io/bucketeer/blob/15a8681012601abfbee96be1a6b849420cbf525d/pkg/push/api/api.go#L942-L945

pkg/push/api/api.go
```diff
@@ -236,20 +241,149 @@ func (s *PushService) CreatePush(
...
+	)
+	if err != nil {
+		return nil, err
+	}
+	if err = s.publisher.Publish(ctx, event); err != nil {
+		return nil, err
 	}
+
+	return &pushproto.CreatePushResponse{
+		Push: push.Push,
```
cre8ivejp, line 382: Same here.

pkg/push/api/api_test.go
```diff
@@ -215,6 +216,127 @@ func TestCreatePushMySQL(t *testing.T) {
 	}
 }
 
+func TestCreatePushV2MySQL(t *testing.T) {
```
cre8ivejp, line 219:
```suggestion
func TestCreatePushNoCommandMySQL(t *testing.T) {
```

pkg/push/api/api.go
```diff
@@ -236,20 +241,149 @@ func (s *PushService) CreatePush(
...
+		string(req.FcmServiceAccount),
+		req.Tags,
+	)
+	if err != nil {
+		s.logger.Error(
+			"Failed to create a new push",
+			log.FieldsFromImcomingContext(ctx).AddFields(
+				zap.Error(err),
+				zap.String("environmentNamespace", req.EnvironmentNamespace),
+				zap.Strings("tags", req.Tags),
```
cre8ivejp, line 270: Let's add the `name` to the logs for better debugging.
Same for the other logs.

pkg/push/api/api.go
```diff
@@ -236,20 +241,149 @@ func (s *PushService) CreatePush(
...
+		})
+		if err != nil {
+			return nil, statusInternal.Err()
+		}
+		return nil, dt.Err()
+	}
+
+	var event *eventproto.Event
+	pushStorage := v2ps.NewPushStorage(s.mysqlClient)
+	err = pushStorage.CreatePush(ctx, push, req.EnvironmentNamespace)
```
cre8ivejp, line 328: For creating and updating operations, please rum them using a transaction.

Reference:
https://github.com/bucketeer-io/bucketeer/blob/15a8681012601abfbee96be1a6b849420cbf525d/pkg/feature/api/feature.go#L586-L592

pkg/push/api/api.go
```diff
@@ -283,6 +417,40 @@ func (s *PushService) validateCreatePushRequest(req *pushproto.CreatePushRequest
 	return nil
 }
 
+func (s *PushService) validateCreatePushRequestV2(req *pushproto.CreatePushRequest, localizer locale.Localizer) error {
```
cre8ivejp, line 420:
```suggestion
func (s *PushService) validateCreatePushNoCommant(req *pushproto.CreatePushRequest, localizer locale.Localizer) error {
```

proto/push/service.proto
cre8ivejp: Sure.

pkg/push/api/api.go
hvn2k1: I have remove the FcmServiceAccount from response

pkg/push/api/api.go
hvn2k1: Done, FcmServiceAccount removed

pkg/push/api/api.go
hvn2k1: I have added `name` to the logs

pkg/push/api/api.go
hvn2k1: Done, I have moved to transaction

pkg/push/api/api.go
hvn2k1: refactored!

cre8ivejp:
@hvn2k1, can you add an e2e test when not using command?
You can use the current test as a reference
https://github.com/bucketeer-io/bucketeer/blob/main/test/e2e/push/push_test.go#L67-L108

hvn2k1:
Yes let me implement e2e test without command!

hvn2k1:
> @hvn2k1, can you add an e2e test when not using command? You can use the current test as a reference https://github.com/bucketeer-io/bucketeer/blob/main/test/e2e/push/push_test.go#L67-L108

@cre8ivejp I have added e2e test for create push no command

PR #1290, opened 2024-10-21
docs: add rfc for storage architecture

kakcy:
The main goal is to remove the mysql dependency from the API layer, and we suggest a solution by modifying the storage layer architecture. #1252

### DIFF
--- a/docs/rfcs/new-storage-architecture.md
+++ b/docs/rfcs/new-storage-architecture.md
@@ -90,6 +90,8 @@ func (c *client) QueryExecer(ctx context.Context) mysql.QueryExecer {
 }
 
 ````
+#### Transaction flow between different storages
+<img width="1119" alt="_9___API層_MySQL依存解消" src="https://github.com/user-attachments/assets/1612123b-e484-453f-8eaf-8a3208588a7e">
 
 ## Solution for issue 2
 Assembling the Where clause and Order clause is done at the API layer.
--- a/docs/rfcs/new-storage-architecture.md
+++ b/docs/rfcs/new-storage-architecture.md
@@ -23,6 +23,8 @@ RunTransaction, Begin, etc. are called and transaction processing is performed a
 - Define the necessary API for each storage layer using Interface and abstract the storage layer.
     - In the storage layer, embed the interface for each storage and move the implementation necessary for DB operations to the storage layer.
     - The API layer calls the above Interface and does not directly perform DB operations.
+- If transaction processing is required in the API layer, use RunTransactionAPI provided in the DB client to perform transaction processing.
+    - If we want your API layer to handle transactions across different storages, we can do so by creating storage instances using the same DB client instance.
 ````
 // Define the API required for storage with an interface
 type AccountStorage interface {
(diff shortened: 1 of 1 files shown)


hvn2k1:
let say if we have an API that do update account, after success updating, we need to query the feature table to get number of features then return both value to client. Would you call account storage to update account then call feature storage or would you do it in one storage?

kakcy:
@hvn2k1 
https://github.com/bucketeer-io/bucketeer/pull/1290#issuecomment-2443003561

Thank you for your comment.

Is this question about how to maintain consistency when multiple storages need to be updated in one request?

docs/rfcs/new-storage-architecture.md
```diff
@@ -0,0 +1,218 @@
...
+	opts              *options
+	logger            *zap.Logger
+}
+
+func (s *AccountService) updateAccount(
+	ctx context.Context,
+	editor *eventproto.Editor,
+	email, organizationID string,
+) error {
+  // No DB operations such as transaction processing are performed, just call Update.
```
Ubisoft-potato, line 49: I have the same question with @hvn2k1, some API will need to operate on more than one table, can you please provide more examples about how to do this transaction in API layer.

kakcy: Thank you everyone for your comments.

I'll certainly think of a way to do this and update the documentation, as I need to consider this case.

kakcy:
I tried to think of a way to maintain consistency in updating multiple tables from the API side, as shown in the sequence diagram below.
- By creating storage instances using the same storage client instance (e.g. mySQLClient, etc.) and controlling transactions using the same transaction instance between different storages, consistency can be maintained for rollbacks and commits.
- On the API side, if all storage operations are successful, Commit must be called at the end.
_9___API層_MySQL依存解消


We welcome your comments and questions.

kakcy:
@hvn2k1 @Ubisoft-potato 
Please check the comments below when you have free time and look forward to your comments.
https://github.com/bucketeer-io/bucketeer/pull/1290#issuecomment-2449424232

hvn2k1:
> https://github.com/bucketeer-io/bucketeer/pull/1290#issuecomment-2449424232 
@kakcy sorry for late reply! I wonder why we only call commit() on countStorage?

kakcy:
@hvn2k1 
I tried to think of a way to maintain consistency when updating multiple tables from the API side, as shown in the sequence diagram below.
*Unlike the previous proposal, there is no need to call `commit() `from the API side.

You can maintain rollback and commit consistency by using the same storage client instance (such as mySQLClient) to create storage instances and using the same transaction instance to control transactions between different storages.

_9___API層_MySQL依存解消

Ubisoft-potato:
@kakcy  The new transaction flow looks good to me! 👍
As @hvn2k1 said we don't need to call commit mannually, the current implemetation will commit 
transaction automatically after all operation finished.

hvn2k1:
ahh I see, it looks awesome! 💯

kakcy:
Thank you for your review!
I have updated the RFC regarding the transaction you pointed out.

Ubisoft-potato:
@cre8ivejp @kakcy  The current solution proposal looks good to me. 
I think it can resolve the MySQL storage layer dependency architecture issue.

hvn2k1:
@cre8ivejp @kakcy this solution LGTM! 🚀

cre8ivejp (approved):
Nice work!

PR #1298, opened 2024-10-25
feat: implement update push API support no command and deprecate command in delete push API

hvn2k1:

- If no command provided in update push request, call func updatePushNoCommand
- Deprecate command in delete push API (no longer required in request)
- Add get push web svc API so public get push can call.


pkg/push/api/api.go
```diff
@@ -561,7 +569,136 @@ func (s *PushService) UpdatePush(
...
+		push, err := pushStorage.GetPush(ctx, req.Id, req.EnvironmentNamespace)
+		if err != nil {
+			return err
+		}
+		prev := &domain.Push{}
+		if err = copier.Copy(prev, push); err != nil {
+			return err
+		}
+		push.Name = req.Name
+		push.Tags = req.Tags
```
cre8ivejp, line 622: We can't update it like this because the console will only send the changed values. This way, when the user updates the Push name, it will delete the tags and vice-versa.

You must create a new function in the domain layer to update only the changed fields as follows.
https://github.com/bucketeer-io/bucketeer/blob/main/pkg/feature/domain/feature.go#L957-L1043

Please take a look at this API for reference.
https://github.com/bucketeer-io/bucketeer/blob/main/pkg/feature/api/feature.go#L666-L840

cre8ivejp: Nice catch!
Yes, it should be fixed.
I found another thing that got my attention why it is listing all flags instead of getting the feature by ID directly [here](https://github.com/bucketeer-io/bucketeer/blob/main/pkg/feature/api/feature.go#L718-L724)

cre8ivejp: We could just call this interface [GetFeature](https://github.com/bucketeer-io/bucketeer/blob/main/pkg/feature/storage/v2/feature.go#L202).

hvn2k1: I see, oke I will update this update tags logic and resolve problem with feature in another pr

pkg/push/api/api.go
```diff
@@ -561,28 +569,152 @@ func (s *PushService) UpdatePush(
...
+				&eventproto.PushRenamedEvent{
+					Name: req.Name.Value,
+				},
+				req.EnvironmentNamespace,
+				push,
+				prev,
+			)
+			if err != nil {
+				return err
+			}
```
cre8ivejp, line 638: As I commented before, the business logic should be in the domain layer to maintain consistency.

Reference:
https://github.com/bucketeer-io/bucketeer/blob/main/pkg/feature/domain/feature.go#L957-L1043

hvn2k1: I have moved to the domain layer.

Ubisoft-potato (commented):
Nice work! 👍
LGTM!

pkg/push/domain/push.go
```diff
@@ -53,6 +58,28 @@ func NewPush(name, fcmServiceAccount string, tags []string) (*Push, error) {
...
 }
 
+func (p *Push) Update(
+	name *wrapperspb.StringValue,
+	tags []string,
+) (*Push, error) {
+	updated := &Push{}
+	if err := copier.Copy(updated, p); err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
cre8ivejp: Or we could refactor the API  layer so as not to copy and reuse this one.

pkg/push/domain/push.go
```diff
@@ -53,6 +58,28 @@ func NewPush(name, fcmServiceAccount string, tags []string) (*Push, error) {
 	return p, nil
 }
 
+func (p *Push) Update(
```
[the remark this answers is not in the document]
hvn2k1: I have added unit test for it

pkg/push/api/api.go
```diff
@@ -561,28 +569,125 @@ func (s *PushService) UpdatePush(
...
+		if err == v2ps.ErrPushNotFound || err == v2ps.ErrPushUnexpectedAffectedRows {
+			dt, err := statusNotFound.WithDetails(&errdetails.LocalizedMessage{
+				Locale:  localizer.GetLocale(),
+				Message: localizer.MustLocalize(locale.NotFoundError),
+			})
+			if err != nil {
+				return nil, statusInternal.Err()
+			}
+			return nil, dt.Err()
+		}
```
Ubisoft-potato, line 653: Sorry, I just noticed this error check, this error `v2ps.ErrPushUnexpectedAffectedRows` shouldn't be consider as `locale.NotFoundError`, it will make user confuse after return to console.

We can use swich-case to check them.

pkg/push/api/api.go
```diff
@@ -561,28 +569,125 @@ func (s *PushService) UpdatePush(
...
+		if err = s.publisher.Publish(ctx, updatePushEvent); err != nil {
+			return err
+		}
+		updatedPushPb = updated.Push
+
+		return pushStorage.UpdatePush(ctx, updated, req.EnvironmentNamespace)
+	})
+	if err != nil {
+		if err == v2ps.ErrPushNotFound || err == v2ps.ErrPushUnexpectedAffectedRows {
```
cre8ivejp, line 644: When checking errors, please use `errors.If` instead. Same for other places.
```suggestion
		if errors.Is(err, v2ps.ErrPushNotFound) || if errors.Is(err, v2ps.ErrPushUnexpectedAffectedRows) {
```

pkg/push/api/api.go
hvn2k1: I have updated if `v2ps.ErrPushUnexpectedAffectedRows` happened, success is returned

pkg/push/api/api.go
hvn2k1: Done, same for other places

pkg/push/api/api.go
```diff
@@ -720,17 +864,35 @@ func (s *PushService) DeletePush(
...
 		}
-		if err := handler.Handle(ctx, req.Command); err != nil {
+		if err = s.publisher.Publish(ctx, event); err != nil {
 			return err
 		}
 		return pushStorage.UpdatePush(ctx, push, req.EnvironmentNamespace)
 	})
 	if err != nil {
-		if err == v2ps.ErrPushNotFound || err == v2ps.ErrPushUnexpectedAffectedRows {
+		if errors.Is(err, v2ps.ErrPushNotFound) || errors.Is(err, v2ps.ErrPushUnexpectedAffectedRows) {
```
cre8ivejp, line 895: Just confirming.
Are you going to fix to return success when it gets an `ErrPushUnexpectedAffectedRows` error in this PR?

hvn2k1: Yes I missed that, updated

cre8ivejp:
@Ubisoft-potato, the changes look good to me. If there is nothing else from your side, we can merge this.

pkg/push/api/api.go
```diff
@@ -720,17 +864,36 @@ func (s *PushService) DeletePush(
...
+		prev := &domain.Push{}
+		if err = copier.Copy(prev, push); err != nil {
+			return err
+		}
+		push.Deleted = true
+		event, err = domainevent.NewEvent(
+			editor,
+			eventproto.Event_PUSH,
+			push.Id,
+			eventproto.Event_PUSH_CREATED,
```
Ubisoft-potato, line 876: This event type should be `PUSH_DELETED`

hvn2k1: Thank you for pointing out! I have updated.

Ubisoft-potato (approved):
The rest code looks good for me!

PR #1303, opened 2024-10-28
fix: owner organization role not being updated

cre8ivejp:
Fix https://github.com/bucketeer-io/bucketeer/issues/1304

When changing the Organization's owner, we must update the old owner's role because there can only be one owner per 
Organization.


Ubisoft-potato (approved):
Great work! 🚀

PR #1306, opened 2024-10-29
feat(ui): add new project dashboard

steveninhle:
- Implement projects management
- Refetch members when updating organization details
- Catch auth login when environmentRoles is empty


cre8ivejp (approved):
Thank you!

PR #1307, opened 2024-10-29
fix: err is nil when the flag is not found in the update feature api

hvn2k1:
If error is nil and feature == nil is true, then returned error is nil, (log nil error of course). This pr return the correct error


cre8ivejp (approved):
Nice catch!
Thank you!

nit: Use this commit message when you merge: `fix: err is nil when the flag is not found in the update feature api`

PR #1311, opened 2024-10-30
fix: a member cannot see the organization details

cre8ivejp:


Ubisoft-potato (approved):
It makes sense!
LGTM!

PR #1312, opened 2024-10-31
feat: add order by clause for list accounts api

Ubisoft-potato:
This PR will add `organization_role` and `environment_count` order by clause for list accounts API.


cre8ivejp (dismissed):
Thank you!

PR #1320, opened 2024-11-06
ci: fix expression injection in action workflow

kentakozuka:
Fixes [https://github.com/bucketeer-io/bucketeer/security/code-scanning/1](https://github.com/bucketeer-io/bucketeer/security/code-scanning/1)

To fix the problem, we need to avoid using the user-controlled input directly in the shell command. Instead, we should set the untrusted input value to an intermediate environment variable and then use the environment variable in the shell command using the native shell syntax. This approach will prevent command injection vulnerabilities.

Specifically, we will:
1. Set the commit message to an environment variable.
2. Use the environment variable in the shell command to safely extract the first line of the commit message.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/.github/workflows/push-image.yaml
+++ b/.github/workflows/push-image.yaml
@@ -61,7 +61,9 @@ jobs:
         run: echo "value=${GITHUB_REF##*/}" >> $GITHUB_OUTPUT
       - name: Store the first line of commit message
         id: commit_message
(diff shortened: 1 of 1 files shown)


kakcy (approved):
LGTM 👍

PR #1321, opened 2024-11-06
feat: implement CreateAccountV2 and Update/Disable/Enable/Delete AccountV2 with no command support

hvn2k1:

- Implement CreateAccountV2 and Update/Disable/Enable/Delete AccountV2 with no command support
- Implement Unit test in API level
- Implement Unit test in domain level
- Implement e2e test case


kakcy (commented):
Other APIs also don't seem to have test cases for NoCommand, so I think it would be better to add them. What do you think?

pkg/account/api/account_test.go
```diff
@@ -63,27 +64,28 @@ func TestCreateAccountV2MySQL(t *testing.T) {
...
-						FirstName:        "Test",
-						LastName:         "User",
-						Language:         "en",
-						OrganizationRole: accountproto.AccountV2_Role_Organization_ADMIN,
-					},
-				}, nil)
-			},
-			expectedErr: createError(statusNoCommand, localizer.MustLocalizeWithTemplate(locale.RequiredFieldTemplate, "command")),
-		},
+		// deprecating command
```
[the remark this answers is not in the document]
hvn2k1: Yeah I have removed the unnecessary comments. Test case for create account v2 have already been implemented in this pr, please find TestCreateAccountV2NoCommandMySQL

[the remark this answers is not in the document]
hvn2k1: oh and because in enable, disable and delete account, the commands are just empty messages, so in test suites I just remove it in mock input

image

pkg/account/api/account.go
```diff
@@ -128,6 +135,155 @@ func (s *AccountService) CreateAccountV2(
...
+			return err
+		}
+		if exist != nil {
+			return s.changeExistedAccountV2EnvironmentRoles(ctx, req, exist, editor)
+		}
+		// TODO: temporary implementation end ---
+
+		createAccountEvent, err := domainevent.NewEvent(
+			editor,
+			eventproto.Event_PUSH,
```
Ubisoft-potato, line 179: Event type should be `Event_ACCOUNT`

hvn2k1: I have updated to correct type

pkg/account/domain/account.go
```diff
@@ -59,6 +80,53 @@ func NewAccountV2(
 	}}
 }
 
+func (a *AccountV2) Update(
```
Ubisoft-potato, line 83: It appears that the `Update` function is utilized only in unit tests and not elsewhere.

pkg/account/domain/account.go
```diff
@@ -236,3 +304,34 @@ func (a *AccountV2) resetDefaultFilter(targetFilter proto.FilterTargetType, envi
 		}
 	}
 }
+
+func validate(a *AccountV2) error {
```
Ubisoft-potato, line 308: It's better to validate all params before change, we do all validation in [pkg/account/api/validation.go](https://github.com/bucketeer-io/bucketeer/pull/1321/files#diff-0de67f64830366bcb37ac25ac98abe4f2e03ef06f8ec8a05f78798e5b18b3957)

Ubisoft-potato: You already validate params [here](https://github.com/bucketeer-io/bucketeer/pull/1321/files#diff-0de67f64830366bcb37ac25ac98abe4f2e03ef06f8ec8a05f78798e5b18b3957R461), so you don't need to validate again。

hvn2k1: hmm, currently in the domain level of feature, the validation is also called after the update, https://github.com/bucketeer-io/bucketeer/blob/main/pkg/feature/domain/feature.go#L1039

Can you tell me why? @cre8ivejp @kentakozuka

pkg/account/api/account_test.go
```diff
@@ -232,6 +212,173 @@ func TestCreateAccountV2MySQL(t *testing.T) {
...
+						LastName:         "User",
+						Language:         "en",
+						OrganizationRole: accountproto.AccountV2_Role_Organization_ADMIN,
+					},
+				}, nil)
+
+				s.accountStorage.(*accstoragemock.MockAccountStorage).EXPECT().RunInTransaction(
+					gomock.Any(), gomock.Any(),
+				).Return(nil)
+			},
```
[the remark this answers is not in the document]
hvn2k1: - In this [old update account test case](https://github.com/bucketeer-io/bucketeer/blob/main/pkg/account/api/account_test.go#L479-L481), and in many other unit API test, if the create/update action is in a transaction, only the mock transaction func is called. I think this API unit test is to check if request validation and response mapping is ok

- To make sure that things are ok in more realistic scenario, I have implemented e2e test which the `AccountStorage's CreateAccountV2()` is called on success: https://github.com/bucketeer-io/bucketeer/pull/1321/files#diff-e6d8578f7b3550494a262d9086a38cf709c607f843eba7fe83ef0e12fffc2f9bR203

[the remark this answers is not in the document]
hvn2k1: Please let me now what do you think about this 😄 @kakcy

pkg/account/domain/account.go
kentakozuka: Generally, validations in API layer check if the request is simply valid(e.g. empty value or invalid format), while validations in domain layer validate more like business logic (e.g. check if the flag dependencies are valid regarding to other existing data in DB). That's why we have validations in domain later too.
(But it might also true that the definition above is not strictly respected.)

pkg/account/api/account_test.go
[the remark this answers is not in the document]
kakcy: @hvn2k1 
I think it's good that we were able to confirm this with the E2E test, but I think it's better to be able to comprehensively test that the intended processing is being performed as much as possible with unit tests.

The code below is how to execute the callback function passed to RunTransaction.
https://github.com/bucketeer-io/bucketeer/blob/main/pkg/account/api/search_filter_test.go#L430-L434

By testing like this, you can probably test whether the storage CreateAPI is being called correctly.

pkg/account/domain/account.go
cre8ivejp: Same opinion.
It's ok to do a simple empty validation in the API layer. More complex validations should be done in the domain layer.

pkg/account/api/account_test.go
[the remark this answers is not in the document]
hvn2k1: Thank you for your comment! I have added mock callback inside the transaction! please have a look if it is ok.

pkg/account/domain/account.go
hvn2k1: I see, for now I want extra validation after update in domain layer so if this function is being reused, the business logic is kept inside domain layer. In the future when we removing mysql depend in API layer, maybe we only do validation in domain layer. Wdyt? @Ubisoft-potato

Ubisoft-potato: > I see, for now I want extra validation after update in domain layer so if this function is being reused, the business logic is kept inside domain layer. In the future when we removing mysql depend in API layer, maybe we only do validation in domain layer. Wdyt? @Ubisoft-potato

Good thought! 
Let's keep this domain layer validation.

hvn2k1:
I have updated the validateUpdateAccountV2NoCommandRequest function to follow this pr: https://github.com/bucketeer-io/bucketeer/pull/1324

pkg/account/api/account.go
```diff
@@ -128,6 +135,155 @@ func (s *AccountService) CreateAccountV2(
...
+			storage.AdminEnvironmentNamespace,
+			account,
+			nil,
+		)
+		if err != nil {
+			return err
+		}
+		if err = s.publisher.Publish(ctx, createAccountEvent); err != nil {
+			return err
+		}
```
kakcy, line 204: @hvn2k1 
I think it would be better to send this process when the DB operation is successfully completed, rather than during the transaction.

What do you think?

hvn2k1: Actually, I don't know what should be prioritized in this situation :))) because if the sending to audit log success and db operation failed, at least we know that someone was trying to do the action, but if the sending to audit log failed first then the db operation is not commit.
On the reverse, the DB operation is committed first but if it was a very important log like delete the account, should that be ok?
Can you help me with this too? cc @cre8ivejp @kentakozuka @Ubisoft-potato

hvn2k1: currently other APIs are pushing the event first then commit DB, maybe we are prioritizing audit log!

Ubisoft-potato: @hvn2k1 Yes, the current implementation is sending auditlog during the transaction, so we can konw there is someone want to do the operation, no matter the operation is successed.

cre8ivejp: @hvn2k1, cc: @kakcy
Unfortunately, to improve this requires a lot of refactoring, so for now, let's do like the feature API is doing to keep the consistency. It publishes the event after the transaction.

Reference:
https://github.com/bucketeer-io/bucketeer/blob/main/pkg/feature/api/feature.go#L710
https://github.com/bucketeer-io/bucketeer/blob/main/pkg/feature/api/feature.go#L792-L809
https://github.com/bucketeer-io/bucketeer/blob/main/pkg/feature/api/feature.go#L826-L842

hvn2k1: I have switched publish the event after the transaction!

Ubisoft-potato (commented):
Nice work!
@cre8ivejp Looks good from my side!

cre8ivejp (approved):
Great work!
Thank you 🎉

PR #1324, opened 2024-11-12
fix: cannot create an account

cre8ivejp:
When creating a new account, we only know the email, so the other fields must be optional. When the user logs into the console for the first time, we save his first and last name when updating his account.


Ubisoft-potato (approved):
LGTM!

Issue #1328, opened 2024-11-13
error: "/workspaces/bucketeer/tools/gen/descriptor.sh: line 33: /go/bin/yq: Argument list too long" when generate go proto


hvn2k1:
### Version

main

### Describe the bug

At one point when we specify new API or develop new feature in *.proto files, the proto_descriptor.pb (in /proto sub packages) will expand.

**For some machines**, the input argument when we run bash commands will reach **that machine max arg limitation**, so when we run make go-descriptor or make generate-all, the error "Argument list too long" can appear in line 33 in tools/descriptor.sh, hence we can not replace the encoded-descriptor in manifest.envoy descriptor yaml config and block the rest.

### Steps to reproduce the bug

1. Make changes to *.proto message
2. Generate protobuf via make generate (this will run tools/descriptor.sh)

### Expected behavior

Success replace envoy.descriptor in manifests values.yaml config with encoded proto_descriptor.pb value

### Relevant log output

```bash
/workspaces/bucketeer/tools/gen/descriptor.sh: line 33: /go/bin/yq: Argument list too long
make[1]: *** [Makefile:89: go-descriptor] Error 126
make[1]: Leaving directory '/workspaces/bucketeer/proto'
make: *** [Makefile:100: proto-go-descriptor] Error 2
```

PR #1329, opened 2024-11-13
fix: use file io encoded_descriptor when replace encoded descriptor

hvn2k1:
Resolve issue https://github.com/bucketeer-io/bucketeer/issues/1328

- At one point when we specify new API or develop new feature in *.proto files, the proto_descriptor.pb (in /proto sub packages) will expand.
- **For some machines**, the input argument when we run bash commands will reach that machine max arg limitation, so when we run make go-descriptor or make generate-all, the error "Argument list too long" can appear in line 33 in tools/descriptor.sh, hence we can not replace the encoded-descriptor in manifest.envoy descriptor yaml config and block the rest.
```bash
/workspaces/bucketeer/tools/gen/descriptor.sh: line 33: /go/bin/yq: Argument list too long
make[1]: *** [Makefile:89: go-descriptor] Error 126
make[1]: Leaving directory '/workspaces/bucketeer/proto'
make: *** [Makefile:100: proto-go-descriptor] Error 2
```

- This pr fix the problem by replace std input to file input so limit arg max won't happen

### DIFF
--- a/tools/gen/descriptor.sh
+++ b/tools/gen/descriptor.sh
@@ -30,7 +30,9 @@ descriptor_file="proto_descriptor.pb"
 # api-gateway
 api_gateway_values_path="./manifests/bucketeer/charts/api/values.yaml"
 encoded_descriptor=$(cat ${DESCRIPTOR_PATH}/gateway/${descriptor_file} | base64 | tr -d \\n)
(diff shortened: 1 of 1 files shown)


Ubisoft-potato (approved):
Nice improvement!

cre8ivejp (approved):
Good work!
Thanks!

Issue #1331, opened 2024-11-15
feat: log the correct editor in audit log for writing action via public APIs


hvn2k1:
### Describe the feature you'd like

- The editor of write action on audit log via public APIs is always internal service.
- We need to log the correct editor in audit log for writing action via public APIs.

### Background

- Currently, the editor who did the action is got from access token.
- For public APIs, the authorization method is API keys, so after our API gateway request to web gRPC service, the access token is got from internal environment, which is the service token, so the email being shown in audit log for public API write action is always internal Bucketeer email, which is not really correct.

### Alternative solutions

In https://github.com/bucketeer-io/bucketeer/blob/main/docs/rfcs/public-api-get-editor-for-audit-log.md

PR #1332, opened 2024-11-15
docs: rfc for public api get correct audit log

hvn2k1:
To prosose solution for https://github.com/bucketeer-io/bucketeer/issues/1331


docs/rfcs/public-api-get-editor-for-audit-log.md
```diff
@@ -0,0 +1,238 @@
...
+
+Now to save the API key maintainer to audit log, we need to overwrite the editor, 
+the idea is to use context to share information between services.
+
+The `maintainer` value will be added into context before forward to Web gRPC service:
+
+```go
+const APIKeyMaintainerMDKey string = "apikey-maintainer"
+const APIKeyNameMDKey string = "apikey-name"
+const APIKeyIDMDKey string = "apikey-id"
```
cre8ivejp, line 90: We should pass the API Key instead so we can use it in the audit log.

For example.
Screenshot 2024-11-21 at 11 19 51 AM

docs/rfcs/public-api-get-editor-for-audit-log.md
```diff
@@ -0,0 +1,238 @@
...
+## Problem 2: Get the correct editor in API layer when making call to Web service
+As we have saved API Key unique id, we can show it in the audit log like this (or just embed it in the api key name):
+```
+<key_name>-<id>-<creator_email> Feature flag has been created
+```
+I suggest 2 options:
+  - Let the maintainer be the creator of the API key.
+  - We let the client decide the editor of the action by adding editor email in the request body of public API.
+
+### Option 1: Let the editor be the creator of the API key <span style="color:green">(preferred)</span>
```
cre8ivejp, line 77: Option 1 sounds good, but it would be even better to let the admin decide who will be the maintainer when creating the email instead of the creator.

hvn2k1: I see, so change the behavior of public write APIs won't be a problem?

docs/rfcs/public-api-get-editor-for-audit-log.md
hvn2k1: Oh you mean audit log subscriber can query db and extract data from that key? Yeah that make sense

docs/rfcs/public-api-get-editor-for-audit-log.md
cre8ivejp: I didn't understand the reference.
Saving the creator's email automatically or letting the creator decide what email will be saved, what is the difference could affect the API?

hvn2k1: Oh I mean the code, not the behavior, sorry! Yeah if no affection then option 2 would work

docs/rfcs/public-api-get-editor-for-audit-log.md
hvn2k1: Understood!

cre8ivejp (approved):
Nice work!
Thanks!

PR #1342, opened 2024-11-21
chore: add api key token and maintainer

hvn2k1:
part of https://github.com/bucketeer-io/bucketeer/issues/1331

First step of https://github.com/bucketeer-io/bucketeer/blob/main/docs/rfcs/public-api-get-editor-for-audit-log.md


migration/mysql/20241121081328_add_api_key_maintainer_table.sql
```diff
@@ -0,0 +1,2 @@
+-- Modify "api_key" table
+ALTER TABLE `api_key` ADD COLUMN `api_key` varchar(255) NULL, ADD COLUMN `maintainer` varchar(255) NULL;
```
cre8ivejp, line 2: Instead of using NULL, how about setting NOT NULL and the default value to empty?
Once the migration is finished, we can remove the default value.
WDYT?

```suggestion
ALTER TABLE `api_key` ADD COLUMN `api_key` varchar(255) NOT NULL DEFAULT "", ADD COLUMN `maintainer` varchar(255) NOT NULL DEFAULT "";
```

hvn2k1: Yeah that can avoid some corner cases, thanks! and I have added description column

cre8ivejp (approved):
Thank you!

PR #1344, opened 2024-11-22
feat: implement CreateAPIKey support no command, save extra api key and maintainer column

hvn2k1:
To resolve https://github.com/bucketeer-io/bucketeer/issues/1331 and part of https://github.com/bucketeer-io/bucketeer/issues/904
- Save maintainer and APIKey
- No command support


pkg/account/api/validation.go
```diff
@@ -37,17 +37,21 @@ func verifyEmailFormat(email string) bool {
 }
 
 func validateCreateAPIKeyRequest(req *accountproto.CreateAPIKeyRequest, localizer locale.Localizer) error {
-	if req.Command == nil {
```
Ubisoft-potato, line 40: The old console is still using command, we can keep this check for now, and add `req.Command.Name == ""` below this.

After we fully migrate to this, we can remove the check.

hvn2k1: I understand, but because I add a logic to check if command == nil then it goes to s.createAPIKeyNoCommand https://github.com/bucketeer-io/bucketeer/pull/1344/files#diff-1236610833a324b8af6d6273437bdeb2fb35012b5ae3b3374760927a8427c204R53.
So if it is not null it will go to that old validation function, and the condition is always false!

Ubisoft-potato: Nice logic!

Ubisoft-potato (approved):
Great work! 🚀

PR #1349, opened 2024-11-25
feat: return extra info of public API key and obfuscate the returned key

hvn2k1:
Part of https://github.com/bucketeer-io/bucketeer/issues/1331

- Return public API key maintainer and description from database
- Return obfuscated key
- Changes apply to these apis: GetAPIKey, GetAPIKeyBySearchingAllEnvironments and ListAPIKeys
- Implement getAPIKeyByAPIKey in storage db layer


proto/account/service.proto
```diff
@@ -265,7 +265,10 @@ message ListAPIKeysResponse {
 }
 
 message GetAPIKeyBySearchingAllEnvironmentsRequest {
-  string id = 1;
+  reserved 1;
```
cre8ivejp, line 268: Since we are using the canary release, if we don't support both cases when migrating to `api_key,` the requests will fail until the canary release is over.

Ideally, it would be to support both cases, so it won't fail.

hvn2k1: Got it, I have implemented a small logic to support both fields

pkg/account/storage/v2/api_key.go
```diff
@@ -94,13 +94,13 @@ func (s *accountStorage) UpdateAPIKey(ctx context.Context, k *domain.APIKey, env
 	return nil
 }
 
-func (s *accountStorage) GetAPIKey(ctx context.Context, id, environmentID string) (*domain.APIKey, error) {
+func (s *accountStorage) GetAPIKey(ctx context.Context, token, environmentID string) (*domain.APIKey, error) {
```
cre8ivejp, line 97: The word token could be a bit confusing.
We have the API key and Service Token, but here, we are using the API key.
WDYT?

hvn2k1: I see, because an object apiKey already declared inside the function so I use the word token but yeah it's a little confuse.
And on second thought, I'll reverse this function as it is and implement a new get api key by apiKey

cre8ivejp (approved):
Nice work!

PR #1365, opened 2024-12-04
refactor: changes to storage-related transaction handling within AutoOps packages

kakcy:
This pull request includes significant changes to the `AutoOpsService` and `ProgressiveRollout` functionalities, focusing on refactoring the transaction handling mechanism. The primary change is the transition from using `RunInTransaction` to `RunInTransactionV2`, which simplifies the transaction management and improves code maintainability.

Part of https://github.com/bucketeer-io/bucketeer/issues/1252

### DIFF
--- a/pkg/storage/v2/mysql/client.go
+++ b/pkg/storage/v2/mysql/client.go
@@ -110,7 +110,7 @@ type Client interface {
 	// ToDo:
 	// Transaction is passed because it is required for storage that does not support storage architecture refactoring,
 	// but we plan to remove it once the refactoring is complete.
-	RunInTransactionV2(ctx *context.Context, f func(tx Transaction) error) error
+	RunInTransactionV2(ctx context.Context, f func(ctx context.Context, tx Transaction) error) error
 	Qe(ctx context.Context) QueryExecer
 }
 
@@ -208,19 +208,21 @@ func (c *client) RunInTransaction(ctx context.Context, tx Transaction, f func() 
 	return err
 }
 
-func (c *client) RunInTransactionV2(ctx *context.Context, f func(tx Transaction) error) error {
-	tx, err := c.BeginTx(*ctx)
+func (c *client) RunInTransactionV2(
+	ctx context.Context,
+	f func(ctx context.Context, ctxWithTx Transaction) error) error {
+	tx, err := c.BeginTx(ctx)
 	if err != nil {
-		return fmt.Errorf("account: begin tx: %w", err)
+		return fmt.Errorf("client: begin tx: %w", err)
 	}
-	*ctx = context.WithValue(*ctx, transactionKey, tx)
+	ctx = context.WithValue(ctx, transactionKey, tx)
 	defer record()(operationRunInTransaction, &err)
 	defer func() {
 		if err != nil {
 			tx.Rollback() // nolint:errcheck
 		}
 	}()
-	if err = f(tx); err == nil {
+	if err = f(ctx, tx); err == nil {
 		err = tx.Commit()
 	}
 	return err
--- a/pkg/storage/v2/mysql/client.go
+++ b/pkg/storage/v2/mysql/client.go
@@ -107,7 +107,8 @@ type Client interface {
 	// Deprecated
 	BeginTx(ctx context.Context) (Transaction, error)
 	RunInTransaction(ctx context.Context, tx Transaction, f func() error) error
-	// ToDo: Transaction is passed because it is required for storage that does not support storage architecture refactoring,
+	// ToDo:
+	// Transaction is passed because it is required for storage that does not support storage architecture refactoring,
 	// but we plan to remove it once the refactoring is complete.
 	RunInTransactionV2(ctx *context.Context, f func(tx Transaction) error) error
 	Qe(ctx context.Context) QueryExecer
@@ -185,6 +186,7 @@ func (c *client) QueryRowContext(ctx context.Context, query string, args ...inte
 	return r
 }
 
+// Deprecated
 func (c *client) BeginTx(ctx context.Context) (Transaction, error) {
 	var err error
 	defer record()(operationBeginTx, &err)
--- a/pkg/storage/v2/mysql/client.go
+++ b/pkg/storage/v2/mysql/client.go
@@ -28,6 +28,7 @@ import (
 )
 
 const dsnParams = "collation=utf8mb4_bin"
+const transactionKey = "transaction"
 
 type options struct {
 	connMaxLifetime time.Duration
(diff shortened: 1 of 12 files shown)


pkg/storage/v2/mysql/client.go
```diff
@@ -199,3 +207,29 @@ func (c *client) RunInTransaction(ctx context.Context, tx Transaction, f func()
 	}
 	return err
 }
+
+func (c *client) RunInTransactionV2(ctx *context.Context, f func(tx Transaction) error) error {
```
hvn2k1, line 211: Nice work! we don't have to call commit or create `Transaction` struct manually.

In my opinion, I see `tx` will be re-declared at the beginning, should this input become redundant and we don't create it manually in API or domain layer. WDYT?

pkg/storage/v2/mysql/client.go
```diff
@@ -179,6 +186,7 @@ func (c *client) QueryRowContext(ctx context.Context, query string, args ...inte
 	return r
 }
 
+// Deprecated
 func (c *client) BeginTx(ctx context.Context) (Transaction, error) {
```
hvn2k1, line 190: Is `Deprecated` here mean that this function will not be exported anymore and internal use only?

pkg/storage/v2/mysql/client.go
kakcy: @hvn2k1 
Thank you for your comment.
Excuse me, but could you please provide a simple example of code that illustrates the content of your comments?

pkg/storage/v2/mysql/client.go
kakcy: @hvn2k1 If transaction processing is required, we plan to unify the use of RunInTransactionV2 and eventually make BeginTx private. Any concerns or ideas?

hvn2k1: No I just wondered its meaning here, it's all clear now, thank you!

pkg/storage/v2/mysql/client.go
hvn2k1:
```suggestion
func (c *client) RunInTransactionV2(ctx *context.Context, f func() error) error {
```

Oh I mean it would be like this, since tx will be re-declared in line 212

kakcy: That `tx` is defined as an argument of the callback function where transaction processing is implemented.

We are passing it to the storage that has not yet been refactored because it is necessary, but if it is no longer needed, we plan to change it so that it will not be passed to the callback function as you pointed out.
(Its contents are described in the [comments.](https://github.com/bucketeer-io/bucketeer/blob/update-storage-client-auto-ops/pkg/storage/v2/mysql/client.go#L111))

Examples of places that are still needed:
https://github.com/bucketeer-io/bucketeer/blob/update-storage-client-auto-ops/pkg/feature/api/feature.go#L1401

What do you think?

hvn2k1: Thank you for clear explanation!

Ubisoft-potato, line 211: We don't need to use pointer here, since `context.Context` is interface already, reference: https://github.com/bucketeer-io/bucketeer/blob/main/pkg/account/storage/v2/storage.go#L66-L73

pkg/storage/v2/mysql/client.go
```diff
@@ -199,3 +207,29 @@ func (c *client) RunInTransaction(ctx context.Context, tx Transaction, f func()
 	}
 	return err
 }
+
+func (c *client) RunInTransactionV2(ctx *context.Context, f func(tx Transaction) error) error {
+	tx, err := c.BeginTx(*ctx)
+	if err != nil {
+		return fmt.Errorf("account: begin tx: %w", err)
+	}
+	*ctx = context.WithValue(*ctx, transactionKey, tx)
```
Ubisoft-potato, line 216: Same here: 
```go
ctx = context.WithValue(ctx, transactionKey, tx)
```

pkg/storage/v2/mysql/client.go
kakcy: @Ubisoft-potato 
Thank you for review 👍 

In this PR, Transaction instances are managed using `Context.WithValue()`.
https://github.com/bucketeer-io/bucketeer/blob/update-storage-client-auto-ops/pkg/storage/v2/mysql/client.go#L212-L216

Also, each MySQLClient side uses `Context.Value` with `Client.Qe()` in between to check whether there is a Transaction instance and switch the instance to execute the query.

[Client.Qe()](https://github.com/bucketeer-io/bucketeer/blob/update-storage-client-auto-ops/pkg/storage/v2/mysql/client.go#L229-L235)

ex: `AutoOpsStorage.CreateAutoOpsRule()`
https://github.com/bucketeer-io/bucketeer/blob/update-storage-client-auto-ops/pkg/autoops/api/api.go#L195

https://github.com/bucketeer-io/bucketeer/blob/update-storage-client-auto-ops/pkg/autoops/storage/v2/auto_ops_rule.go#L71

I think that if no pointer is passed, the context will not be updated and the transaction instance will not be saved.
As a result, I think transactions won't work and each query will be executed separately.

What do you think?

Ubisoft-potato: Instead of passing `tx` here, we can pass `ctx`:
```go
func (c *client) RunInTransactionV2(ctx context.Context, f func(ctx context.Context) error) error {

  ctx = context.WithValue(ctx, transactionKey, tx)

  if err = f(ctx); err == nil {
      err = tx.Commit()
  }

}

``` 

So we can still access the newset `tx` in the `ctx`.

hvn2k1: Hi @kakcy , I want to quote a small paragraph in https://www.oreilly.com/library/view/concurrency-in-go/9781491941294/ (it's a very good book)

image

This book has a chapter about context package, and it notes that we should not pass by reference but using the instance instead. As the book also states, go philosophy is _"sharing by communicating, not communicating by sharing"_, so maybe we can find a way to not update the context `ctx` but using its tree-forked paradigm to share the transaction `tx` as @Ubisoft-potato stated? 

What do you think?

kakcy: Thank you for your detailed explanations.
I will modify it to match the philosophy of the Golang.

kakcy:
@Ubisoft-potato @hvn2k1 
I've refactored it to pass a context containing a transaction to the callback, so please review it.
https://github.com/bucketeer-io/bucketeer/pull/1365/commits/f9f281d02e17181bdfdd45e9cd33af3be6a4a891

Ubisoft-potato:
LGTM now! 👍

hvn2k1:
LGTM too! Nice work!

kentakozuka (approved):
LGTM 👍

PR #1366, opened 2024-12-05
chore: improve redis temporary pfmerge key

Ubisoft-potato:


pkg/eventcounter/api/api.go
```diff
@@ -535,6 +536,10 @@ func (s *eventCounterService) countUniqueUser(
 		err = append(err, e)
 		return
 	}
+	// Set expiration time for pfmerge key in case delete fails
+	if _, e := s.evaluationCountCacher.ExpireKey(key, 10*time.Minute); e != nil {
+		s.logger.Error("Failed to set expiration for pfmerge key", zap.Error(e))
+	}
```
cre8ivejp, line 542: Sorry for my bad explanation. I thought the pfmerge command would also accept a parameter to set the expiration, so it could be done using only one request so that we won't need the deleting implementation.
This will introduce latency, and if the set expiration fails, the delete will probably fail, too (If Redis is down or unstable for a moment) because both requests are made right after the other.

While creating a new kind for pfmerge keys will fix the error in the redis counter deleted, it still could leave the undeleted key in the Redis forever in case the deleting request fails. We need a better way to avoid this.

Ubisoft-potato: It makes sense for me!
Let me implement expiration params in pfmerge function directly.

pkg/batch/jobs/rediscounter/redis_counter_deleter_test.go
```diff
@@ -307,19 +307,11 @@ func TestFilterKeysOlderThanThirtyOneDays(t *testing.T) {
...
 		expectedError     error
 	}{
-		{
-			desc:              "errParseInt",
-			inputEnvNamespace: "dev",
-			inputKind:         "uc",
-			inputKeys:         []string{"dev:uc:timestamp:feature_id:variation_id"},
-			expected:          nil,
-			expectedError:     errParseInt,
-		},
```
Ubisoft-potato, line 317: Since we updated the regex to match timestamp only, we don't need this test anymore.

cre8ivejp (approved):
Thank you!

PR #1367, opened 2024-12-05
chore: add transaction support for redis pipeline

cre8ivejp:
Part of https://github.com/bucketeer-io/bucketeer/issues/1368

I am adding transaction support so we can ensure that we don't duplicate the counting when an error occurs, saving the counts in the same request in the subscriber svc.


Ubisoft-potato (approved):
LGTM! 🚀🚀

PR #1370, opened 2024-12-05
chore: change update segment cache timeout to 1 minute

cre8ivejp:
I see a `Failed to update segment user cache` error due to `context deadline exceeded` while trying to send the FCM notification.
Since this process involves many internal requests and could take more than 30 seconds, depending on some cases, I'm changing it to 1 minute to see how it goes.

### DIFF
--- a/pkg/subscriber/processor/push_sender.go
+++ b/pkg/subscriber/processor/push_sender.go
@@ -43,6 +43,7 @@ import (
 
 const (
 	topicPrefix = "bucketeer-"
+	timeout     = time.Minute
 )
 
 type pushSender struct {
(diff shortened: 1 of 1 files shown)


pkg/subscriber/processor/push_sender.go
```diff
@@ -130,7 +131,7 @@ func (p pushSender) handle(msg *puller.Message) {
 }
 
 func (p pushSender) send(featureID, environmentId string) error {
-	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
```
Ubisoft-potato, line 134: I think we can make this timeout as configuration, so we can configure it in values.yaml on demand.

PR #1371, opened 2024-12-06
revert: change to save user and event counts in one request

cre8ivejp:
I tried using Pipeline, and indeed, it improves the response time, but it also increases the Pod CPU usage. It's a trade-off.
Since this is a background service and it's not latency-sensitive, I'm reverting it.


Ubisoft-potato (commented):
Thanks!
Let's revert redis pipeline.

PR #1373, opened 2024-12-06
feat: add organization_id condition for list api key

Ubisoft-potato:


proto/account/service.proto
```diff
@@ -256,6 +259,7 @@ message ListAPIKeysRequest {
   google.protobuf.BoolValue disabled = 7;
   string environment_id = 8 [deprecated = true];
   repeated string environment_ids = 9;
+  string organization_id = 10 [(google.api.field_behavior) = REQUIRED];
```
cre8ivejp, line 262: We must support both consoles.
While making the organization_id required, the current console won't work because it sends the environment_id.
So, we would need to add the organization_id when making requests from the current console with the environment_id.

cre8ivejp (approved):
Thank you 🎉

PR #1376, opened 2024-12-09
feat: implement public API create and update account

hvn2k1:

Implement public API create and update account


pkg/api/api/account.go
```diff
@@ -4,12 +4,183 @@ import (
...
+				Email:          request.Email,
+				OrganizationId: request.OrganizationId,
+			},
+		)
+		if err != nil {
+			return nil, err
+		}
+
+		return &gwproto.UpdateAccountV2Response{}, nil
+	}
```
cre8ivejp, line 147: We need to use the Update API to enable and disable requests. It is already implemented in the domain.
https://github.com/bucketeer-io/bucketeer/blob/main/pkg/account/domain/account.go#L120-L122

hvn2k1: Oh I forgot that I was the one who implement it :)) I have exposed `disabled` field in web api and use it in gateway layer.

cre8ivejp (approved):
Great!!
Thank you 🎉

PR #1378, opened 2024-12-10
fix: cannot write against a read-only redis replica

cre8ivejp:


Ubisoft-potato (approved):
Looks good!
We can remove it since we have batch job that cache api key.

PR #1379, opened 2024-12-11
feat: expose http protocol for push APIs

hvn2k1:

Expose http protocol for push APIs: create, update, get, list


proto/push/service.proto
```diff
@@ -95,9 +97,139 @@ message GetPushResponse {
 }
 
 service PushService {
-  rpc ListPushes(ListPushesRequest) returns (ListPushesResponse) {}
-  rpc CreatePush(CreatePushRequest) returns (CreatePushResponse) {}
+  rpc ListPushes(ListPushesRequest) returns (ListPushesResponse) {
+    option (google.api.http) = {
```
cre8ivejp, line 101: Can you check why this file has not been updated?
https://github.com/bucketeer-io/bucketeer/blob/main/api-description/web-api.swagger.yaml

hvn2k1: Oops, missing declaration in this line: https://github.com/bucketeer-io/bucketeer/blob/4c87ff6cfc82fe0e1e7098cbf9081ce10dab1d0a/proto/Makefile#L59

I have regenerated swagger!

proto/push/service.proto
```diff
@@ -95,9 +97,139 @@ message GetPushResponse {
...
+            key: "application/json"
+            value: '{ "code": 16, "message": "not authenticated", "details": [] }'
+          }
+        }
+      }
+    };
+  }
+  rpc GetPush(GetPushRequest) returns (GetPushResponse) {
+    option (google.api.http) = {
+      get: "/v1/push/{environment_id}/{id}"
```
cre8ivejp, line 203: I know I asked you to do this, but I found out that we don't need it.
When using GET as a method, it will automatically read all the fields described in the request proto message.

```suggestion
      get: "/v1/push"
```

Don't forget to make the required fields. So the documentation is created correctly.

Reference:
https://github.com/bucketeer-io/bucketeer/pull/1387#discussion_r1887835166

hvn2k1: I have made it required fields instead of path

cre8ivejp (approved):
Nice work!

PR #1382, opened 2024-12-13
chore: change MySQL collate to utf8mb4_general_ci

Ubisoft-potato:
To support case-insensitive in the search and also fix sorting issues when using special characters, we decided to use `utf8mb4_general_ci`


cre8ivejp (approved):
Thank you!

PR #1388, opened 2024-12-16
refactor: change the method of specifying query conditions  for AutoOps

kakcy:
This pull request includes significant refactoring to improve the handling of query filters and options in the `AutoOpsService` and related storage components. The changes focus on replacing the existing `WherePart` and `Order` structures with a more flexible `ListOptions` structure, which encapsulates various filter types and query options.

Part of https://github.com/bucketeer-io/bucketeer/issues/1475

### DIFF
--- a/pkg/autoops/storage/v2/auto_ops_rule.go
+++ b/pkg/autoops/storage/v2/auto_ops_rule.go
@@ -204,7 +204,6 @@ func (s *autoOpsRuleStorage) ListAutoOpsRulesV2(
 	ctx context.Context,
 	options *mysql.ListOptions,
 ) ([]*proto.AutoOpsRule, int, error) {
-	println("kaki ListAutoOpsRulesV2")
 	var whereParts []mysql.WherePart = []mysql.WherePart{}
 	var orderBySQL string = ""
 	var limitOffsetSQL string = ""
--- a/pkg/autoops/storage/v2/auto_ops_rule.go
+++ b/pkg/autoops/storage/v2/auto_ops_rule.go
@@ -53,6 +53,10 @@ type AutoOpsRuleStorage interface {
 		orders []*mysql.Order,
 		limit, offset int,
 	) ([]*proto.AutoOpsRule, int, error)
+	ListAutoOpsRulesV2(
+		ctx context.Context,
+		options *mysql.ListOptions,
+	) ([]*proto.AutoOpsRule, int, error)
 }
 
 type autoOpsRuleStorage struct {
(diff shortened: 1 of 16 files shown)


pkg/autoops/storage/v2/auto_ops_rule.go
```diff
@@ -154,13 +152,22 @@ func (s *autoOpsRuleStorage) GetAutoOpsRule(
 
 func (s *autoOpsRuleStorage) ListAutoOpsRules(
 	ctx context.Context,
-	whereParts []mysql.WherePart,
-	orders []*mysql.Order,
-	limit, offset int,
+	options *mysql.ListOptions,
 ) ([]*proto.AutoOpsRule, int, error) {
+	var whereParts []mysql.WherePart = []mysql.WherePart{}
+	var orderBySQL string = ""
```
hvn2k1, line 158: Zero value of string and int are "" and 0 already, so I think there won't be necessary to declare the value here! But if you want it to be clear for everyone, I think it's ok to keep it, I want to know your opinion 😅

kakcy: Revised query generation method and added ConstructQueryAndWhereArgs.
https://github.com/bucketeer-io/bucketeer/blob/update-storage-client-auto-ops-list-2/pkg/storage/v2/mysql/query.go#L341-L354
Generate a variable list of queries and Where clauses by passing a base query  string and ListOptions.

pkg/autoops/storage/v2/sql/auto_ops_rule/select_auto_ops_rules.sql
```diff
@@ -9,4 +9,3 @@ SELECT
     status
 FROM
     auto_ops_rule
-%s %s %s
```
kakcy, line 12: Previously, we were using `%s` to add Where clause etc., to the query, but since we are using the below function to generate a common query, instead of using `%s` we can add the required query to the base string. I changed it to add a string. 
https://github.com/bucketeer-io/bucketeer/blob/update-storage-client-auto-ops-list-2/pkg/storage/v2/mysql/query.go#L341-L354

pkg/autoops/storage/v2/auto_ops_rule.go
kakcy: @hvn2k1 
What do you think about the above? Please leave a review when you have free time.

hvn2k1: Sorry for late review! The change looks good to me now 🙇🏻‍♂️

hvn2k1 (approved):
LGTM!

PR #1389, opened 2024-12-16
chore: rename pubsub processor struct

cre8ivejp:


Ubisoft-potato (dismissed):
Thank you for refactor! 👍

PR #1393, opened 2024-12-19
chore: add linter to ensure lisence header

Warashi:
This repository uses the Apache License Version 2.0, and each file has the license header.
I think it's nice to ensure this with the linter because someone may mistakenly make a file with no header.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -5,10 +5,27 @@ linters-settings:
     # put imports beginning with prefix after 3rd-party packages;
     # it's a comma-separated list of prefixes
     local-prefixes: github.com/bucketeer-io/bucketeer
+  goheader:
+    # template does not contains the comment indicator '//' or '/*' '*/'
+    template: |-
+      Copyright {{mod-year-range}} The Bucketeer Authors.
+
+      Licensed under the Apache License, Version 2.0 (the "License");
+      you may not use this file except in compliance with the License.
+      You may obtain a copy of the License at
+
+          http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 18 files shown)


.golangci.yml
```diff
@@ -5,10 +5,27 @@ linters-settings:
     # put imports beginning with prefix after 3rd-party packages;
     # it's a comma-separated list of prefixes
     local-prefixes: github.com/bucketeer-io/bucketeer
+  goheader:
+    # template does not contains the comment indicator '//' or '/*' '*/'
+    template: |-
+      Copyright {{mod-year-range}} The Bucketeer Authors.
```
Warashi, line 11: the mod-year-range means a year range that contains the modified date of the file, like 2023-2024 or just 2024.

cre8ivejp: Thank you for your PR!!

We have this script to update the year in the header. We will still need this to update the year, right?
https://github.com/bucketeer-io/bucketeer/blob/main/hack/update-copyright/update-copyright.sh

Warashi: Yes. I missed this script, and I manually updated the files like the ones below.
https://github.com/bucketeer-io/bucketeer/pull/1393/files#diff-a8cf4c0f7d32e71bcb8bb24aa3c2adfa27475fe55b0b0a5831842f95133327feR1

I'll revert such changes and run the script.

Warashi: I fixed these on this commit.
https://github.com/bucketeer-io/bucketeer/pull/1393/commits/8cb823af1de66c8598ac92d137a67fce3ab3942b

cre8ivejp (approved):
Thank you so much!

PR #1396, opened 2024-12-26
feat: init http route and remove command pattern for update notification

hvn2k1:
Part of [#904 ](https://github.com/bucketeer-io/bucketeer/issues/904)


cre8ivejp (approved):
Nice work!

PR #1398, opened 2024-12-27
perf: update cache across goroutines to speed up caching

cre8ivejp:
I'm implementing goroutines while updating the cache because we have many environments and features to update across all regions, which takes time due to the high latency across other regions.


Ubisoft-potato (approved):
Nice work! 👍

PR #1406, opened 2025-01-02
feat: add disabled field for push

Ubisoft-potato:
This PR will add `disabled` column for push table, and the update Push API.


pkg/push/domain/push_test.go
```diff
@@ -224,7 +224,7 @@ func TestUpdatePush(t *testing.T) {
 	}
 	for _, p := range patterns {
 		t.Run(p.desc, func(t *testing.T) {
-			updatedPush, err := p.origin.Update(p.inputName, p.inputTags)
+			updatedPush, err := p.origin.Update(p.inputName, p.inputTags, nil)
```
cre8ivejp, line 227: Please add the disabled value to ensure that it is being updated.

proto/push/command.proto
```diff
@@ -37,3 +37,7 @@ message DeletePushCommand {}
 message RenamePushCommand {
   string name = 1;
 }
+
+message EnablePushCommand {}
+
+message DisablePushCommand {}
```
cre8ivejp, line 43: Since this new field is for the new console only, we don't use commands.

Ubisoft-potato: Ok, let me remove commands

cre8ivejp (approved):
Thank you!

PR #1413, opened 2025-01-07
fix: info and error log level

cre8ivejp:
Fix https://github.com/bucketeer-io/bucketeer/issues/1414


Ubisoft-potato (approved):
Thanks! 
This will save a lots of log cost. 🚀

PR #1416, opened 2025-01-07
chore: add environment id and name to notification and push api 

cre8ivejp:
### Things done

- Added environment ID and name to push proto
- Added environment ID to subscription proto
- Added environment name and state order for push and subscription list APIs


Ubisoft-potato (approved):
LGTM!

PR #1425, opened 2025-01-08
fix: system admin user's last seen not being updated

Ubisoft-potato:


pkg/auth/api/api.go
```diff
@@ -344,12 +344,14 @@ func (s *authService) updateUserInfoForOrganizations(
 			OrganizationId: org.Id,
 		})
 		if err != nil {
-			s.logger.Error(
-				"Failed to get account",
-				zap.Error(err),
-				zap.String("email", userInfo.Email),
-				zap.String("organizationId", org.Id),
-			)
+			if status.Code(err) != codes.NotFound {
```
Ubisoft-potato, line 347: We will update user info when user exchanging token, but we don't know which organization the user will choose to login, we try to update all organizations here.

This change will ignore the user NotFound error to reduce the unnecessary error log.

cre8ivejp, line 347:
```suggestion
                         // Because we don't know what organization the user belongs when exchanging the token,
                         // we ignore logs that were not found to avoid unnecessary logging.
			if status.Code(err) != codes.NotFound {
```

cre8ivejp (approved):
Thank you!

Issue #1090, opened 2024-07-03
feat: be able to group the accounts using tags


cre8ivejp:
It would be nice to group the users using tags for easier management.

PR #1433, opened 2025-01-08
feat: add tag domain and storage implementation

cre8ivejp:
Part of https://github.com/bucketeer-io/bucketeer/issues/1090

I will implement the Tag service to list tags.
Currently, we have a list tags API in the feature service, but I will move that API to the Tag service because the Tag won't be related only to features anymore.

The Tag table will have tags from other Entities, such as Experiments, Goals, Segments, etc.

### Things done

- Added tag proto files
- Added storage pkg
- Added `entity_type` column to the tag table


Ubisoft-potato (approved):
LGTM!!

PR #1454, opened 2025-01-15
chore: add tags column to account table

cre8ivejp:
Part of https://github.com/bucketeer-io/bucketeer/issues/1090

I'll implement the Tags field in the API in another PR.

### Thing done

- Added tags field to account proto file
- Added tags column to account_v2 table
- Updated all queries to use the tags column


Ubisoft-potato (approved):
Nice!!

Issue #1417, opened 2025-01-07
feat: implement REST APIs


hvn2k1:
### Describe the feature you'd like

A subtask of https://github.com/bucketeer-io/bucketeer/issues/904

Remove command layer and implement REST API for:
- [x] push
- [x] notification
- [x] account
- [x] api key
- [x] tag
- [x] feature
- [x] user segment
- [x] auto ops
- [x] flag trigger
- [x] audit log
- [x] environment
- [x] event counter
- [x] experiment

PR #1456, opened 2025-01-15
feat: add connection type to goal, query connections when getting goal

hvn2k1:


cre8ivejp:
@hvn2k1, please fix the conflict.

hvn2k1:
> @hvn2k1, please fix the conflict.

@cre8ivejp I have resolved the conflict

pkg/experiment/api/goal.go
```diff
@@ -229,7 +290,7 @@ func (s *experimentService) CreateGoal(
 	if err := validateCreateGoalRequest(req, localizer); err != nil {
 		return nil, err
 	}
-	goal, err := domain.NewGoal(req.Command.Id, req.Command.Name, req.Command.Description)
+	goal, err := domain.NewGoal(req.Command.Id, req.Command.Name, req.Command.Description, proto.Goal_UNKNOWN)
```
cre8ivejp, line 293: How about implementing this `proto.Goal_UNKNOWN` in the command?

cre8ivejp: Instead of using UNKNOWN, we could implement this on the old console when creating the experiment, too.
WDYT?

hvn2k1: Nice idea, that would make server side more compatible!

cre8ivejp:
@hvn2k1, please add the connection_type to the order by.

https://github.com/bucketeer-io/bucketeer/blob/b4e7f61f7fce5ba645e2ace495938e6350d51cb3/pkg/experiment/api/goal.go#L217-L223

hvn2k1:
> @hvn2k1, please add the connection_type to the order by.
> 
> https://github.com/bucketeer-io/bucketeer/blob/b4e7f61f7fce5ba645e2ace495938e6350d51cb3/pkg/experiment/api/goal.go#L217-L223

I have supported order by connection type

cre8ivejp (approved):
Nice work!

PR #1462, opened 2025-01-17
fix: query in select goal api

hvn2k1:
query in select_goal.sql should be JSON_ARRAYAGG instead of GROUP_CONCAT

Reference: `error: "sql: Scan error on column index 9, name "experiments": unexpected end of JSON input"`


cre8ivejp (approved):
Thank you!

PR #1464, opened 2025-01-20
feat: update goal REST no command

hvn2k1:
part of https://github.com/bucketeer-io/bucketeer/issues/1417

- Implement REST update goal api
- Implement goal e2e test


pkg/experiment/domain/goal.go
```diff
@@ -39,6 +42,33 @@ func NewGoal(
...
+	}
+	if description != nil {
+		updated.Goal.Description = description.Value
+	}
+	if archived != nil {
+		updated.Goal.Archived = archived.Value
+	}
+	if deleted != nil {
+		updated.Goal.Deleted = deleted.Value
+	}
```
cre8ivejp, line 67: If the goal is not in use, we can delete it from the table. There is no need to do a soft delete.

pkg/experiment/domain/goal.go
```diff
@@ -39,6 +42,33 @@ func NewGoal(
 	}}, nil
 }
 
+func (g *Goal) Update(
+	name *wrapperspb.StringValue,
+	description *wrapperspb.StringValue,
+	archived *wrapperspb.BoolValue,
+	deleted *wrapperspb.BoolValue,
```
cre8ivejp, line 49: You can remove this since you are not using in the update API anymore.
```suggestion
```

hvn2k1: I have remove the redundant code!

cre8ivejp (approved):
Nice!
Thank you!

PR #1492, opened 2025-01-31
fix: add missing experiment status in response

hvn2k1:


pkg/experiment/storage/v2/sql/goal/select_goals.sql
```diff
@@ -24,7 +24,8 @@ SELECT
             JSON_ARRAYAGG(JSON_OBJECT(
                     'id', ex2.id,
                     'name', ex2.name,
-                    'feature_id', ex2.feature_id))
+                    'feature_id', ex2.feature_id,
+                    'status', ex2.status))
```
cre8ivejp, line 28: When retrieving an int32 to a proto enum, we need to convert it as follows.

https://github.com/bucketeer-io/bucketeer/blob/7fbcbcb0c587c39671e03fed14fbfee4d6332e21/pkg/experiment/storage/v2/experiment.go#L152
https://github.com/bucketeer-io/bucketeer/blob/7fbcbcb0c587c39671e03fed14fbfee4d6332e21/pkg/experiment/storage/v2/experiment.go#L186

Don't we need to do the same when getting it here?
https://github.com/bucketeer-io/bucketeer/blob/7fbcbcb0c587c39671e03fed14fbfee4d6332e21/pkg/experiment/storage/v2/goal.go#L138

hvn2k1: I have converted it, same to get goal query!

cre8ivejp (approved):
Thank you!

PR #1498, opened 2025-02-03
feat: implement delete experiment on update rest

hvn2k1:


hvn2k1:
The delete option in console no longer necessary!

Issue #893, opened 2024-03-25
feat: be able to schedule the flag changes


cre8ivejp:
We support scheduling for enabling/disabling a flag, but we don't support scheduling for the changes made to the flag.
It would be nice to schedule a change to the variation value or a rule, etc. To do something similar, we need to change the default value and off variation and create schedule operations to switch on/off to change the value returned to the user.

- [x] Implement granular changes

- [x] Add `scheduled_flag_update` table and proto message

- [x] Implement scheduled flag update API

- [x] Implement conflict detection

- [x] Frontend

- [x] Implement cronjob to execute the updates

- [ ] Implement check for update conflicts

cre8ivejp:
I'm thinking of adding a new column to the feature table called schedule.
It will store an object, including the timestamp and the updated feature flag object.

When the cronjob runs (every minute), it checks the timestamp and updates the feature flag object with the scheduled one.
Regarding restrictions, when a change is scheduled, the user cannot make any other changes until the schedule is finished or if the user decides to delete the schedule.
This way is simple and doesn't require complex backend and front-end implementations.

cre8ivejp:
> I'm thinking of adding a new column to the feature table called schedule. It will store an object, including the timestamp and the updated feature flag object.
> 
> When the cronjob runs (every minute), it checks the timestamp and updates the feature flag object with the scheduled one. Regarding restrictions, when a change is scheduled, the user cannot make any other changes until the schedule is finished or if the user decides to delete the schedule. This way is simple and doesn't require complex backend and front-end implementations.

Because our users want to schedule multiple changes at different times, I'm changing the update flow to support it.

PR #1499, opened 2025-02-03
chore: add schedule update field to feature proto message

cre8ivejp:
Part of https://github.com/bucketeer-io/bucketeer/issues/893

The batch svc will check all the flags with scheduled updates.
If the `update_at` is older than the current time, it will replace the current feature object with the object in the `ScheduleUpdate` in the DB.


proto/feature/feature.proto
```diff
@@ -53,6 +53,7 @@ message Feature {
   bool archived = 20;
   repeated Prerequisite prerequisites = 21;
   string sampling_seed = 22;
+  repeated ScheduleUpdate schedule_update = 23;
```
cre8ivejp, line 56: Because the user might want to schedule multiple times, I'm using an array, so when implementing it on the frontend, we won't need to make changes on the backend.

hvn2k1 (approved):
lgtm!

Issue #1474, opened 2025-01-22
refactoring of transaction processing for storage


kakcy:
RunTransaction, Begin, etc. are called and transaction processing is performed at the API layer.

### Example of code that is an issue
- `RunTransaction`
    - https://github.com/bucketeer-io/bucketeer/blob/main/pkg/account/api/account.go#L64
- `Begin`
    - https://github.com/bucketeer-io/bucketeer/blob/main/pkg/autoops/api/api.go#L185

### Solution
- Define the necessary API for each storage layer using Interface and abstract the storage layer.
    - In the storage layer, embed the interface for each storage and move the implementation necessary for DB operations to the storage layer.
    - The API layer calls the above Interface and does not directly perform DB operations.
- If transaction processing is required in the API layer, use RunTransactionAPI provided in the DB client to perform transaction processing.
    - If we want your API layer to handle transactions across different storages, we can do so by creating storage instances using the same DB client instance.

### Package
- [x] account

- [x] autoops

- [x] environment

- [x] experiment

- [x] feature

- [x] notification

- [x] push

- [x] opsevent

PR #1501, opened 2025-02-04
refactor: changed transaction handling related to Environment package and Auth package

kakcy:
This pull request focuses on refactoring the `authService` and `EnvironmentService` structures to use storage interfaces instead of creating new instances within methods. This change improves code maintainability and testability. The most important changes include updating the `authService` and `EnvironmentService` structures, modifying various methods to use the new storage interfaces, and updating related tests.

Part of https://github.com/bucketeer-io/bucketeer/issues/1474

### DIFF
--- a/pkg/auth/api/api.go
+++ b/pkg/auth/api/api.go
@@ -122,6 +122,7 @@ func NewAuthService(
 		config:              config,
 		mysqlClient:         mysqlClient,
 		organizationStorage: envstotage.NewOrganizationStorage(mysqlClient),
+		environmentStorage:  envstotage.NewEnvironmentStorage(mysqlClient),
 		projectStorage:      envstotage.NewProjectStorage(mysqlClient),
 		accountClient:       accountClient,
 		verifier:            verifier,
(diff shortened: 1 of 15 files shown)


kakcy:
In the previous refactoring, the transaction instance was obtained from `Qe()`, but we have modified it so that if the transaction instance can be obtained from the Context received within each query execution function, the query will be executed based on it.

This modification has the following effects:
- When creating a Storage instance, the scope of storage is reduced by spawning the instance using a "QueryExecer" instance instead of "Client" as before the refactoring.
  - [In the previous refactoring code](https://github.com/bucketeer-io/bucketeer/blob/main/pkg/experiment/storage/v2/experiment.go#L73-L75)

- There is no need to go through `Qe()`, and the code can be simplified.
  -  [In the previous refactoring code](https://github.com/bucketeer-io/bucketeer/blob/main/pkg/experiment/storage/v2/experiment.go#L82)

hvn2k1 (approved):
the refactor LGTM!

nnnkkk7 (approved):
LGTM!

PR #1512, opened 2025-02-12
docs: mongodb and postgre implementation rfc

Ubisoft-potato:


cre8ivejp (approved):
I liked both options!
For now, let's implement PostgreSQL then if there is a demand ,we can also support Mongo DB.

PR #1522, opened 2025-02-17
fix: add order by options and correct experiment summary

hvn2k1:


pkg/experiment/storage/v2/sql/experiment/summarize_experiment.sql
```diff
@@ -0,0 +1,9 @@
+SELECT
+    COALESCE(SUM(IF(status = 0, 1, 0)), 0) AS waiting,
+    COALESCE(SUM(IF(status = 1, 1, 0)), 0) AS running,
+    COALESCE(SUM(IF(status = 2 OR status = 3, 1, 0)), 0) AS stopped
+FROM
+    experiment
+WHERE
+    environment_id = ?
+    AND deleted = false
```
cre8ivejp, line 9: Please add the following condition, too.

```suggestion
    AND deleted = false
    AND archived = false
```

hvn2k1: I've added the condition

cre8ivejp (approved):
Nice work!

PR #1523, opened 2025-02-18
chore: improve bucket hash algorithm using murmurHash3 instead of md5

cre8ivejp:
Because we bucket millions of users, the MD5 might not be the best solution for our case. Currently, our implementation 
 uses MD5, and it returns a 128-bit hash, but we only use the first 8 bytes (64-bit) hash.

I have searched and saw that many companies use the MurmurHash3 algorithm to determine the variation for the user. While MurmurHash3 is slightly faster, the goal is good randomness and avoiding collisions when bucketing millions of users.

Also, since we are just bucketing the user without storing the hash, we don't need cryptography when hashing, so I'm choosing MurmurHash3 over the sha-256 because it is much faster, and 128-bit is enough to handle millions of users.

**Note: This change might change the variation result compared to the MD5, which means it could affect running experiments. So, I will try to merge this when no experiments are running.**

## Comparison of Hashing Algorithms for Bucketing

| Feature               | **MD5**                            | **SHA-256**                          | **MurmurHash3**                        |
|-----------------------|------------------------------------|--------------------------------------|--------------------------------------|
| **Security**           | ❌ Weak (vulnerable to collisions) | ✅ Strong (Collision resistant)      | ✅ Strong for non-cryptographic uses |
| **Output Size**        | 128-bit (16 bytes)                | 256-bit (32 bytes)                  | 128-bit (16 bytes)                   |
| **Performance**        | ⚡ Fast (~2x-3x faster than SHA-256) | ⏳ Slower (about 2x slower than MD5) | ⚡ Very fast (optimized for speed)   |
| **Randomness**         | ❌ Less uniform (more clustering)  | ✅ More uniform (better distribution) | ✅ Very good randomness, low bias   |
| **Collision Resistance**| ❌ Weak                           | ✅ Strong                           | ✅ Very low risk of collisions      |
| **Use Cases**          | ✅ Fast lookups, non-secure cases  | ✅ Cryptographic and secure use cases| ✅ General-purpose hashing, fast distribution |
| **Implementation Complexity** | 🔧 Simple                    | 🔧 More complex                     | 🔧 Simple                           |
| **Commonly Used In**   | Legacy systems, checksum          | Secure applications, signatures     | Non-cryptographic systems (databases, hashing for features) |
| **Hashing Library Support** | ✅ Wide support               | ✅ Widely supported                 | ✅ Supported by many libraries      |

## MurmurHash3 Comparison Between Node.js and Go SDKs

| SDK         | High                    | Low                     | Normalized Value         |
|-------------|-------------------------|-------------------------|--------------------------|
| Node.js     | 16291349615033665716    | 10595109440957946947    | 0.8831558322670193       |
| Go          | 16291349615033665716    | 10595109440957946947    | 0.8831558322670193       |

**Note: We must also update the Go and Nodejs server SDKs to maintain the bucket algorithm consistency.**

### Things done

- Changed to bucket the user using MurmurHash3 instead of MD5 for Go and Nodejs
- Added unit tests to ensure that Go and Nodejs have the same hash output

---

This pull request introduces significant changes to the `evaluation` package, including the integration of MurmurHash3 for hashing, modifications to the bucketing logic, and updates to related tests. The changes span both the Go and TypeScript implementations, ensuring consistency across different parts of the codebase.

#### Go Implementation:
* **Bucketing Logic and MurmurHash3 Integration:**
  - Added a new `bucketeer` struct with methods to compute MurmurHash3 (128-bit) and convert it to a normalized float64 value. (`evaluation/go/bucketeer.go`)
  - Updated `strategyEvaluator` to use the new `bucketeer` for bucketing logic, replacing the previous MD5-based implementation. (`evaluation/go/strategy_evaluator.go`) [[1]](diffhunk://#diff-2fe5bbfd5b860b8e07f79d0f740d66e043c99d9b9276c8eebbb243a2665c5d64L18-L27) [[2]](diffhunk://#diff-2fe5bbfd5b860b8e07f79d0f740d66e043c99d9b9276c8eebbb243a2665c5d64L42-R37) [[3]](diffhunk://#diff-2fe5bbfd5b860b8e07f79d0f740d66e043c99d9b9276c8eebbb243a2665c5d64L53-L85)

* **Testing:**
  - Added tests for the new `bucketeer` methods to ensure correct hash computation and normalization. (`evaluation/go/bucketeer_test.go`)
  - Modified existing tests to reflect changes in variation assignment logic. (`evaluation/go/evaluation_test.go`)

#### TypeScript Implementation:
* **Bucketing Logic and MurmurHash3 Integration:**
  - Introduced a `Bucketeer` class with methods to compute MurmurHash3 (128-bit) and convert it to a normalized float64 value. (`evaluation/typescript/src/bucketeer.ts`)
  - Updated `StrategyEvaluator` to use the new `Bucketeer` for bucketing logic, replacing the previous MD5-based implementation. (`evaluation/typescript/src/strategyEvaluator.ts`) [[1]](diffhunk://#diff-c53761187730a367ad5d9d06a3b234960423edb9127b6771c1d3285c515f3878L1-L5) [[2]](diffhunk://#diff-c53761187730a367ad5d9d06a3b234960423edb9127b6771c1d3285c515f3878L36-L65)

* **Testing:**
  - Added tests for the new `Bucketeer` methods to ensure correct hash computation and normalization. (`evaluation/typescript/src/__tests__/evaluator/bucketeer.ts`)
  - Modified existing tests to reflect changes in variation assignment logic. (`evaluation/typescript/src/__tests__/evaluator/assign_user_sampling_seed.ts`)

#### Dependencies:
* **Go:**
  - Added `github.com/spaolacci/murmur3` to the `go.mod` file. (`go.mod`)

* **TypeScript:**
  - Added `murmurhash3js` and its type definitions to the `package.json` file. (`evaluation/typescript/package.json`) [[1]](diffhunk://#diff-4e552c36cddb498024f82e0b6a55d75319b249ba56b576a447ad0541aba84e8fR13-L22) [[2]](diffhunk://#diff-4e552c36cddb498024f82e0b6a55d75319b249ba56b576a447ad0541aba84e8fR28-R32)


evaluation/typescript/src/bucketeer.ts
```diff
@@ -0,0 +1,39 @@
+import murmurHash3 from 'murmurhash3js';
+
+export default class Bucketeer {
```
duyhungtnn, line 3:
```suggestion
class Bucketeer {
```

evaluation/typescript/src/__tests__/evaluator/bucketeer.ts
```diff
@@ -0,0 +1,24 @@
+import test from 'ava';
+import Bucketeer from '../../bucketeer';
```
duyhungtnn, line 2:
```suggestion
import { Bucketeer } from '../../bucketeer';
```

cre8ivejp: @duyhungtnn, thanks. I have fixed them.

hvn2k1 (approved):
Nice work!

PR #1508, opened 2025-02-06
refactor: changed transaction handling related to Account package

kakcy:
This pull request includes significant changes to the `pkg/account/api/account.go` file, focusing on updating transaction handling and integrating audit log creation. Additionally, it includes corresponding updates to the test file `pkg/account/api/account_test.go`.

Part of https://github.com/bucketeer-io/bucketeer/issues/1474
Part of #1002

Issue #1002, opened 2024-06-10
Remove audditlog-persister


kentakozuka:
### Describe the feature you'd like

In the case that publishing domain events fails after the transaction of DB success, the object data is not compatible with audit log page.
The solution is to remove the audit log-persister and to include saving audit log data to transactions.

PR #1526, opened 2025-02-20
feat: implement progressive rollout REST api

hvn2k1:

Implement REST and remove command from create/list/get/stop/delete


pkg/autoops/api/progressive_rollout.go
```diff
@@ -850,9 +1010,13 @@ func (s *AutoOpsService) getFeature(
 	req *autoopsproto.CreateProgressiveRolloutRequest,
 	localizer locale.Localizer,
 ) (*featureproto.Feature, error) {
+	featureID := req.FeatureId
+	if req.Command != nil {
+		featureID = req.Command.FeatureId
+	}
```
cre8ivejp, line 1016: Just confirming.
Is there an issue that we need to check the featureID in the command, too?

cre8ivejp:
@hvn2k1, please fix the conflict so we can merge it.

pkg/autoops/api/progressive_rollout.go
```diff
@@ -83,8 +104,83 @@ func (s *AutoOpsService) CreateProgressiveRollout(
...
+			return nil, statusProgressiveRolloutInternal.Err()
+		}
+		return nil, dt.Err()
+	}
+	return &autoopsproto.CreateProgressiveRolloutResponse{
+		ProgressiveRollout: progressiveRollout.ProgressiveRollout,
+	}, nil
+}
+
+func (s *AutoOpsService) CreateProgressiveRolloutNoCommand(
```
kakcy, line 138: The function name starts with a capital letter, is that intentional?

Similar functions in other packages are in lower case, so the scope of disclosure is likely to be different.

pkg/autoops/api/progressive_rollout.go
```diff
@@ -83,8 +104,83 @@ func (s *AutoOpsService) CreateProgressiveRollout(
...
+		req.FeatureId,
+		req.ProgressiveRolloutManualScheduleClause,
+		req.ProgressiveRolloutTemplateScheduleClause,
+	)
+	if err != nil {
+		return nil, err
+	}
+	err = s.mysqlClient.RunInTransactionV2(ctx, func(contextWithTx context.Context, tx mysql.Transaction) error {
+		storage := v2as.NewProgressiveRolloutStorage(s.mysqlClient)
+		event, err := domainevent.NewEvent(
```
kakcy, line 157: @cre8ivejp 
In order to respond to [this issue](https://github.com/bucketeer-io/bucketeer/issues/1002), I understand that the `AuditLog` will be updated within the same transaction from now on, but is this correct?

I implemented this support in the `Account` package in [that PR](https://github.com/bucketeer-io/bucketeer/pull/1508).
※ [This](https://github.com/bucketeer-io/bucketeer/blob/main/pkg/account/api/account.go#L232) is the actual corresponding code.


If the above idea is correct, I think it would be more efficient to convert to `NoCommand` at the same time.What do you think?
When converting to `NoCommand`, a domain event is always generated, so you can use it to update the AuditLog storage.

pkg/autoops/api/progressive_rollout.go
hvn2k1: No it should be lowercase, thank you for pointing that out!

hvn2k1:
> @hvn2k1, please fix the conflict so we can merge it.

I've resolved conflict

pkg/autoops/api/progressive_rollout.go
kakcy: @cre8ivejp 
That's Looks good!

cre8ivejp (approved):
Thanks 👍

cre8ivejp (approved):
👍

PR #1529, opened 2025-02-24
feat(ui/dashboard): remove command object all payload

steveninhle:
### Things done:

- Removed `command` object in all APIs payload
- Changed field name with new APIs
- Updated type name for consistency


ui/dashboard/src/@api/account/account-updater.ts
```diff
@@ -2,46 +2,29 @@ import axiosClient from '@api/axios-client';
...
 
-type WriteType =
-  | 'WriteType_UNSPECIFIED'
-  | 'WriteType_OVERRIDE'
-  | 'WriteType_PATCH';
-
+export type AvatarCommand = {
+  avatarImage: string;
+  avatarFileType: string;
+};
```
cre8ivejp, line 8: We are removing the commands, but here, you use a command.
Shouldn't this be `Avatar`, instead?

steveninhle: I see that this type isn’t used anymore. I'll remove it. Thanks!

cre8ivejp (approved):
Thank you!

PR #1532, opened 2025-02-25
fix: evaluation unique count not working for go sdk

cre8ivejp:
Fix https://github.com/bucketeer-io/bucketeer/issues/1533

Because the `userId` field in the EvaluationEvent proto message is already in the `User` field, we should remove it to avoid sending the same value twice.
To keep compatibility, we must check both fields until all the SDKs are updated.


hvn2k1 (approved):
LGTM!

Ubisoft-potato (commented):
LGTM!

PR #1537, opened 2025-02-27
fix: cannot delete all tags in the account update api

cre8ivejp:
Fix https://github.com/bucketeer-io/bucketeer/issues/1538

The user can't delete all flags when sending an empty list.


hvn2k1 (approved):
Thank you!

PR #1544, opened 2025-02-28
feat: implement update organization no command

hvn2k1:


proto/environment/service.proto
```diff
@@ -286,12 +286,32 @@ message CreateOrganizationResponse {
...
+    }
+  ];
+  google.protobuf.StringValue name = 5;
+  google.protobuf.StringValue description = 6;
+  google.protobuf.StringValue owner_email = 7;
 }
 
-message UpdateOrganizationResponse {}
+message UpdateOrganizationResponse {
+  Organization organization = 1;
```
cre8ivejp, line 313: Currently, we only return the object when creating it.
Did you change this in other update APIs, too?

pkg/environment/api/organization.go
```diff
@@ -755,6 +759,138 @@ func (s *EnvironmentService) UpdateOrganization(
...
+	}
+
+	if err = s.publisher.Publish(ctx, event); err != nil {
+		s.logger.Error(
+			"Failed to publish the event",
+			log.FieldsFromImcomingContext(ctx).AddFields(
+				zap.Error(err),
+			)...,
+		)
+		return nil, statusInternal.Err()
```
cre8ivejp, line 825: We must localize the error for internal errors, too.
https://github.com/bucketeer-io/bucketeer/pull/1544/files#diff-ec188e5743280db370beaa0e9d0ced0b2988402e2486be35e25dedc1fb75d653R884-R887

proto/environment/service.proto
hvn2k1: Yes!

cre8ivejp: Ok. I missed that.
We shouldn't be doing this because it will introduce inconsistency between other APIs.
But if there is a use for this, we can consider it.

hvn2k1: I've update the response in all API environment service

pkg/environment/api/organization.go
hvn2k1: I missed that! updated

cre8ivejp (approved):
Thank you!

PR #1549, opened 2025-03-03
refactoring: changed transaction handling related to Segment in Feature package.

kakcy:
This pull request includes significant changes to the `FeatureService` class in the `pkg/feature/api` package, focusing on integrating `SegmentStorage` and refactoring transaction management. The main changes involve replacing direct transaction management with a new transactional method and updating the tests accordingly.

I reviewed the refactoring method I used up until now.
- Modified to switch the query execution instance within the function that executes the query.
- `Client.Qe()` is deprecated.

Therefore, Storage can be generated in the `QueryExecer` scope instead of the Client, so It is now possible to reduce the scope of a layer.

### DIFF
--- a/pkg/feature/api/api.go
+++ b/pkg/feature/api/api.go
@@ -58,6 +58,7 @@ func WithLogger(l *zap.Logger) Option {
 type FeatureService struct {
 	flagTriggerStorage    v2fs.FlagTriggerStorage
 	featureStorage        v2fs.FeatureStorage
+	segmentStorage        v2fs.SegmentStorage
 	mysqlClient           mysql.Client
 	accountClient         accountclient.Client
 	experimentClient      experimentclient.Client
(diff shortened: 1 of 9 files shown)


kakcy:
@cre8ivejp 
In the previous refactoring, the transaction instance was obtained from `Qe()`, but in this PR, the transaction instance was obtained from the Context received within each query execution function, and the query was executed based on that.

This modification has the following effects:

- When creating a Storage instance, the scope of storage is reduced by spawning the instance using a `QueryExecer` instance instead of `Client` as before the refactoring.

[In the previous refactoring code](https://github.com/bucketeer-io/bucketeer/blob/main/pkg/experiment/storage/v2/experiment.go#L73-L75)
- There is no need to go through `Qe()`, and the code can be simplified.

[In the previous refactoring code](https://github.com/bucketeer-io/bucketeer/blob/main/pkg/experiment/storage/v2/experiment.go#L82)

Ubisoft-potato (commented):
Nice refactor!

hvn2k1 (approved):
The changes LGTM!

PR #1550, opened 2025-03-03
fix: flag not found when using dependency flag in targeting rule

cre8ivejp:
Fix https://github.com/bucketeer-io/bucketeer/issues/1552


hvn2k1 (approved):
👍🏻

PR #1551, opened 2025-03-03
test: fix e2e environment should call get environment API again

hvn2k1:
CreateEnvironment no longer return environment, should call get environment to verify again


cre8ivejp (approved):
👍

PR #1581, opened 2025-03-12
fix: remove `environment_id` validation from code references api

cre8ivejp:
Because we have an environment in production that has an empty `environment_id`, we can't validate it. Otherwise, all the code references API will fail for that specific environment


Ubisoft-potato (approved):
LGTM!👍

hvn2k1 (approved):
Thank you!

PR #1585, opened 2025-03-13
fix: update refresh token TTL to 7 days

### LINKED ISSUE #1579: chore: investigate why the console is getting a 1-hour refresh token instead of 1 month

nnnkkk7:
fix #1579

- Changed default refresh token TTL  from 1 hour to 7 days
- Added explicit WithRefreshTokenTTL() in service initialization

In the `generateToken` function, there was logic to set a default 30-day TTL but override it with the options value if present:

```go
refreshTokenTTL := 30 * day
if s.opts.refreshTokenTTL > 0 {
    refreshTokenTTL = s.opts.refreshTokenTTL
}
```

However, the `defaultOptions` had the refresh token TTL set to just 1 hour:
```go
var defaultOptions = options{
    refreshTokenTTL: time.Hour,
    logger:          zap.NewNop(),
}
```

And the service initialization did not explicitly specify the WithRefreshTokenTTL option, so the default 1-hour value was used.

The condition `if s.opts.refreshTokenTTL > 0` was always true, causing the default 30-day setting to always be overridden with the 1-hour value.

### DIFF
--- a/pkg/web/cmd/server/server.go
+++ b/pkg/web/cmd/server/server.go
@@ -126,6 +126,7 @@ type server struct {
 	autoOpsService       *string
 	codeReferenceService *string
 	// auth
+	refreshTokenTTL     *time.Duration
 	emailFilter         *string
 	oauthConfigPath     *string
 	oauthPublicKeyPath  *string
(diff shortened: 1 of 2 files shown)


pkg/web/cmd/server/server.go
```diff
@@ -825,6 +825,7 @@ func (s *server) createAuthService(
 	}
 	serviceOptions := []authapi.Option{
 		authapi.WithLogger(logger),
+		authapi.WithRefreshTokenTTL(30 * 24 * time.Hour),
```
cre8ivejp, line 828: Instead of passing a fixed value here, how about implementing it as a config value being passed using flags?
This way, the user can set what is best for him when hosting the project.

E.g.

```go
refreshTokenTTL: cmd.Flag("refresh-token-ttl", "refresh token ttl").Default("168h").Duration() // 7 days
```

nnnkkk7: I see. Thank you!
Is it ok ? [feat: add refresh token TTL config](https://github.com/bucketeer-io/bucketeer/pull/1585/commits/0eec63d6a347e8610413ca6635330267826b80f4)

cre8ivejp (approved):
Thank you!

PR #1589, opened 2025-03-14
build(deps): upgrade Go version to 1.24

nnnkkk7:
It would be better to update Go version to update the dependent packages.


.github/workflows/push-image.yaml
```diff
@@ -36,7 +36,7 @@ on:
 env:
   GHCR_REGISTRY: ghcr.io/bucketeer-io
   GAR_REGISTRY: asia-docker.pkg.dev
-  GO_VERSION: 1.21.10
+  GO_VERSION: 1.24.1
```
cre8ivejp, line 39: Instead of setting the go version in all CI workflows, we can use `go-version-file: 'path/to/go.mod'` option.
This way, we will need to maintain only the go.mod file and ensure that all other places will use the same version.
WDYT?

Reference.
https://github.com/actions/setup-go?tab=readme-ov-file#getting-go-version-from-the-gomod-file

nnnkkk7: Thank you!
I agree with it!
I fixed it. [chore: remove hardcoded Go version and use go.mod with go-version-file](https://github.com/bucketeer-io/bucketeer/pull/1589/commits/c588cbb643331aa65899970d214f50ee1475b252)

cre8ivejp (approved):
Nice!
Thanks.

PR #1592, opened 2025-03-17
refactor: change the Account storage scope to QueryExecer

kakcy:
This pull request includes changes to the `pkg/account/storage/v2` package, focusing on refactoring the database query execution to use a simpler and more direct approach. The changes also include updates to the corresponding tests to align with the new approach.

This PR is a refactor to the deprecation of `Qe()` in the PR below.
#1549

### DIFF
--- a/pkg/account/storage/v2/admin_account_test.go
+++ b/pkg/account/storage/v2/admin_account_test.go
@@ -14,74 +14,74 @@
 
 package v2
 
-// func TestGetAdminAccountV2(t *testing.T) {
-// 	t.Parallel()
-// 	mockController := gomock.NewController(t)
-// 	defer mockController.Finish()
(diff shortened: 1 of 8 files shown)


pkg/account/storage/v2/admin_account_test.go
```diff
@@ -14,86 +14,74 @@
 
 package v2
 
-import (
-	"context"
-	"errors"
-	"testing"
+// func TestGetAdminAccountV2(t *testing.T) {
```
cre8ivejp, line 17: Don't we need this test?

kakcy: @cre8ivejp 
Sorry, there was an unintended commit mixed in, I fixed it.

PR #1597, opened 2025-03-18
ci: pin versions of actions to commit hashes

nnnkkk7:
Updates all GitHub Actions references from version tags to specific commit hashes.

- Replaced all version tags with exact commit SHAs.
- Added version comments for future reference.

ref: [Harden-Runner detection: tj-actions/changed-files action is compromised - Recovery Steps](https://www.stepsecurity.io/blog/harden-runner-detection-tj-actions-changed-files-action-is-compromised#recovery-steps)


cre8ivejp (approved):
Nice!

PR #1606, opened 2025-03-19
fix: ops count is unexpectedly greater than evaluation count

cre8ivejp:
The evaluation and goal events are sent from the client SDK, and for many reasons, those events could get lost in the client.
So, I implemented clamping the computed rate so that it never exceeds 100% (i.e., 1.0). In other words, if you detect that the opsCount is greater than the evaluationCount, it will adjust the opsCount to be equal to the evaluationCount before calculating the rate. This way, you avoid a rate above 1.0, and the downstream logic stays consistent.

Fix https://github.com/bucketeer-io/bucketeer/issues/1607


hvn2k1 (dismissed):
thank you <3

PR #1620, opened 2025-03-26
feat: support semver with v prefix when evaluate values

### LINKED ISSUE #1248: feat: support `v` prefix when validating semver
Currently, the [module](https://github.com/blang/semver) being used in Bucketeer doesn't support `v` prefix.
Some platforms do and do not support the `v` prefix, which creates an inconsistency.

The semver value is used in the Flag targeting as a custom rule. Since the flag can be used in any supported SDKs (NodeJS, Android, etc), we should also support the `v` prefix.

It seems Go is trying to make it a default standard requiring the `v` prefix.
`
This package follows Semantic Versioning 2.0.0 (see semver.org) with two exceptions. First, it requires the "v" prefix
`
Reference:
https://pkg.go.dev/golang.org/x/mod/semver

nnnkkk7:
fix #1248 

Updated the `parseSemver` function to maintain consistency when comparing semantic versions by:

1. Detecting if the target value has a 'v' prefix
2. When the target has a 'v' prefix, only comparing it with values that also have a 'v' prefix
3. When the target doesn't have a 'v' prefix, only comparing it with values without a 'v' prefix

This approach ensures that version comparisons are done consistently within the same format, aligning with the broader Go ecosystem's practice of supporting the 'v' prefix in semantic versioning (as noted in golang.org/x/mod/semver).

### DIFF
--- a/evaluation/go/clause_evaluator.go
+++ b/evaluation/go/clause_evaluator.go
@@ -276,12 +276,27 @@ func (c *clauseEvaluator) parseFloat(targetValue string, values []string) (float
 }
 
 func (c *clauseEvaluator) parseSemver(targetValue string, values []string) (semver.Version, []semver.Version, error) {
-	versionTarget, err := semver.Parse(targetValue)
+	targetValueWithoutV, targetHasV := strings.CutPrefix(targetValue, "v")
+	versionTarget, err := semver.Parse(targetValueWithoutV)
 	if err != nil {
 		return semver.Version{}, nil, err
 	}
 	versionValues := make([]semver.Version, 0, len(values))
+	if targetHasV {
+		for _, value := range values {
+			// if target value has v prefix, compare only values with v prefix
+			valueWithoutV, valueHasV := strings.CutPrefix(value, "v")
+			if valueHasV {
+				v, err := semver.Parse(valueWithoutV)
+				if err == nil {
+					versionValues = append(versionValues, v)
+				}
+			}
+		}
+		return versionTarget, versionValues, nil
+	}
 	for _, value := range values {
+		// if target value has no v prefix, compare only values without v prefix
 		v, err := semver.Parse(value)
 		if err == nil {
 			versionValues = append(versionValues, v)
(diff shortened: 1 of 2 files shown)


evaluation/go/clause_evaluator.go
```diff
@@ -276,12 +276,27 @@ func (c *clauseEvaluator) parseFloat(targetValue string, values []string) (float
 }
 
 func (c *clauseEvaluator) parseSemver(targetValue string, values []string) (semver.Version, []semver.Version, error) {
-	versionTarget, err := semver.Parse(targetValue)
+	targetValueWithoutV, targetHasV := strings.CutPrefix(targetValue, "v")
+	versionTarget, err := semver.Parse(targetValueWithoutV)
```
cre8ivejp, line 280: If you cut the prefix and check if it is a semver, I think the following logic is unnecessary since we use it only for comparison. We can cut the prefix and do it as it is.

If the version is `v1.0.0`, it returns as `1.0.0`. WDYT?

https://github.com/bucketeer-io/bucketeer/pull/1620/files#diff-d4ec5591580e1c94e9ec4a13a487728208bc9e7ff389bc388dff3023b498a2aeR285-R297

nnnkkk7: Thank you for the suggestion!!

I'd like to clarify the background and intention of the current implementation.

The original design deliberately compares only versions with matching notation formats:
- When the target has a 'v' prefix, we only compare with values that also have the 'v' prefix
- When the target has no 'v' prefix, we only compare with values without the 'v' prefix

This was intentional to maintain consistency in version notation and prevent confusion. Users expect that when they specify a format (with or without 'v'), the comparison should respect that format choice.

If we simplify the code to strip all 'v' prefixes and mix both formats in comparisons (treating "v1.0.0" and "1.0.0" as equivalent), it would indeed make the code cleaner. However, this could represents a change in behavior that could confuse users who rely on the current format-specific comparison🤔

WDYT?

cre8ivejp: I agree!
Thank you!

cre8ivejp (approved):
Thank you 🎉

PR #1622, opened 2025-03-26
chore: replace deprecated ioutil

nnnkkk7:
`ioutil` pakcage is deprecated. 
We should use package io or package os instead.
https://pkg.go.dev/io/ioutil


kakcy (approved):
Thank you! LGTM 👍

PR #1624, opened 2025-03-27
chore: validate events from the sdk in api gateway

### LINKED ISSUE #1614: chore: validate events from the sdk
When running experiments, the evaluation and goal events are saved into BigQuery. All the fields saved in the BigQuery table are required, but we don't check before saving them. This could accumulate unnecessary messages in the PubSub for 7 days until they are purged. Also, because we append the events in batches in the BigQuery, if one event fails, all the others will fail, too, which could affect other experiments.

nnnkkk7:
fix #1614 

- This PR adds validation for required fields in EvaluationEvent and GoalEvent to ensure all essential data is present before processing.

- When running experiments, evaluation and goal events are saved to BigQuery. Older SDK versions may send events with missing required fields, causing: #1614

### DIFF
--- a/pkg/api/api/api_grpc_test.go
+++ b/pkg/api/api/api_grpc_test.go
@@ -3750,11 +3750,28 @@ func TestGrcpRegisterEvents(t *testing.T) {
 	mockController := gomock.NewController(t)
 	defer mockController.Finish()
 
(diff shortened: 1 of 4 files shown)


cre8ivejp (approved):
Nice work!

PR #1637, opened 2025-03-28
revert: validate events from the sdk in api gateway

nnnkkk7:
Reverts bucketeer-io/bucketeer#1624

Some e2e tests fail becase of this change..🙇


kakcy (approved):
👍

PR #1639, opened 2025-03-28
feat: implement execute progressive rollout no command

hvn2k1:


proto/autoops/service.proto
```diff
@@ -784,5 +790,49 @@ service AutoOpsService {
...
   }
   rpc ExecuteProgressiveRollout(ExecuteProgressiveRolloutRequest)
-      returns (ExecuteProgressiveRolloutResponse) {}
+      returns (ExecuteProgressiveRolloutResponse) {
+    option (google.api.http) = {
+      post: "/v1/progressive_rollout/execute"
+      body: "*"
+    };
+    option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
+      summary: "Execute"
```
cre8ivejp, line 799:
```suggestion
      summary: "Execute Progressive Rollout"
```

pkg/autoops/api/progressive_rollout.go
```diff
@@ -733,6 +738,199 @@ func (s *AutoOpsService) ExecuteProgressiveRollout(
...
+		)
+		return nil, err
+	}
+	var events []publisher.Message
+	err = s.mysqlClient.RunInTransactionV2(ctx, func(contextWithTx context.Context, tx mysql.Transaction) error {
+		progressiveRollout, err := s.prStorage.GetProgressiveRollout(contextWithTx, req.Id, req.EnvironmentId)
+		if err != nil {
+			return err
+		}
+		ftStorage := ftstorage.NewFeatureStorage(tx)
```
kakcy, line 764: @hvn2k1 
At this point, a FeatureStorage instance is generated, but by providing an AutoOpsService.featureStorage property to the AutoOpsService class, you can mock GetFeature() in the test, which I think will enable more effective testing. What do you think?
(The current test only tests whether QueryRowContext() is called, so I think it's a bit weak.)

hvn2k1: I was doing this dependency injection in other domain too, but missed that in this pr. Thanks for pointing out!

proto/autoops/service.proto
hvn2k1: I've updated the description

cre8ivejp (approved):
LGTM 👍

kakcy (approved):
Thank you 👍

Issue #1632, opened 2025-03-27
fix(ui/dashboard): can't update a slack integration


cre8ivejp:
When clicking on the name or on the button in the action menu, nothing happens.

PR #1641, opened 2025-03-28
fix(ui/dashboard): can't update a slack integration

steveninhle:
When clicking on the name or on the button in the action menu, nothing happens.

Fix: #1632


cre8ivejp (approved):
Thank you!

PR #1644, opened 2025-03-28
fix: ambiguous column in get features

hvn2k1:
Fix ambiguous column in whereparts
```
"message":"Failed to get feature","serviceContext":{"service":"bucketeer-web.server","version":"-"}
"error":"Error 1052 (23000): Column 'id' in where clause is ambiguous"
```


nnnkkk7 (approved):
Thank you!

PR #1650, opened 2025-03-28
chore: validate events from the sdk in api gateway

nnnkkk7:
- Reverts bucketeer-io/bucketeer#1637
- Validate if userID is empty or userID in User Object is empty.(We want to use User Object, but userID is still used in old sdk)
- Unify userID retrieval in subscriver because some processers use userID not in User Object.

### DIFF
--- a/pkg/api/api/grpc_validation.go
+++ b/pkg/api/api/grpc_validation.go
@@ -122,7 +122,7 @@ func (v *eventGoalValidator) validate(ctx context.Context) (string, error) {
 		)
 		return codeEmptyField, errEmptyGoalID
 	}
-	if ev.User.Id == "" && ev.UserId == "" {
+	if (ev.User == nil || (ev.User != nil && ev.User.Id == "")) && ev.UserId == "" {
 		v.logger.Debug(
 			"Empty user_id",
 			log.FieldsFromImcomingContext(ctx).AddFields(
@@ -199,7 +199,7 @@ func (v *eventEvaluationValidator) validate(ctx context.Context) (string, error)
 		)
 		return codeEmptyField, errEmptyVariationID
 	}
-	if ev.User.Id == "" && ev.UserId == "" {
+	if (ev.User == nil || (ev.User != nil && ev.User.Id == "")) && ev.UserId == "" {
 		v.logger.Debug(
 			"Empty user_id",
 			log.FieldsFromImcomingContext(ctx).AddFields(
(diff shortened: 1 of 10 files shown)


pkg/api/api/grpc_validation.go
```diff
@@ -105,6 +110,30 @@ func (v *eventGoalValidator) validate(ctx context.Context) (string, error) {
...
+		v.logger.Debug(
+			"Empty goal_id",
+			log.FieldsFromImcomingContext(ctx).AddFields(
+				zap.String("id", v.event.Id),
+				zap.String("goal_id", ev.GoalId),
+			)...,
+		)
+		return codeEmptyField, errEmptyGoalID
+	}
+	if ev.User.Id == "" && ev.UserId == "" {
```
cre8ivejp, line 125: The `ev.User` could be nil, so we must check it.

pkg/api/api/grpc_validation.go
```diff
@@ -148,6 +177,50 @@ func (v *eventEvaluationValidator) validate(ctx context.Context) (string, error)
...
+		v.logger.Debug(
+			"Empty variation_id",
+			log.FieldsFromImcomingContext(ctx).AddFields(
+				zap.String("id", v.event.Id),
+				zap.String("variation_id", ev.VariationId),
+			)...,
+		)
+		return codeEmptyField, errEmptyVariationID
+	}
+	if ev.User.Id == "" && ev.UserId == "" {
```
cre8ivejp, line 202: Same here.

pkg/api/api/grpc_validation.go
nnnkkk7: Thank you!
I fixed it!
[fix: add nil check for User object](https://github.com/bucketeer-io/bucketeer/pull/1650/commits/e06e1f1b6d99c9de003c2da605050d0abc61d0db)

cre8ivejp: Thank you!
Just to be safe, can you add a unit test for nil case?

nnnkkk7: ok!
I added it!
[fix: add test case for nil check](https://github.com/bucketeer-io/bucketeer/pull/1650/commits/455db9ad6ba00586c2980efae0f19e291e7b850f)

cre8ivejp (approved):
Nice!
Thank you 🎉

PR #1653, opened 2025-03-28
fix(ui/dashboard): can't update api key

### LINKED ISSUE #1628: fix(ui/dashboard): can't update api key
The list page shows which environment the key belongs to, but on the update page, the environment isn't shown in the drop-down menu, which I think is the issue.

steveninhle:
The list page shows which environment the key belongs to, but on the update page, the environment isn't shown in the drop-down menu, which I think is the issue.

Fix #1628


cre8ivejp (approved):
Thank you!

PR #1654, opened 2025-03-29
fix(ui/dashboard): disabled add environment btn when it reaches the max number of environments

### LINKED ISSUE #1629: fix(ui/dashboard): add environment button is adding more than the number of environments
The add environment button can be clicked infinitive on the member's create/update page. It should disable the button when it reaches the max 
 number of environments in the organization.

steveninhle:
The add environment button can be clicked infinitive on the member's create/update page. It should disable the button when it reaches the max number of environments in the organization.

Fix #1629


cre8ivejp (approved):
Thank you!

PR #1655, opened 2025-03-29
fix(ui/dashboard): date column is not showing the full date when passing the mouse over

### LINKED ISSUE #1651: fix(ui/dashboard): date column is not showing the full date when passing the mouse over
Pages affected (List page)

- Slack
- Members
- User Segments

Need to check other pages.

steveninhle:
Pages affected (List page)

Slack
Members
User Segments
Need to check other pages.

Fix #1651


cre8ivejp (approved):
Thanks!

PR #1656, opened 2025-03-30
fix(ui/dashboard): submit button is not being validated on the filter on the project list page

### LINKED ISSUE #1645: fix(ui/dashboard): submit button is not being validated on the filter on the project list page
The button should only be enabled after the user selects the filter value.
Check all the other list pages.

steveninhle:
The button should only be enabled after the user selects the filter value.
Check all the other list pages.

Fix #1645


cre8ivejp (approved):
Thank you!

PR #1657, opened 2025-03-30
fix(ui/dashboard): the clear button is different from the design

### LINKED ISSUE #1646: fix(ui/dashboard): the clear button is different from the design
Some pages, such as the Member list page, are, showing the background color button incorrectly. It should be white with a purple border.
Check all other pages.

steveninhle:
Some pages, such as the Member list page, are, showing the background color button incorrectly. It should be white with a purple border.
Check all other pages.

Fix #1646


cre8ivejp (approved):
👍

PR #1659, opened 2025-03-30
fix(ui/dashboard): variation value validation when creating a new flag

### LINKED ISSUE #1648: fix(ui/dashboard): variation value validation when creating a new flag
The validation fails in the variation value when changing the flag type from String to Boolean.

steveninhle:
The validation fails in the variation value when changing the flag type from String to Boolean.

Fix #1648


cre8ivejp (approved):
Thanks!

PR #1660, opened 2025-03-30
fix(ui/dashboard): admin console always selects the first environment in the list

### LINKED ISSUE #1647: fix(ui/dashboard): admin console always selects the first environment in the list
When accessing the console after a while, it always selects the first environment from the list.
The local storage should have the last selected environment ID after the user logs in. If the selected environment ID is not present in the list, it should force the user to re-login.

steveninhle:
When accessing the console after a while, it always selects the first environment from the list.
The local storage should have the last selected environment ID after the user logs in. If the selected environment ID is not present in the list, it should force the user to re-login.

Fix #1647

### DIFF
--- a/ui/dashboard/src/auth/utils.ts
+++ b/ui/dashboard/src/auth/utils.ts
@@ -11,11 +11,11 @@ export const currentEnvironmentRole = (
     ? currentEnvId
     : account.environmentRoles[0].environment.id;
 
-  let curEnvRole = account.environmentRoles.find(
-    environmentRole =>
-      environmentRole.environment.id === curEnvId ||
-      environmentRole.environment.urlCode === curEnvId
-  );
+  let curEnvRole = account.environmentRoles.find(environmentRole => {
+    const { environment } = environmentRole;
+    if (environment.id && curEnvId) return environment.id === curEnvId;
+    return environment.urlCode === curEnvId;
+  });
   if (!curEnvRole) {
     curEnvRole = account.environmentRoles[0];
   }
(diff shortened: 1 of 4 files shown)


ui/dashboard/src/auth/utils.ts
```diff
@@ -11,11 +11,11 @@ export const currentEnvironmentRole = (
...
 
-  let curEnvRole = account.environmentRoles.find(
-    environmentRole =>
-      environmentRole.environment.id === curEnvId ||
-      environmentRole.environment.urlCode === curEnvId
-  );
+  let curEnvRole = account.environmentRoles.find(environmentRole => {
+    const { environment } = environmentRole;
+    if (environment.id && curEnvId) return environment.id === curEnvId;
+    return environment.urlCode === curEnvId;
```
cre8ivejp, line 17: @steveninhle, just confirming.
Why do we need to compare the urlCode with the environment ID?
In most cases the environment uses a UUID, so it always will be false.

steveninhle: @cre8ivejp  `curEnvId ` variable will get 2 values from local storage, if the environment has the `id` will save the `id` or not (like `production` env), we will save the `urlCode`. 

It's a bit confusing for this comparison, but we need to check with environments without `id`

cre8ivejp: @steveninhle, is this implementation necessary because we have environments using empty IDs?

steveninhle: @cre8ivejp yep, since we have environments that use empty IDs, we need to check `urlCode` instead,

cre8ivejp (approved):
Thank you 🎉

Issue #1475, opened 2025-01-22
change the method of specifying query conditions


kakcy:
Assembling the Where clause and Order clause is done at the API layer.

### Example of code that is an issue
- `AutoOps`
https://github.com/bucketeer-io/bucketeer/blob/main/pkg/autoops/api/api.go#L1115

### Solution
- Define a common condition specification structure(ListOptions) that specifies acquisition conditions in the List system, and specify conditions from the API layer via it.
 The storage layer converts it into conditions suitable for DB operations and uses it.

### Package

- [ ] account

- [ ] auditlog

- [ ] autoops

- [ ] environment

- [ ] eventcounter

- [ ] experiment

- [ ] feature

- [ ] notification

- [ ] push

- [ ] experimentcalculator

- [ ] mau

- [ ] opsevent

- [ ] subscriber

PR #1661, opened 2025-03-31
refactor: change the list retrieval API in the notification package to use ListOption

kakcy:
This pull request focuses on refactoring the subscription listing methods in the `NotificationService` to simplify the code and enhance maintainability. The changes include replacing the use of `whereParts` and other individual parameters with a unified `ListOptions` struct, which consolidates filters, orders, and other query options.

relate of #1475

### DIFF
--- a/pkg/notification/api/subscription.go
+++ b/pkg/notification/api/subscription.go
@@ -903,7 +903,7 @@ func (s *NotificationService) listSubscriptionsMySQL(
 		NullFilters: nil,
 		JSONFilters: jsonFilters,
 		SearchQuery: seachQuery,
-		Orders:      nil,
+		Orders:      orders,
 	}
 
 	subscriptions, nextCursor, totalCount, err := s.subscriptionStorage.ListSubscriptions(
--- a/pkg/notification/api/subscription.go
+++ b/pkg/notification/api/subscription.go
@@ -722,30 +722,6 @@ func (s *NotificationService) ListSubscriptions(
 	if err != nil {
 		return nil, err
 	}
-	var whereParts []mysql.WherePart
-	if req.OrganizationId != "" {
-		// New console
-		whereParts = append(whereParts, mysql.NewFilter("env.organization_id", "=", req.OrganizationId))
-	} else {
-		// Current console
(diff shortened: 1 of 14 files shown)


pkg/notification/storage/v2/subscription.go
```diff
@@ -179,20 +176,15 @@ func (s *subscriptionStorage) GetSubscription(
...
-	limitOffsetSQL := mysql.ConstructLimitOffsetSQLString(limit, offset)
-	query := fmt.Sprintf(selectSubscriptionV2AnySQLQuery, whereSQL, orderBySQL, limitOffsetSQL)
+	query, whereArgs := mysql.ConstructQueryAndWhereArgs(selectSubscriptionV2AnySQLQuery, options)
 	rows, err := s.qe.QueryContext(ctx, query, whereArgs...)
 	if err != nil {
 		return nil, 0, 0, err
 	}
 	defer rows.Close()
-	subscriptions := make([]*proto.Subscription, 0, limit)
+	subscriptions := make([]*proto.Subscription, 0)
```
nnnkkk7, line 187: How about this?
```suggestion
	var offset int
	var limit int
	if options != nil {
		offset = options.Offset
		limit = options.Limit
	}
	subscriptions := make([]*proto.Subscription, 0, limit)
```

pkg/notification/api/subscription.go
```diff
@@ -854,12 +828,58 @@ func (s *NotificationService) ListEnabledSubscriptions(
...
+	disabled *bool,
+	searchKeyword string,
 	orders []*mysql.Order,
 	pageSize int64,
 	cursor string,
 	localizer locale.Localizer,
 ) ([]*notificationproto.Subscription, string, int64, error) {
+	var filters []*mysql.FilterV2
+	if organizationId != "" {
+		// New console
```
nnnkkk7, line 843: I'd think`// console v3` is clearer.

nnnkkk7: @kakcy 
Thanks！
And the comment in L850 would be `console v2`...🙇

nnnkkk7 (approved):
Thank you very much!
Nice!!

PR #1663, opened 2025-03-31
fix: duplicate features returned in the list api

hvn2k1:
List features return duplicated data because pair of feature id and environment can exist in different version in feature_last_used_info table

### DIFF
--- a/pkg/feature/storage/v2/sql/feature/select_features_by_experiment.sql
+++ b/pkg/feature/storage/v2/sql/feature/select_features_by_experiment.sql
@@ -58,5 +58,6 @@ LEFT OUTER JOIN feature_last_used_info ON
     feature.version = feature_last_used_info.version
 LEFT OUTER JOIN experiment ON
     feature.id = experiment.feature_id AND
-    feature.environment_id = experiment.environment_id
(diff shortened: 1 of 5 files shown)


pkg/feature/storage/v2/sql/feature/select_features_by_experiment.sql
```diff
@@ -54,7 +54,8 @@ FROM
     feature
 LEFT OUTER JOIN feature_last_used_info ON
     feature.id = feature_last_used_info.feature_id AND
-    feature.environment_id = feature_last_used_info.environment_id
+    feature.environment_id = feature_last_used_info.environment_id AND
+    feature.version = feature_last_used_info.version
 LEFT OUTER JOIN experiment ON
     feature.id = experiment.feature_id AND
     feature.environment_id = experiment.environment_id
```
cre8ivejp, line 61: We also should filter by version when joining the experiment table.

hvn2k1: thank you, I have added version when joining experiment too

cre8ivejp (approved):
Thank you 🎉

Issue #1670, opened 2025-04-01
fix: duplicate flags when listing flags that contain experiments


cre8ivejp:
Because a single flag can contain many experiments, we must use `INNER JOIN` instead.
Also, the condition is to list a flag that contains at least one experiment, so I removed the `version` from the condition. Otherwise, it could return nothing if the version didn't match.

PR #1671, opened 2025-04-01
chore: correctly filter feature with experiment

hvn2k1:
- Left join can return features without experiment, switch to join
- Remove feature version join condition, should find in any version

Fix https://github.com/bucketeer-io/bucketeer/issues/1670


cre8ivejp (approved):
LGTM 👍

PR #1672, opened 2025-04-01
fix: add validation for empty variation_id and set 'default'

nnnkkk7:
- Add validation for empty variation_id when the reason is about error.
- Set  variation_id 'default' when the reason is default, variation_id is empty in dwh subscriber. 
- Add tests.

### DIFF
--- a/pkg/subscriber/processor/evaluation_events_dwh.go
+++ b/pkg/subscriber/processor/evaluation_events_dwh.go
@@ -196,13 +196,20 @@ func (w *evalEvtWriter) convToEvaluationEvent(
 		// Tag is optional, so we insert none when is empty.
 		tag = "none"
 	}
+	variationID := e.VariationId
+	if variationID == "" {
+		// When the default value is returned to the app in the SDK,
+		// it creates a default evaluation event with an empty variation ID.
+		// Because we don't have the variation ID we insert "default" as the variation_id
+		variationID = "default"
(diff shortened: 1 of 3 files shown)


pkg/subscriber/processor/evaluation_events_dwh.go
```diff
@@ -196,13 +196,19 @@ func (w *evalEvtWriter) convToEvaluationEvent(
 		// Tag is optional, so we insert none when is empty.
 		tag = "none"
 	}
+	var variationId string
+	if e.VariationId == "" {
+		// if the reason is default, the variation_id is empty
+		// so we insert "default" as the variation_id
+		variationId = "default"
+	}
```
cre8ivejp, line 204: In the current logic, if the `e.VariationId` is not empty, the `variationId` will be empty when saving the event to BigQuery, causing errors.

```suggestion
	variationID := e.VariationId
	if variationID == "" {
		// When the default value is returned to the app in the SDK,
		// it creates a default evaluation event with an empty variation ID.
		// Because we don't have the variation ID we insert "default" as the variation_id
		variationID = "default"
	}
```

nnnkkk7: Oh,thank you!
I made a mistake...

cre8ivejp (approved):
Thank you!

PR #1675, opened 2025-04-01
chore: remove unnecessary error logs in the api svc

cre8ivejp:
Related to https://github.com/bucketeer-io/bucketeer/pull/1566

### DIFF
--- a/pkg/subscriber/processor/evaluation_events_evaluation_count_event_persister.go
+++ b/pkg/subscriber/processor/evaluation_events_evaluation_count_event_persister.go
@@ -37,6 +37,11 @@ import (
 	featureproto "github.com/bucketeer-io/bucketeer/proto/feature"
 )
 
+var (
+	// goredis doesn't export a dedicated sentinel for client close error
+	errRedisClientClosed = errors.New("redis: client is closed")
+)
+
 const (
 	eventCountKey      = "ec"
 	userCountKey       = "uc"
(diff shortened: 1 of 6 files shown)


nnnkkk7 (approved):
Thanks!
LGTM!

pkg/subscriber/processor/evaluation_events_evaluation_count_event_persister.go
```diff
@@ -247,7 +252,7 @@ func (p *evaluationCountEventPersister) incrementEvaluationCount(
 		ucKey := p.newEvaluationCountkeyV2(userCountKey, e.FeatureId, vID, environmentId, e.Timestamp)
 		userID := getUserID(e.UserId, e.User)
 		if err := p.countUser(ucKey, userID); err != nil {
-			if err != nil {
+			if !errors.Is(err, errRedisClientClosed) {
```
nnnkkk7, line 255: I think this `errors.Is(err, errRedisClientClosed)` always returns false..👀
`errors.Is()` returns true if the target has same instance (not same error message)
So we should add custom Is() method or compare error message string.
https://pkg.go.dev/errors#Is

cre8ivejp: Thanks for pointing it out.
I have fixed https://github.com/bucketeer-io/bucketeer/pull/1675/commits/82473d34e8b0fab6c22d5d138c2826f6779fcbc1.

nnnkkk7 (approved):
Thank you!
LGTM!

PR #1679, opened 2025-04-01
fix: docker load image failure when it doesn't have privileges

sivchari:
When the docker doesn't have privilege, docker load must be failed since the permission issue.


Ubisoft-potato (approved):
Thanks!

cre8ivejp (approved):
Thank you so much for your contribution!

PR #1680, opened 2025-04-01
chore: improve docker build when using dev container on macos

sivchari:
To develop on macOS, it's better to pass the platform flag and fix the platform env for Go.


Ubisoft-potato (approved):
Great! 👍

cre8ivejp (approved):
Nice!
Thank you!

PR #1683, opened 2025-04-02
fix(ui/dashboard): select flags and goal drop menu are not showing the option to create a new one

### LINKED ISSUE #1662: fix(ui/dashboard): select flags and goal drop menu are not showing the option to create a new one
When there is no flag or goal, the drop menu should show "No options found" and also, show in the bottom of the drop menu an option to create a new one.

steveninhle:
When there is no flag or goal, the drop menu should show "No options found" and also, show in the bottom of the drop menu an option to create a new one.

Fix #1662


cre8ivejp (approved):
Thank you!

Issue #311, opened 2023-05-08
feat: new dashboard V3


cre8ivejp:
A new UI is coming!

- [x] UX Audit
- [x] Design
- [x] Backend

- [x] Frontend development
  - [x] Login 
  - [x] Organizations
  - [x] Projects
  - [x] Environments
  - [x] Members (Account)
  - [x] API Keys
  - [x] Integrations (Slack and FCM)
  - [x] Debugger
  - [x] User Segments
  - [x] Audit Logs
  - [x] Goals
  - [x] Experiments
  - [x] Features
    - [x] List
    - [x] Settings
    - [x] Insights (Evaluations)
    - [x] Audit Logs
    - [x] Triggers
    - [x] Targeting
    - [x] Variations
    - [x] Auto Operations

PR #1684, opened 2025-04-02
feat: return entity name and editor image for list audit logs API

hvn2k1:
Part of https://github.com/bucketeer-io/bucketeer/issues/311


proto/event/domain/event.proto
```diff
@@ -279,12 +279,14 @@ message Editor {
...
     string maintainer = 2;
     string name = 3;
+    string avatar_image_url = 4;
   }
   string email = 1;
   reserved 2;  // bucketeer.account.Account.Role role = 2 [deprecated = true];
   bool is_admin = 3;
   string name = 4;
   PublicAPIEditor public_api_editor = 5;
+  string avatar_image_url = 6;
```
cre8ivejp, line 289: Same here.
Reference: [r2025044893](https://github.com/bucketeer-io/bucketeer/pull/1684/files#r2025044893)

pkg/account/storage/v2/account.go
```diff
@@ -211,6 +213,71 @@ func (s *accountStorage) GetAccountV2ByEnvironmentID(
...
+			&mysql.JSONObject{Val: &account.SearchFilters},
+		)
+		if err != nil {
+			return nil, err
+		}
+		account.OrganizationRole = proto.AccountV2_Role_Organization(organizationRole)
+		accounts = append(accounts, account)
+	}
+	if rows.Err() != nil {
+		return nil, err
```
nnnkkk7, line 276:
```suggestion
		return nil, rows.Err()
```

pkg/account/storage/v2/account.go
```diff
@@ -211,6 +213,71 @@ func (s *accountStorage) GetAccountV2ByEnvironmentID(
...
+	rows, err := s.qe.QueryContext(
+		ctx,
+		query,
+		whereArgs...,
+	)
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
+	accounts := make([]*proto.AccountV2, 0)
```
nnnkkk7, line 246: How about this?
```suggestion
	accounts := make([]*proto.AccountV2, 0, len(emails))
```

pkg/account/storage/v2/account_test.go
```diff
@@ -284,6 +284,60 @@ func TestGetAccountV2ByEnvironmentID(t *testing.T) {
...
+				).Return(rows, nil)
+			},
+			expectedErr:   nil,
+			emails:        []string{"test@example.com"},
+			environmentID: "env-0",
+		},
+	}
+	for _, p := range patterns {
+		t.Run(p.desc, func(t *testing.T) {
+			storage := newAccountStorageWithMock(t, mockController)
```
nnnkkk7, line 327: It would be better to add `t.Parallel()`  below `t.Run(p.desc, func(t *testing.T) `
ref: https://engineering.mercari.com/en/blog/entry/20220408-how_to_use_t_parallel/

pkg/auditlog/api/api.go
```diff
@@ -164,9 +169,39 @@ func (s *auditlogService) ListAuditLogs(
 		}
 		return nil, dt.Err()
 	}
-	for _, auditlog := range auditlogs {
-		auditlog.LocalizedMessage = domainevent.LocalizedMessage(auditlog.Type, localizer)
+	editorEmails := make([]string, len(auditlogs))
```
nnnkkk7, line 172:
```suggestion
	editorEmails := make([]string, 0, len(auditlogs))
```

pkg/auditlog/api/api.go
```diff
@@ -164,9 +169,39 @@ func (s *auditlogService) ListAuditLogs(
...
+	for i, auditlog := range auditlogs {
+		auditlogs[i].LocalizedMessage = domainevent.LocalizedMessage(auditlog.Type, localizer)
+		// unmarshal json string to map entity name
+		var entityData map[string]interface{}
+		err := json.Unmarshal([]byte(auditlog.EntityData), &entityData)
+		if err != nil {
+			continue
+		}
+		if entityName, ok := entityData["name"]; ok {
+			auditlogs[i].EntityName = entityName.(string)
```
nnnkkk7, line 182: Does this need type assertion check?

nnnkkk7 (commented):
Thank you!
I commented!

pkg/auditlog/api/api.go
```diff
@@ -164,9 +169,39 @@ func (s *auditlogService) ListAuditLogs(
...
+
+	for i := range auditlogs {
+		if account, ok := accounts[auditlogs[i].Editor.Email]; ok {
+			auditlogs[i].Editor.AvatarImage = account.AvatarImage
+			auditlogs[i].Editor.AvatarFileType = account.AvatarFileType
+			if auditlogs[i].Editor.PublicApiEditor != nil {
+				auditlogs[i].Editor.PublicApiEditor.AvatarImage = account.AvatarImage
+				auditlogs[i].Editor.PublicApiEditor.AvatarFileType = account.AvatarFileType
+			}
+		}
```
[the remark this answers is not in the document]
cre8ivejp: When creating the Editor, we should ensure we also set the `imageAvatar` and the `AvatarFileType`.
So, for the new logs, the Editor saved in the Audit Logs will have the avatar, not needing the implementation above.

And for old logs, we can do a migration or implement that logic on the fronend.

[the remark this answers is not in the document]
hvn2k1: I've thinking about it, but it would make audit log table carry more data. Also in pubsub, we have to stream editor images too, I also worry about whether we need to share the avatar image between context. Whereas in this api alone, which is not requested so frequently, it only took another query to get all the avatar images of the editor if I do this. WDYT?

pkg/account/storage/v2/account.go
hvn2k1: There might be a case that the account with email not found, so I'm not so sure about the size of the accounts slice

pkg/auditlog/api/api.go
hvn2k1: Yes, because auditlog.EntityData is a json string with possible multiple level, so we'll need entityData as map[string]interface{}, and we'll need to cast the interface to string after

pkg/auditlog/api/api.go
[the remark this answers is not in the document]
cre8ivejp: By saving the avatar at the time the log is created, we would have the avatar even if the user was deleted from the organization.
It depends on what we want. If you decide that it is ok not to have the avatar for cases where the user was removed from the organization, then we don't need to save it, and we can keep the current implementation.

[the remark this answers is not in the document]
hvn2k1: Thank you, if the account is deleted, I think not showing the avatar is ok!
I also tested the query with multiple avatar, it's not affecting the list audit log performance.

pkg/auditlog/api/api.go
hvn2k1: Thank you, I have updated

pkg/account/storage/v2/account.go
hvn2k1: I saw similar errors in the file too, updated

pkg/auditlog/api/api.go
```diff
@@ -390,6 +412,55 @@ func (s *auditlogService) ListFeatureHistory(
...
+	environmentID string,
+	localizer locale.Localizer,
+) (accountMap map[string]*accountproto.AccountV2, err error) {
+	accountMap = make(map[string]*accountproto.AccountV2)
+	if len(emails) == 0 {
+		return accountMap, nil
+	}
+	if len(emails) == 0 {
+		return nil, nil
+	}
```
cre8ivejp, line 427:
```suggestion
	if len(emails) == 0 {
		return accountMap, nil
	}
```

cre8ivejp (approved):
Thank you 🎉

PR #1689, opened 2025-04-03
fix: docker build fails when using apple cpu architecture

cre8ivejp:
Due to the changes made in this [PR](https://github.com/bucketeer-io/bucketeer/pull/1680), the docker build fails on the `aarch64` Mac.


nnnkkk7 (approved):
Thank you!

PR #1696, opened 2025-04-08
chore: apply convention for public API url

hvn2k1:


hvn2k1:
> @hvn2k1, just confirming, you are using the same URL path used in the web svc, right?

Yes I am

cre8ivejp (approved):
Thank you 🎉

PR #1698, opened 2025-04-08
perf: optimize GetEvaluationTimeseriesCount for concurrent processing

### LINKED ISSUE #1635: slow response when getting the evaluation count
When showing the evaluation count on the flag details evaluation tab, it takes several seconds, even when getting only 7 days of data. I checked the Redis monitoring, and the requests are done really fast and in a few milliseconds. So, something is happening between the response from Redis and the response to the console.
The app creates the time series data using the counts retrieved from Redis. Maybe something is being done wrong.

nnnkkk7:
fix #1635 

The local benchmark test showed this change makes the execution 2x faster.

### DIFF
--- a/pkg/eventcounter/api/api.go
+++ b/pkg/eventcounter/api/api.go
@@ -32,6 +32,7 @@ import (
 	accountclient "github.com/bucketeer-io/bucketeer/pkg/account/client"
 	"github.com/bucketeer-io/bucketeer/pkg/cache"
 	cachev3 "github.com/bucketeer-io/bucketeer/pkg/cache/v3"
+	"github.com/bucketeer-io/bucketeer/pkg/errgroup"
 	v2ecstorage "github.com/bucketeer-io/bucketeer/pkg/eventcounter/storage/v2"
 	experimentclient "github.com/bucketeer-io/bucketeer/pkg/experiment/client"
 	featureclient "github.com/bucketeer-io/bucketeer/pkg/feature/client"
@@ -297,14 +298,10 @@ func (s *eventCounterService) GetEvaluationTimeseriesCount(
 	variationTotalUserCountsMap := make(map[string]int64, len(vIDs))
 
 	var mu sync.Mutex
-	var wg sync.WaitGroup
-	var processingErr error
-	errChan := make(chan error, len(vIDs))
+	var eg errgroup.Group
 
-	for _, vID := range vIDs {
-		wg.Add(1)
-		go func(variationID string) {
-			defer wg.Done()
+	for _, variationID := range vIDs {
+		eg.Go(func() error {
 
 			// Get event data
 			eventCountKeys := s.getEventCountKeys(
@@ -322,8 +319,7 @@ func (s *eventCounterService) GetEvaluationTimeseriesCount(
 					resp.Feature.Version,
 					timestampUnit, req.TimeRange,
 				)
-				errChan <- err
-				return
+				return err
 			}
 
 			// Get user data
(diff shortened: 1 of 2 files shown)


pkg/eventcounter/api/api.go
```diff
@@ -289,104 +290,139 @@ func (s *eventCounterService) GetEvaluationTimeseriesCount(
...
+
+	variationEventCountsMap := make(map[string][]float64, len(vIDs))
+	variationUserCountsMap := make(map[string][]float64, len(vIDs))
+	variationTotalEventCountsMap := make(map[string]int64, len(vIDs))
+	variationTotalUserCountsMap := make(map[string]int64, len(vIDs))
+
+	var mu sync.Mutex
+	var wg sync.WaitGroup
+	var processingErr error
+	errChan := make(chan error, len(vIDs))
```
cre8ivejp, line 302: It's just a suggestion.
Instead of using channels and mutex, you could simplify the code using `errgroup.Group` and `sync.Map`.
When the flag has multiple variations (3 or more), the `sync.Map` might slightly perform better, too.

```suggestion
	var variationEventCountsMap sync.Map    // []float64
	var variationUserCountsMap sync.Map       // []float64
	var variationTotalEventCountsMap sync.Map // int64
	var variationTotalUserCountsMap sync.Map  // int64

         var eg errgroup.Group
```

```go
for _, vID := range vIDs {
        vID := vID
        eg.Go(func() error {
                count, err : getCount(vID)
                if err != nil {
                        // log error
                        return err
                }
                // save the count in the map
                return nil
        }
}

// Wait for all goroutines to complete
if err := eg.Wait(); err != nil {
        dt, errDt := statusInternal.WithDetails(&errdetails.LocalizedMessage{
	        Locale:  localizer.GetLocale(),
	        Message: localizer.MustLocalize(locale.InternalServerError),
        })
        if errDt != nil {
	        return nil, statusInternal.Err()
        }
        return nil, dt.Err()
}
```

nnnkkk7: Thank you for your suggestion!
I've introduced `errgroup.Group` because it made error handling easier and a bit faster!!
But when I use `sync.Map`, it made this function a bit slower maybe because of type assertion of getting values from `sync.Map`...
So I only fixed errror handling with  `errgroup.Group`!!
Thank you very much!

pkg/eventcounter/api/api.go
```diff
@@ -289,104 +291,124 @@ func (s *eventCounterService) GetEvaluationTimeseriesCount(
...
+	variationEventCountsMap := make(map[string][]float64, len(vIDs))
+	variationUserCountsMap := make(map[string][]float64, len(vIDs))
+	variationTotalEventCountsMap := make(map[string]int64, len(vIDs))
+	variationTotalUserCountsMap := make(map[string]int64, len(vIDs))
+
+	var mu sync.Mutex
+	var eg errgroup.Group
+
+	for _, variationID := range vIDs {
+		eg.Go(func() error {
```
cre8ivejp, line 304: By using the `variationID` directly from the loop, depending on the time, the `variationID` will be overridden by the loop, and the functions will get the wrong counter when making requests to Redis.

```suggestion
	for _, vID := range vIDs {
	        variationID := vID
		eg.Go(func() error {
```

nnnkkk7: Oh, this problem was fixed in Go1.22(https://go.dev/blog/loopvar-preview).
From Go1.22,  each iteration of the for loop gets its own copy of the loop variable. So we don't need redeclare the variable!

cre8ivejp (approved):
Thank you 🎉

PR #1685, opened 2025-04-02
refactor: change the list retrieval API in the auditlog package to use ListOption

kakcy:
This pull request improves the `auditlogService` by refactoring the code to use a new filtering mechanism and updating the associated tests. The main changes involve replacing the old `mysql.WherePart` with the new `mysql.FilterV2` and consolidating query options into a single `mysql.ListOptions` struct.

relate of https://github.com/bucketeer-io/bucketeer/issues/1475

PR #1702, opened 2025-04-09
revert: change the list retrieval API in the auditlog package to use ListOption

kakcy:
Reverts bucketeer-io/bucketeer#1685


nnnkkk7 (approved):
Thank you!

PR #1705, opened 2025-04-10
feat(ui/dashboard): feature flag details variations page

steveninhle:


ui/dashboard/src/@icons/customized-icons/filename (1).txt
cre8ivejp, line 1: Is this filename correct?

cre8ivejp (approved):
Thank you!

Issue #1617, opened 2025-03-25
feat: be able to evaluate a user for a specific version in the debugger


cre8ivejp:
Currently, we only evaluate the user using the latest flag data in the Debugger. For data analysis, it would also be very helpful to check what variation the user got for old flag versions.

Since we save the previous and the current flag objects in the audit log table, we could make this work by getting the specific flag using the `updated_at` field.

- [ ] Support a specific flag version using the `updated_at` field
- [ ] Implement the Public API for the Debugger tool

hvn2k1:
Currently the depend flags are always refer to the latest version, so prerequisite and rules can not know which version to calculate when we refer to older version.
So due to the data structure, we cannot evaluate correctly the past data for now. The feature request will be multi user evaluation at the moment.

PR #1706, opened 2025-04-10
feat: implement debug evaluate multiple users with features

hvn2k1:
Part of https://github.com/bucketeer-io/bucketeer/issues/1617

- Currently not able to evaluate depend features for older version


proto/feature/service.proto
```diff
@@ -542,6 +542,26 @@ message EvaluateFeaturesResponse {
...
+  message Feature {
+    string id = 1;
+    int32 feature_version = 2
+        [(grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
+          description: "If not set, the latest version will be used."
+        }];
+  }
+  repeated bucketeer.user.User users = 1
+      [(google.api.field_behavior) = REQUIRED];
+  string tag = 2;
```
cre8ivejp, line 555: We won't need  the tag.

pkg/feature/api/feature.go
```diff
@@ -2390,6 +2391,222 @@ func (s *FeatureService) EvaluateFeatures(
...
+		}
+		return nil, nil, dt.Err()
+	}
+	latestFeatures := fs.([]*featureproto.Feature)
+	latestEvaluationResult := make([]*featureproto.Evaluation, 0)
+	archivedFS := make([]string, 0)
+	for i := range req.Users {
+		userEvaluations, err := s.evaluateFeatures(
+			ctx,
+			latestFeatures, req.Users[i], req.EnvironmentId, req.Tag, localizer)
```
cre8ivejp, line 2481: You are evaluating all the flags in the environment per user, and the request might have only one feature to be returned, causing slow issues in an environment with thousands of flags.

It's ok to get all the flags from the cache or the DB, but you should only evaluate the target flags, including the prerequisite and rules using flags if needed.

You should filter the flags by using this function `GetFeaturesDependedOnTargets`.
https://github.com/bucketeer-io/bucketeer/blob/main/pkg/feature/domain/feature.go#L1450-L1471https://github.com/bucketeer-io/bucketeer/blob/main/pkg/feature/domain/feature.go#L1450-L1471

pkg/feature/api/feature.go
```diff
@@ -2390,6 +2391,222 @@ func (s *FeatureService) EvaluateFeatures(
...
+		olderFeatures[i].Id = fmt.Sprintf("%s:%d", olderFeatures[i].Id, olderFeatures[i].Version)
+	}
+
+	evaluations := make([]*featureproto.Evaluation, 0)
+	archivedFS := make([]string, 0)
+
+	for i := range req.Users {
+		userEvaluations, err := s.evaluateFeatures(
+			ctx,
+			olderFeatures,
```
cre8ivejp, line 2534: The old features here don't include the flags being used in the prerequisites and rules targeting.

pkg/feature/api/feature.go
hvn2k1: After discussions let's skip the version for now

pkg/feature/api/feature.go
```diff
@@ -2390,6 +2390,117 @@ func (s *FeatureService) EvaluateFeatures(
...
+		}
+		return nil, dt.Err()
+	}
+
+	features := fs.([]*featureproto.Feature)
+	var evaluations = make([]*featureproto.Evaluation, 0)
+	var archivedFS = make([]string, 0)
+
+	for i := range req.Users {
+		if len(req.FeatureIds) == 1 {
```
nnnkkk7, line 2444: Could you tell me why we get target features when featureid is one?

hvn2k1: Currently, because there are depended features so we get all features in the environment first and use flightgroup to reduce latency. And to avoid complex logics, we evaluate all features if we find dependency in the same environment https://github.com/bucketeer-io/bucketeer/blob/76320c376abf3a8b5fe718edf252eff53fe9a7c1/pkg/feature/api/feature.go#L2406

If there is no dependency and only 1 flag is requested to evaluate, we only evaluate that flag only. So in here, we get target features just to check if there is any dependency in the flag, else it will return every flag in the env. This is the logic I derived from the original evaluate feature api: https://github.com/bucketeer-io/bucketeer/blob/76320c376abf3a8b5fe718edf252eff53fe9a7c1/pkg/feature/api/feature.go#L2067

pkg/feature/api/feature.go
```diff
@@ -2390,6 +2390,117 @@ func (s *FeatureService) EvaluateFeatures(
...
+			}
+			return nil, dt.Err()
+		}
+
+		evaluations = append(evaluations, userEvaluations.Evaluations...)
+		archivedFS = append(archivedFS, userEvaluations.ArchivedFeatureIds...)
+	}
+	evaluationResults := make([]*featureproto.Evaluation, 0)
+	for _, eval := range evaluations {
+		for _, id := range req.FeatureIds {
```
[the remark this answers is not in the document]
hvn2k1: updated, thank you

pkg/feature/api/feature.go
nnnkkk7: Thank you!
I got it!
So how about moving this logic(`features, err = s.getTargetFeatures(features, req.FeatureIds[0], localizer)...`) above for loop?

nnnkkk7: And it would be better to add comment about the reason!

hvn2k1: > So how about moving this logic(features, err = s.getTargetFeatures(features, req.FeatureIds[0], localizer)...) above for loop?
Nice catch, I've moved that outside! and added the comment

proto/gateway/service.proto
```diff
@@ -557,6 +568,19 @@ service Gateway {
...
   }
+  rpc DebugEvaluateFeatures(DebugEvaluateFeaturesRequest)
+      returns (DebugEvaluateFeaturesResponse) {
+    option (google.api.http) = {
+      post: "/v1/debug_evaluate_features"
+      body: "*"
+    };
+    option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
+      summary: "Debug Evaluate Features",
+      description: "This API is for debugging the evaluation of feature flags. To call this API, you need a `READ` API Key role.",
```
cre8ivejp, line 579:
```suggestion
      description: "This API is used to debug one or more end-user evaluations. To call this API, you need a `READ` API Key role.",
```

proto/feature/service.proto
```diff
@@ -542,6 +542,18 @@ message EvaluateFeaturesResponse {
...
 
+message DebugEvaluateFeaturesRequest {
+  repeated bucketeer.user.User users = 1
+      [(google.api.field_behavior) = REQUIRED];
+  repeated string feature_ids = 3 [(google.api.field_behavior) = REQUIRED];
+  string environment_id = 4 [(google.api.field_behavior) = REQUIRED];
+}
+
+message DebugEvaluateFeaturesResponse {
+  repeated Evaluation evaluations = 1;
```
cre8ivejp, line 553: The UI groups the results by user or flags.
Returning a list of all evaluations would make it hard to implement this.

We should think about how to group the results in the response.
WDTY?

hvn2k1: Hmm so we add the group option and return response base on the input?

cre8ivejp: I talked to the engineer and he will do this on the fronend. Thanks!

cre8ivejp (approved):
Nice!
Thank you 🎉

pkg/feature/api/feature.go
```diff
@@ -2392,6 +2393,115 @@ func (s *FeatureService) EvaluateFeatures(
...
+			dt, err := statusInternal.WithDetails(&errdetails.LocalizedMessage{
+				Locale:  localizer.GetLocale(),
+				Message: localizer.MustLocalize(locale.InternalServerError),
+			})
+			if err != nil {
+				return nil, statusInternal.Err()
+			}
+			return nil, dt.Err()
+		}
+	}
```
nnnkkk7, line 2465: Now,
If `len(req.FeatureIds) == 1`→
getFeatures in  `flightgroup.Do` and  execute `getTargetFeatures(features, req.FeatureIds[0], localizer)`
If `len(req.FeatureIds) ！= 1`→
getFeatures in  `flightgroup.Do`

So how about only executing `getTargetFeatures(features, req.FeatureIds[0], localizer)` if `len(req.FeatureIds) == 1`?

nnnkkk7 (commented):
Sorry for the late reply, I have one more question🙏

pkg/feature/api/feature.go
nnnkkk7: Thank you!

pkg/feature/api/feature.go
hvn2k1: Because getTargetFeatures function check all depend flags in the same env, so if we don't have list of features in that env, the result might not correct.
You can check the implementation here: https://github.com/bucketeer-io/bucketeer/blob/37e2f58cba7afeaf83c8488ab5fe82cee1752eed/pkg/feature/api/feature.go#L2395-L2414

proto/feature/service.proto
```diff
@@ -1389,6 +1401,52 @@ service FeatureService {
...
   rpc EvaluateFeatures(EvaluateFeaturesRequest)
       returns (EvaluateFeaturesResponse) {}
+  rpc DebugEvaluateFeatures(DebugEvaluateFeaturesRequest)
+      returns (DebugEvaluateFeaturesResponse) {
+    option (google.api.http) = {
+      post: "/v1/debug_evaluate_features"
+      body: "*"
+    };
+    option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
+      summary: "Debug Evaluate Features",
```
cre8ivejp, line 1411:
```suggestion
      summary: "Debug Feature Evaluations",
```

proto/gateway/service.proto
```diff
@@ -557,6 +568,19 @@ service Gateway {
...
     };
   }
+  rpc DebugEvaluateFeatures(DebugEvaluateFeaturesRequest)
+      returns (DebugEvaluateFeaturesResponse) {
+    option (google.api.http) = {
+      post: "/v1/debug_evaluate_features"
+      body: "*"
+    };
+    option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
+      summary: "Debug Evaluate Features",
```
cre8ivejp, line 578:
```suggestion
      summary: "Debug Feature Evaluations",
```

nnnkkk7 (approved):
LGTM!
Thanks!

PR #1707, opened 2025-04-10
feat: mysql event table

Ubisoft-potato:


hvn2k1:
Hi @Ubisoft-potato @cre8ivejp, we will support mysql for bigquery alternative then we migrate to postgres, is that correct?

Ubisoft-potato:
> Hi @Ubisoft-potato @cre8ivejp, we will support mysql for bigquery alternative then we migrate to postgres, is that correct?

Yes, we will support mysql first, then postgre.

manifests/bucketeer/values.yaml
```diff
@@ -2,6 +2,7 @@ global:
   image:
     tag: v1.3.0 # x-release-please-version
     imagePullSecrets:
+  useMySQL: true
```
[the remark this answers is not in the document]
Ubisoft-potato: That's great!
This way will be more manageable for the db in the future. 

Let me make this changes.

manifests/bucketeer/values.dev.yaml
```diff
@@ -491,13 +527,24 @@ subscriber:
...
+          useMainConnection: false  # Will be set by template
+          host: ""  # Will be set by template
+          port: 0  # Will be set by template
+          user: ""  # Will be set by template
+          password: ""  # Will be set by template
+          database: ""  # Will be set by template
+        bigquery:
+          project: ""  # Will be set by template
+          dataset: ""  # Will be set by template
+          location: ""  # Will be set by template
```
cre8ivejp, line 547: Can you remove the comments from here, too?

migration/mysql/20250502025508_add_events_table.sql
```diff
@@ -0,0 +1,46 @@
...
+    PRIMARY KEY (`id`),
+    INDEX `idx_evaluation_environment_id` (`environment_id`),
+    INDEX `idx_evaluation_timestamp` (`timestamp`),
+    INDEX `idx_evaluation_feature_id` (`feature_id`),
+    INDEX `idx_evaluation_user_id` (`user_id`),
+    INDEX `idx_evaluation_variation_id` (`variation_id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
+
+-- Create Goal Event Table
+CREATE TABLE `goal_event` (
```
cre8ivejp, line 24: @Ubisoft-potato, cc: @hvn2k1 

I have one concern regarding the migration.
By doing that in the same database as the console database, we are assuming that when using MySQL for the datahouse, we will use the same database for both cases, but the user is allowed to configure the database manually, and it could be different.
In that case, we would need an initialization DB including those tables, and we don't have one at the moment.

Any suggestions?

Ubisoft-potato: Great point!
I think we can add a script to let user configure the database and initialize the table.
WDYT?

hack/create-mysql-event-tables/command.go
```diff
@@ -0,0 +1,239 @@
...
+			existingTables = append(existingTables, table)
+		}
+	}
+
+	return existingTables, nil
+}
+
+func (c *command) readSQLFile() (string, error) {
+	// Try to read from the provided path first
+	if content, err := os.ReadFile(*c.sqlFilePath); err == nil {
```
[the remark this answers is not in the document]
Ubisoft-potato: Ah, I missed it, let me change to use embedded file! 👍

hack/create-mysql-event-tables/create_events_table.sql
```diff
@@ -0,0 +1,46 @@
...
+    PRIMARY KEY (`id`),
+    INDEX `idx_evaluation_environment_id` (`environment_id`),
+    INDEX `idx_evaluation_timestamp` (`timestamp`),
+    INDEX `idx_evaluation_feature_id` (`feature_id`),
+    INDEX `idx_evaluation_user_id` (`user_id`),
+    INDEX `idx_evaluation_variation_id` (`variation_id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
+
+-- Create Goal Event Table
+CREATE TABLE `goal_event` (
```
cre8ivejp, line 24:
```suggestion
CREATE TABLE IF NOT EXISTS `goal_event` (
```

hack/create-mysql-event-tables/create_events_table.sql
```diff
@@ -0,0 +1,46 @@
+-- Create Evaluation Event Table
+CREATE TABLE `evaluation_event` (
```
cre8ivejp, line 2:
```suggestion
CREATE TABLE IF NOT EXISTS `evaluation_event` (
```

cre8ivejp (approved):
Excellent work!
Thank you 🎉

PR #1708, opened 2025-04-10
fix: add localized message to ListAuditLogs

nnnkkk7:
Fix this error in audit log page,
`TypeError: Cannot read properties of undefined (reading 'message')`

In this pr (https://github.com/bucketeer-io/bucketeer/pull/1684),
removing setting of `auditlog.LocalizedMessage` could cause this error.


cre8ivejp (approved):
Thank you!

PR #1709, opened 2025-04-10
fix: admin audit logs not being acknowledged

cre8ivejp:
The `createAuditLogsMySQL` func is also being used to create admin logs, and it was not acknowledging the audit logs messages correctly.


kakcy (approved):
LGTM 👍 
Thanks you!

PR #1710, opened 2025-04-11
fix: add variationID parameter to newPFMergeKey

nnnkkk7:
### Issue

The user count of day unit causes the results to change each time on the evaluation console.

The `newPFMergeKey` function was generating cache keys based only on feature ID and environment ID, without including the variation ID. This caused all goroutines to operate on the same Redis keys, leading to race conditions where one goroutine would delete or modify keys that other goroutines were still using.

### Solution

We modified the `newPFMergeKey` function to include the variation ID in the generated keys:

```go
func newPFMergeKey(
  kind, featureID, environmentId string,
  variationID string, // Added variation ID parameter
) string {
  return cache.MakeKey(
    fmt.Sprintf("%s:%s", pfMergeKind, kind),
    fmt.Sprintf("%s:%s:%s", pfMergeKey, featureID, variationID), // Include variation ID in key
    environmentId,
  )
}
```

This change ensures that each goroutine operates on separate, unique keys in Redis based on the specific variation ID it's processing, eliminating race conditions and providing consistent results across multiple executions.

### DIFF
--- a/pkg/eventcounter/api/api.go
+++ b/pkg/eventcounter/api/api.go
@@ -333,6 +333,7 @@ func (s *eventCounterService) GetEvaluationTimeseriesCount(
 				userCountKeys,
 				req.FeatureId,
 				req.EnvironmentId,
+				variationID,
 				timestampUnit,
 			)
 			if err != nil {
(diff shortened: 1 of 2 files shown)


cre8ivejp (approved):
Nice catch!
Thank you 🎉

PR #1716, opened 2025-04-15
refactor: change to use tag package storage

kakcy:
This PR removes the code related to tag storage implemented in the feature package, and instead uses the tag storage implemented in the tag package.


hvn2k1 (approved):
The refactor LGTM!

PR #1723, opened 2025-04-17
refactor: change the list retrieval API in the enviroment package to use ListOption

kakcy:
This pull request refactors the query-building logic for listing environments, organizations, and projects in the `EnvironmentService` by introducing a unified `mysql.ListOptions` structure. This change simplifies the codebase, improves maintainability, and enhances test coverage. 

relate of https://github.com/bucketeer-io/bucketeer/issues/1475

### DIFF
--- a/pkg/environment/storage/v2/environment.go
+++ b/pkg/environment/storage/v2/environment.go
@@ -19,7 +19,6 @@ import (
 	"context"
 	_ "embed"
 	"errors"
-	"fmt"
 
 	"github.com/bucketeer-io/bucketeer/pkg/environment/domain"
 	"github.com/bucketeer-io/bucketeer/pkg/storage/v2/mysql"
@@ -49,9 +48,7 @@ type EnvironmentStorage interface {
 	GetEnvironmentV2(ctx context.Context, id string) (*domain.EnvironmentV2, error)
 	ListEnvironmentsV2(
 		ctx context.Context,
-		whereParts []mysql.WherePart,
-		orders []*mysql.Order,
-		limit, offset int,
+		options *mysql.ListOptions,
 	) ([]*proto.EnvironmentV2, int, int64, error)
 }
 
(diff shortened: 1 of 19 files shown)


pkg/environment/storage/v2/environment.go
```diff
@@ -141,18 +139,33 @@ func (s *environmentStorage) GetEnvironmentV2(ctx context.Context, id string) (*
...
+		whereParts := options.CreateWhereParts()
+		whereSQL, whereArgs = mysql.ConstructWhereSQLString(whereParts)
+		orderBySQL := mysql.ConstructOrderBySQLString(options.Orders)
+		limitOffsetSQL := mysql.ConstructLimitOffsetSQLString(options.Limit, options.Offset)
+		query = fmt.Sprintf(selectEnvironmentsSQL, whereSQL, orderBySQL, limitOffsetSQL)
+	} else {
+		query = selectEnvironmentsSQL
+		whereArgs = []interface{}{}
+	}
+
```
nnnkkk7, line 159: I think we can't use the following function because the problem is in the part where `query = baseQuery + whereQuery + orderByQuery + limitOffsetQuery`.
Therefore, what if we modify `ConstructQueryAndWhereArgs` func to use `query = fmt.Sprintf(selectEnvironmentsSQL, whereSQL, orderBySQL, limitOffsetSQL)` and add %s at the end of all raw SQL statements? 

This is just an idea I came up with, so the current implementation is fine too.

nnnkkk7 (commented):
Thank you!
I commented!

pkg/environment/storage/v2/environment.go
kakcy: @nnnkkk7 
Thank you for your good suggestion.
The query defined in `select_environments.sql `is a bit complex, as shown below, and contains a fixed string (`GROUP BY`) between two variable strings.
https://github.com/bucketeer-io/bucketeer/blob/main/pkg/environment/storage/v2/sql/environment/select_environments.sql#L9

I considered changing it to `ConstructQueryAndWhereArgs()`, which takes into account `GROUP BY` and `HAVING` clauses, but since it's not used much in this project at the moment, 
I'd like to leave the refactoring as it is and consider addressing it in a separate PR. What do you think?

nnnkkk7: I see..
I got it!
I agree👌

Thank you!

nnnkkk7 (approved):

PR #1745, opened 2025-05-01
refactor: change the list retrieval API in the flagTrigger package to use ListOption 

kakcy:
This pull request refactors the `ListFlagTriggers` functionality in the `FeatureService` and its associated storage layer to use a new `mysql.ListOptions` struct for query construction. This change simplifies and standardizes the query logic, improves maintainability, and updates corresponding tests and mock implementations.

relate of https://github.com/bucketeer-io/bucketeer/issues/1475

### DIFF
--- a/pkg/feature/api/flag_trigger.go
+++ b/pkg/feature/api/flag_trigger.go
@@ -883,9 +883,17 @@ func (s *FeatureService) ListFlagTriggers(
 		)
 		return nil, err
 	}
-	whereParts := []mysql.WherePart{
-		mysql.NewFilter("feature_id", "=", request.FeatureId),
-		mysql.NewFilter("environment_id", "=", request.EnvironmentId),
+	filters := []*mysql.FilterV2{
+		{
+			Column:   "feature_id",
+			Operator: mysql.OperatorEqual,
+			Value:    request.FeatureId,
+		},
+		{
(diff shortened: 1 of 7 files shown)


pkg/feature/api/flag_trigger.go
```diff
@@ -883,9 +883,17 @@ func (s *FeatureService) ListFlagTriggers(
...
+		{
+			Column:   "feature_id",
+			Operator: mysql.OperatorEqual,
+			Value:    request.FeatureId,
+		},
+		{
+			Column:   "environment_id",
+			Operator: mysql.OperatorEqual,
+			Value:    request.EnvironmentId,
+		},
```
cre8ivejp, line 896: I have a question regarding using a join in the query.
To avoid ambiguity errors, when the table has a common column, we must set the table name before the column. For example, `feature.feature_id`.

For those cases, we just need to do as follows?

```go
{
  Column:   "feature.feature_id",
  Operator: mysql.OperatorEqual,
  Value:    request.FeatureId,
},
```

kakcy: Yes, if you need to resolve ambiguity, you need to name the columns and tables as suggested.

cre8ivejp (approved):
Thank you!

PR #1747, opened 2025-05-01
test: fix flaky list audit logs e2e test

cre8ivejp:
It fails randomly with `upstream request timeout` error.
I'm implementing a retry logic in the test, just like other places, to improve this.


hvn2k1 (approved):
thank you

PR #1748, opened 2025-05-01
test: change log info in the upsert tag e2e test

cre8ivejp:
I want to debug why the test fails here randomly.
The unique key is `UNIQUE KEY `name` (`name`,`environment_id`,`entity_type`)`.
It should be fine, but sometimes it creates a new one.

```sh
    tag_test.go:84: Different create time. Expected: 1746009776, Actual: 1746009782
--- FAIL: TestUpsertAndListTag (5.85s)
```

### DIFF
--- a/test/e2e/tag/tag_test.go
+++ b/test/e2e/tag/tag_test.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"slices"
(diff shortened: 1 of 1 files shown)


kakcy (approved):
Thank you 👍

nnnkkk7 (approved):

PR #1749, opened 2025-05-02
fix(ui/dashboard): flag status showing incorrect

### LINKED ISSUE #1746: fix(ui/dashboard): flag status shows active but the summary shows inactive
I checked the last used info object, and the timestamp is indeed older than 7 days, but on the list page, it shows active.

steveninhle:
I checked the last used info object, and the timestamp is indeed older than 7 days, but on the list page, it shows active.

Fix #1746


cre8ivejp (approved):
Thank you!

PR #1753, opened 2025-05-06
fix: flag update api allowing update fixed and rollout strategy at the same time

cre8ivejp:
Fix https://github.com/bucketeer-io/bucketeer/issues/1754


Ubisoft-potato (approved):
Nice! Thank you!!

Ubisoft-potato (dismissed):

PR #1759, opened 2025-05-07
chore: add new source ids for new sdks

cre8ivejp:


api-description/apidocs.swagger.yaml
```diff
@@ -2176,6 +2176,14 @@ definitions:
       - GO_SERVER
       - NODE_SERVER
       - JAVASCRIPT
+      - FLUTTER
+      - REACT
```
nnnkkk7, line 2180: Are we going to add React client?

cre8ivejp: Yes.

api-description/apidocs.swagger.yaml
```diff
@@ -2176,6 +2176,14 @@ definitions:
...
+      - REACT_NATIVE
+      - OPEN_FEATURE_KOTLIN
+      - OPEN_FEATURE_SWIFT
+      - OPEN_FEATURE_JAVASCRIPT
+      - OPEN_FEATURE_GO
+      - OPEN_FEATURE_NODEJS
```
nnnkkk7, line 2186: If we're using `GO_SERVER` and `NODE_SERVER`, and we have `OPEN_FEATURE_GO`, wouldn't `OPEN_FEATURE_NODE` be more appropriate than OPEN_FEATURE_NODEJS?
WDYT?

cre8ivejp: Indeed.
Done https://github.com/bucketeer-io/bucketeer/pull/1759/commits/562ff4c1972e58e98ba44fec92252239df988a66.

PR #1760, opened 2025-05-07
test: fix flaky e2e audit log

cre8ivejp:


test/e2e/auditlog/audit_log_test.go
```diff
@@ -22,8 +22,7 @@ import (
 )
 
 const (
-	timeout    = 60 * time.Second
```
nnnkkk7, line 25: In L121, we're still using timeout:
`ctx, cancel := context.WithTimeout(context.Background(), timeout)`

test/e2e/auditlog/audit_log_test.go
```diff
@@ -83,7 +80,7 @@ func TestListAndGetAuditLog(t *testing.T) {
 		t.Fatal("GetAuditLog ID error")
 	}
 
-	listAdminResp, err := auditlogClient.ListAdminAuditLogs(ctx, &auditlog.ListAdminAuditLogsRequest{
+	listAdminResp, err := listAdminAuditLogsWithRetry(t, auditlogClient, &auditlog.ListAdminAuditLogsRequest{
```
nnnkkk7, line 83: How about passing `ctx` arg in this func?

cre8ivejp: Using the retry with timeout together limits the retry logic. If the context times out before the retry, it will fail before retrying all.

nnnkkk7: ok!
Thank you!
It'd be fine as it is.

test/e2e/auditlog/audit_log_test.go
cre8ivejp: Fixed https://github.com/bucketeer-io/bucketeer/pull/1760/commits/96b559049f510f7f3c32dc2f9a0616acaf865808.

nnnkkk7 (approved):
LGTM!

PR #1780, opened 2025-05-14
fix(ui/dashboard): showing wrong variations colors in the tooltip

### LINKED ISSUE #1764: fix(ui/dashboard): debugger add flag button not disabling after added all flags
The add flag button is not disabling even after adding all the flags available from the list.

steveninhle:
The add flag button is not disabling even after adding all the flags available from the list.

Fix #1764


cre8ivejp (approved):
Thanks!

PR #1781, opened 2025-05-14
fix: mysql no rows error when using pagination

kakcy:
This pull request addressed the following issue.
- Due to refactoring of ListOption, there was an issue with the query to get the total number, which caused the total number to be retrieved incorrectly.
- There was an issue where paging on the Auditlog list screen, etc. could not be displayed properly.

This pull request refactors multiple `List` methods across various storage modules to improve query construction for counting total records. The changes replace the use of `mysql.ConstructQueryAndWhereArgs` with a more flexible approach that directly appends `WHERE` clauses to the base SQL query when filtering options are provided. If no options are specified, the base query is executed without modification.

### DIFF
--- a/pkg/account/storage/v2/api_key.go
+++ b/pkg/account/storage/v2/api_key.go
@@ -314,7 +314,7 @@ func (s *accountStorage) ListAPIKeys(
 	}
 	nextOffset := offset + len(apiKeys)
 	var totalCount int64
-	countQuery, countWhereArgs := mysql.ConstructQueryAndWhereArgsOnlyUseWherePart(selectAPIKeyV2CountSQLQuery, options)
+	countQuery, countWhereArgs := mysql.ConstructCountQuery(selectAPIKeyV2CountSQLQuery, options)
 	err = s.qe.QueryRowContext(ctx, countQuery, countWhereArgs...).Scan(&totalCount)
 	if err != nil {
 		return nil, 0, 0, err
--- a/pkg/account/storage/v2/api_key.go
+++ b/pkg/account/storage/v2/api_key.go
@@ -314,12 +314,8 @@ func (s *accountStorage) ListAPIKeys(
 	}
 	nextOffset := offset + len(apiKeys)
 	var totalCount int64
-	if options != nil {
-		whereQuery, countWhereArgs := mysql.ConstructWhereSQLString(options.CreateWhereParts())
-		err = s.qe.QueryRowContext(ctx, selectAPIKeyV2CountSQLQuery+whereQuery, countWhereArgs...).Scan(&totalCount)
-	} else {
-		err = s.qe.QueryRowContext(ctx, selectAPIKeyV2CountSQLQuery).Scan(&totalCount)
-	}
+	countQuery, countWhereArgs := mysql.ConstructQueryAndWhereArgsOnlyUseWherePart(selectAPIKeyV2CountSQLQuery, options)
+	err = s.qe.QueryRowContext(ctx, countQuery, countWhereArgs...).Scan(&totalCount)
 	if err != nil {
 		return nil, 0, 0, err
 	}
--- a/pkg/account/storage/v2/api_key.go
+++ b/pkg/account/storage/v2/api_key.go
@@ -314,8 +314,12 @@ func (s *accountStorage) ListAPIKeys(
 	}
 	nextOffset := offset + len(apiKeys)
 	var totalCount int64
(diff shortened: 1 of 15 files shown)


pkg/account/storage/v2/api_key.go
```diff
@@ -314,8 +314,12 @@ func (s *accountStorage) ListAPIKeys(
...
 	nextOffset := offset + len(apiKeys)
 	var totalCount int64
-	countQuery, countWhereArgs := mysql.ConstructQueryAndWhereArgs(selectAPIKeyV2CountSQLQuery, options)
-	err = s.qe.QueryRowContext(ctx, countQuery, countWhereArgs...).Scan(&totalCount)
+	if options != nil {
+		whereQuery, countWhereArgs := mysql.ConstructWhereSQLString(options.CreateWhereParts())
+		err = s.qe.QueryRowContext(ctx, selectAPIKeyV2CountSQLQuery+whereQuery, countWhereArgs...).Scan(&totalCount)
+	} else {
+		err = s.qe.QueryRowContext(ctx, selectAPIKeyV2CountSQLQuery).Scan(&totalCount)
+	}
```
cre8ivejp, line 322: WDYT of creating a `ConstructWhereSQLString` helper in the query.go to reduce duplication logic?

```suggestion
	countQuery, countArgs := mysql.ConstructCountQuery(countAccountsV2SQL, options)
	err = s.qe.QueryRowContext(ctx, countQuery, countArgs...).Scan(&totalCount)
```
---

`mysql/query.go`
```go
// ConstructCountQuery builds a count query with optional filtering.
// It takes a base count SQL query and ListOptions, and returns the complete query and its arguments.
// The base SQL should be a valid COUNT query without WHERE clause.
func ConstructCountQuery(baseSQL string, options *ListOptions) (query string, args []interface{}) {
	if options != nil {
		whereQuery, countWhereArgs := ConstructWhereSQLString(options.CreateWhereParts())
		return baseSQL + whereQuery, countWhereArgs
	}
	return baseSQL, nil
}
```

kakcy: Added `ConstructQueryAndWhereArgsOnlyUseWherePart()` and consolidated common code.
https://github.com/bucketeer-io/bucketeer/pull/1781/commits/0a94775d255f55041e38165421777980d3c08003

pkg/storage/v2/mysql/query.go
```diff
@@ -360,6 +360,21 @@ func ConstructQueryAndWhereArgs(baseQuery string, options *ListOptions) (query s
 	return
 }
 
+// ConstructQueryAndWhereArgsOnlyUseWherePart builds a query with only the WHERE part.
+// This Function is intended to be used when the purpose is
+// to generate a query that does not require any conditions other than the Where clause,
+// such as "SELECT COUNT(1) FROM XXX".
+func ConstructQueryAndWhereArgsOnlyUseWherePart(
```
cre8ivejp, line 367: The name ConstructQueryAndWhereArgsOnlyUseWherePart is too verbose and not very clear.
WDYT?

```suggestion
// ConstructCountQuery builds a count query with optional filtering.
// It takes a base count SQL query and ListOptions, and returns the complete query and its arguments.
// The base SQL should be a valid COUNT query (e.g., "SELECT COUNT(1) FROM table").
// Additional WHERE clauses from ListOptions will be appended to the base query.
func ConstructCountQuery(
```

kakcy: @cre8ivejp 
There was a concern that `ConstructCountQuery()` would give the impression that `COUNT(1)` was added to the query within the function,
so we gave it this name to distinguish it from the original `ConstructQueryAndWhereArgs().`
https://github.com/bucketeer-io/bucketeer/blob/main/pkg/storage/v2/mysql/query.go#L348

This name is certainly long, so how about setting a flag to` ConstructQueryAndWhereArgs()` to branch the processing as shown below?

```
func ConstructQueryAndWhereArgs(baseQuery string, options *ListOptions, bool isOnlyUseWhereParts) (query string, whereArgs []interface{})
```

kakcy: @cre8ivejp 
If I think about it, this function constructs a query based on ListOption for the base query, so I think `ConstructsCountQuery` would be fine.
I'm going to fix that.

kakcy: I changed to `ConstructCountQuery`.
Please check again.

cre8ivejp (approved):
Nice work!

PR #1792, opened 2025-05-20
feat: grpc gateway implementation

Ubisoft-potato:


pkg/rpc/gateway/gateway.go
```diff
@@ -0,0 +1,157 @@
...
+
+	// Create and start the HTTP server
+	g.httpServer = &http.Server{
+		Addr:    g.restAddr,
+		Handler: mux,
+	}
+
+	// Start the server in a goroutine
+	errChan := make(chan error, 1)
+	go func() {
```
hvn2k1, line 121: So in here,  we don't create a new pod, but run it in another goroutine in the same backend server?

nnnkkk7 (approved):
LGTM！！

cre8ivejp (dismissed):
Nice work!
Thank you!

Issue #1803, opened 2025-05-30
chore: add custom JSON unmarshaler for boolean string compatibility


cre8ivejp:
Add a temporary custom JSON unmarshaler to handle boolean fields sent as
strings from the Android SDK. This addresses a compatibility issue that
arose after switching from Envoy JSON transcoder to gRPC-Gateway.

PR #1804, opened 2025-05-30
chore: add custom JSON unmarshaler for boolean string compatibility

cre8ivejp:
Fix https://github.com/bucketeer-io/bucketeer/issues/1803

Related to https://github.com/bucketeer-io/android-client-sdk/pull/230

Add a temporary custom JSON unmarshaler to handle boolean fields sent as strings from the Android SDK. This addresses a compatibility issue that arose after switching from Envoy JSON transcoder to gRPC-Gateway.

Changes:
- Implement customJSONPb struct with custom Unmarshal method
- Support multiple boolean string representations ("true", "false", "1", "0", etc.)
- Handle nested JSON structures with recursive boolean string conversion
- Preserve original errors for invalid inputs
- Add comprehensive test coverage for all scenarios

The unmarshaler follows this logic:
1. First attempts at standard protobuf unmarshaling
2. If that fails, checks for boolean string representations
3. Recursively converts boolean strings in nested objects and arrays
4. Returns the original error for non-boolean invalid strings

Supported boolean string formats:
- True: "true", "True", "TRUE", "1"
- False: "false", "False", "FALSE", "0"

This is a temporary solution to maintain backward compatibility while the Android SDK is updated to send proper boolean types instead of strings.

### DIFF
--- a/pkg/rpc/gateway/gateway.go
+++ b/pkg/rpc/gateway/gateway.go
@@ -157,7 +157,7 @@ func (m *BooleanConversionMarshaler) convertStringBooleansRecursive(data interfa
 
 // stringToBool converts string boolean values to actual booleans
 func stringToBool(s string) (bool, bool) {
-	if val, ok := stringBoolMap[strings.ToLower(s)]; ok {
+	if val, ok := stringBoolMap[strings.ToLower(strings.TrimSpace(s))]; ok {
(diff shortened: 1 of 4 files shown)


Ubisoft-potato (approved):
Nice work, LGTM! 🚀

PR #1826, opened 2025-06-06
fix: slack notification not being in some cases

cre8ivejp:
When there are multiple subscriptions, and one of them fails due to a flag not found error, it returns an error, ignoring the other subscriptions.

### DIFF
--- a/pkg/notification/sender/sender.go
+++ b/pkg/notification/sender/sender.go
@@ -17,7 +17,6 @@ package sender
 
 import (
 	"context"
-	"errors"
 
 	"go.uber.org/zap"
 	"google.golang.org/protobuf/encoding/protojson"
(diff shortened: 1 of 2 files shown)


pkg/notification/sender/sender.go
```diff
@@ -118,14 +114,11 @@ func (s *sender) Send(ctx context.Context, notificationEvent *senderproto.Notifi
...
 		// When a flag changes it must be checked before sending notifications
-		send, err := s.checkForFeatureDomainEvent(
+		send := s.checkForFeatureDomainEvent(
 			subscription,
 			notificationEvent.SourceType,
 			notificationEvent.Notification.DomainEventNotification.EntityData,
 		)
-		if err != nil {
-			return err
-		}
```
cre8ivejp, line 128: The issue happened here.

nnnkkk7 (approved):
LGTM!
Thank you!

PR #1831, opened 2025-06-06
chore: remove error log when failing to read the request body

cre8ivejp:
When the grpc-gateway reads the request body, if the user disconnects, it will print error logs.
Since that kind of error is too common, I removed the logging until we implement metrics instead.
This way we won't pollute the logs or incur increased costs.

```
error: "stream error: stream ID 31057; NO_ERROR"
```


Ubisoft-potato (approved):
Nice! 👍
Thanks!

PR #1834, opened 2025-06-09
test: fix list audit log flaky e2e test

cre8ivejp:
The dev environment has more than 1.3 million entries, so when using ordering desc, it tries to read the entire table, causing timeout issues.
So, I have added the `from` and `to` to limit the data, increasing performance and fixing the flaky test.
Also, I have changed to use a fresh context for each request because if we reused the same context, a timeout from the first attempt would affect all subsequent retries.


Ubisoft-potato (approved):
LGTM! Thank you!

PR #1835, opened 2025-06-09
chore: remove proto descriptor

cre8ivejp:
Because we use grpc-gateway to transcode the JSON requests, we don't need the proto_descriptor anymore that was being used in the envoy.


Ubisoft-potato (approved):
Great! We don't need pb descriptor anymore!

PR #1812, opened 2025-06-03
refactor: change ListTags to utilize ListOptions

kakcy:
This pull request refactors the `ListTags` method across multiple services to use a new `mysql.ListOptions` structure instead of separate parameters for filters, orders, limits, and offsets. This change simplifies the codebase and improves maintainability. Corresponding updates were made to tests and mock implementations to align with the new method signature.

relate of https://github.com/bucketeer-io/bucketeer/issues/1475

PR #1839, opened 2025-06-10
fix: add EntityType value when getting tag list

kakcy:
When addressing #1812, the code to specify the EntityType value was missing.


cre8ivejp (approved):
👍

Issue #1840, opened 2025-06-10
feat: return and filter objects in organization settings by environment for member users


hvn2k1:
### Describe the feature you'd like

In new console, we have the organization setting console, for system admin, or organization admin, it's ok to display everything.

## 1. For member that have access to only some environments in the organization, only return objects inside those environments. This apply for: 
  - [x] Member or account
  - [x] API key
  - [x] Slack
  - [x] FCM or Push notification
  - [ ] Project
## 2. Support filter by environment Id for objects in the organization settings. This apply for:
  - [x] Member or account
  - [x] API key
  - [x] Slack
  - [x] FCM or Push notification

PR #1847, opened 2025-06-12
feat: only return objects in the environment for organization member

hvn2k1:
To resolve https://github.com/bucketeer-io/bucketeer/issues/1840
- First we check if the account role is admin, by adding environment roles and organization role inside 
- If admin => allow to view any resource in the organization
- If member => get environment that account have access. In the next step, for each API we use different filter

- List Accounts:
  - Use multiple JSON_CONTAINS query to filter account in the accessible environments:
```sql
JSON_CONTAINS(environment_roles, '{"environment_id": "env1", "role": 1}') OR 
JSON_CONTAINS(environment_roles, '{"environment_id": "env2", "role": 1}') OR 
JSON_CONTAINS(environment_roles, '{"environment_id": "env3", "role": 2}')
```

- List Pushes (FCM):
  - Add environmentIDs in request, check organizationRole for new console
  - Use IN filter

- List API Keys:
  - Add environmentIDs in request, check organizationRole for new console
  - Use IN filter

- List Subscriptions (Slack):
  - Use IN filter


pkg/account/api/account.go
```diff
@@ -1129,6 +1129,16 @@ func (s *AccountService) ListAccountsV2(
 	if err != nil {
 		return nil, err
 	}
+
+	// If not organization admin, user can only view accounts in their environments
```
cre8ivejp, line 1133:
```suggestion
	// If not an organization admin, a user can only view accounts in their environments
```

pkg/account/api/account.go
```diff
@@ -1247,6 +1286,37 @@ func (s *AccountService) ListAccountsV2(
 	}, nil
 }
 
+func (s *AccountService) constructEnvironmentRoles(
+	req *accountproto.ListAccountsV2Request,
+	editor *eventproto.Editor,
+) ([]*accountproto.AccountV2_EnvironmentRole, error) {
+	requestEnvironmentRoles := make([]*accountproto.AccountV2_EnvironmentRole, 0)
+	// no allowed roles means user has no access to any environment in the organization
```
cre8ivejp, line 1294:
```suggestion
	// No allowed roles means the user has no access to any environment in the organization
```

pkg/account/api/account.go
```diff
@@ -1247,6 +1286,37 @@ func (s *AccountService) ListAccountsV2(
...
+
+	if req.EnvironmentId != nil && req.EnvironmentRole != nil {
+		for _, role := range editor.EnvironmentRoles {
+			if role.EnvironmentId == req.EnvironmentId.Value &&
+				role.Role == accountproto.AccountV2_Role_Environment(req.EnvironmentRole.Value) {
+				requestEnvironmentRoles = append(requestEnvironmentRoles, role)
+				break
+			}
+		}
+	} else if req.EnvironmentId != nil && req.EnvironmentRole == nil {
```
cre8ivejp, line 1307: In which case, the environment is not nil, but the environment role is?

proto/event/domain/event.proto
```diff
@@ -289,6 +289,10 @@ message Editor {
   PublicAPIEditor public_api_editor = 5;
   bytes avatar_image = 6;       // this is not save in auditlog table
   string avatar_file_type = 7;  // this is not save in auditlog table
+  repeated bucketeer.account.AccountV2.EnvironmentRole environment_roles =
+      8;  // this is not saved in auditlog table
```
cre8ivejp, line 293:
```suggestion
      8;  // this is not saved in the auditlog table
```

proto/event/domain/event.proto
```diff
@@ -289,6 +289,10 @@ message Editor {
...
+  bucketeer.account.AccountV2.Role.Organization organization_role =
+      9;  // this is not saved in auditlog table
```
cre8ivejp, line 295:
```suggestion
      9;  // this is not saved in the auditlog table
```

pkg/account/api/account.go
```diff
@@ -1164,26 +1174,54 @@ func (s *AccountService) ListAccountsV2(
...
+			} else if r.Role != accountproto.AccountV2_Role_Environment_UNASSIGNED {
+				values[0] = fmt.Sprintf("{\"role\": %d}", r.Role)
+			}
+			if values[0] != nil && values[0] != "" {
+				rawEnvQuery += fmt.Sprintf("JSON_CONTAINS(environment_roles, '%s')", values[0])
+			}
+			if i < len(requestEnvironmentRoles)-1 {
+				rawEnvQuery += " OR "
+			}
+		}
```
cre8ivejp, line 1213: @hvn2k1 cc: @kakcy 
How about creating a new OR filter instead of creating a raw query?
A raw query should be the last approach if there is no other way.


E.g.
```go
type OrFilter struct {
    Filters []WherePart
}

func NewOrFilter(filters ...WherePart) WherePart {
    return &OrFilter{
        Filters: filters,
    }
}

func (f *OrFilter) SQLString() (sql string, args []interface{}) {
    if len(f.Filters) == 0 {
        return "", nil
    }
    
    var sb strings.Builder
    sb.WriteString("(")
    for i, filter := range f.Filters {
        if i > 0 {
            sb.WriteString(" OR ")
        }
        filterSQL, filterArgs := filter.SQLString()
        sb.WriteString(filterSQL)
        args = append(args, filterArgs...)
    }
    sb.WriteString(")")
    return sb.String(), args
}
```

Example
```go
if len(requestEnvironmentRoles) == 0 {
    // ... existing code for admin case ...
} else {
    var orFilters []WherePart
    for _, r := range requestEnvironmentRoles {
        values := make([]interface{}, 1)
        if r.EnvironmentId != "" && r.Role != accountproto.AccountV2_Role_Environment_UNASSIGNED {
            values[0] = fmt.Sprintf("{\"environment_id\": \"%s\", \"role\": %d}", r.EnvironmentId, r.Role)
        } else if r.EnvironmentId != "" {
            values[0] = fmt.Sprintf("{\"environment_id\": \"%s\"}", r.EnvironmentId)
        } else if r.Role != accountproto.AccountV2_Role_Environment_UNASSIGNED {
            values[0] = fmt.Sprintf("{\"role\": %d}", r.Role)
        }
        if values[0] != nil && values[0] != "" {
            orFilters = append(orFilters, &JSONFilter{
                Column: "environment_roles",
                Func:   mysql.JSONContainsJSON,
                Values: values,
            })
        }
    }
    if len(orFilters) > 0 {
        jsonFilters = append(jsonFilters, NewOrFilter(orFilters...))
    }
}
```

pkg/account/api/account.go
```diff
@@ -1164,26 +1174,54 @@ func (s *AccountService) ListAccountsV2(
...
+	var rawEnvQuery string
+	if len(requestEnvironmentRoles) == 0 {
+		values := make([]interface{}, 1)
+		if req.EnvironmentId != nil && req.EnvironmentRole != nil {
+			values[0] = fmt.Sprintf("{\"environment_id\": \"%s\", \"role\": %d}", req.EnvironmentId.Value, req.EnvironmentRole.Value) // nolint:lll
+		} else if req.EnvironmentId != nil {
+			values[0] = fmt.Sprintf("{\"environment_id\": \"%s\"}", req.EnvironmentId.Value)
+		} else if req.EnvironmentRole != nil {
+			values[0] = fmt.Sprintf("{\"role\": %d}", req.EnvironmentRole.Value)
+		}
```
cre8ivejp, line 1187: We could improve the code readability of the duplicate code for admin and non-admin users.

### createEnvironmentRoleJSON
I prefer using `json.Marshal` to properly serialize a struct, making it type-safe and less prone to errors instead of using `fmt.Sprintf`

E.g.
```go
type EnvironmentRole struct {
    EnvironmentID string `json:"environment_id"`
    Role         int32  `json:"role"`
}

func createEnvironmentRoleJSON(envID string, role accountproto.AccountV2_Role_Environment) (string, bool) {
    if envID == "" && role == accountproto.AccountV2_Role_Environment_UNASSIGNED {
        return "", false
    }
    
    er := EnvironmentRole{}
    if envID != "" {
        er.EnvironmentID = envID
    }
    if role != accountproto.AccountV2_Role_Environment_UNASSIGNED {
        er.Role = int32(role)
    }
    
    if er.EnvironmentID == "" && er.Role == 0 {
        return "", false
    }
    
    jsonBytes, err := json.Marshal(er)
    if err != nil {
        return "", false
    }
    return string(jsonBytes), true
}

func (s *AccountService) createEnvironmentRoleFilter(envID string, role accountproto.AccountV2_Role_Environment) (*mysql.JSONFilter, error) {
    jsonStr, ok := createEnvironmentRoleJSON(envID, role)
    if !ok {
        return nil, nil
    }
    
    return &mysql.JSONFilter{
        Column: "environment_roles",
        Func:   mysql.JSONContainsJSON,
        Values: []interface{}{jsonStr},
    }, nil
}

```

### Example
```go
    var jsonFilters []*mysql.JSONFilter
    
    // Handle environment role filters
    if len(requestEnvironmentRoles) == 0 {
        // Admin case
        if req.EnvironmentId != nil || req.EnvironmentRole != nil {
            filter, err := s.createEnvironmentRoleFilter(
                req.EnvironmentId.Value,
                accountproto.AccountV2_Role_Environment(req.EnvironmentRole.Value),
            )
            if err != nil {
                return nil, err
            }
            if filter != nil {
                jsonFilters = append(jsonFilters, filter)
            }
        }
    } else {
        // Non-admin case
        var orFilters []mysql.WherePart
        for _, r := range requestEnvironmentRoles {
            filter, err := s.createEnvironmentRoleFilter(r.EnvironmentId, r.Role)
            if err != nil {
                return nil, err
            }
            if filter != nil {
                orFilters = append(orFilters, filter)
            }
        }
        if len(orFilters) > 0 {
            jsonFilters = append(jsonFilters, mysql.NewOrFilter(orFilters...))
        }
    }
```

pkg/account/api/account.go
hvn2k1: This is from request, maybe to filter users with some specific role like member or admin, it's already in the proto request: https://github.com/bucketeer-io/bucketeer/blob/main/proto/account/service.proto#L161-L188

pkg/account/api/account.go
hvn2k1: I've applied struct marshal for this

hvn2k1:
rename branch https://github.com/bucketeer-io/bucketeer/pull/1852

PR #1852, opened 2025-06-16
feat: only return objects in the environment for organization member

hvn2k1:
To resolve https://github.com/bucketeer-io/bucketeer/issues/1840
- First we check if the account role is admin, by adding environment roles and organization role inside 
- If admin => allow to view any resource in the organization
- If member => get environment that account have access. In the next step, for each API we use different filter

- List Accounts:
  - Use multiple JSON_CONTAINS query to filter account in the accessible environments:
```sql
JSON_CONTAINS(environment_roles, '{"environment_id": "env1", "role": 1}') OR 
JSON_CONTAINS(environment_roles, '{"environment_id": "env2", "role": 1}') OR 
JSON_CONTAINS(environment_roles, '{"environment_id": "env3", "role": 2}')
```

- List Pushes (FCM):
  - Add environmentIDs in request, check organizationRole for new console
  - Use IN filter

- List API Keys:
  - Add environmentIDs in request, check organizationRole for new console
  - Use IN filter

- List Subscriptions (Slack):
  - Use IN filter


cre8ivejp (approved):
Nice work!

Issue #1798, opened 2025-05-22
feat: view feature flags associated with a rule in the console


kakcy:
### Describe the feature you'd like

Feature flags associated with prerequisites are displayed in the Targets tab along with a message, or you can use filters to find feature flags that have prerequisites.

It would be nice to be able to do something similar with feature flags associated with rules.

### Background

Currently, there is no way in the console to see which feature flags are associated with a rule.Currently, there is no way in the console to see which feature flags are associated with a rule.
When there are a large number of feature flags, finding the associated feature flag among them can be very difficult.

PR #1854, opened 2025-06-17
feat: support filter features has rollout rule depend on another flags

hvn2k1:
To resolve https://github.com/bucketeer-io/bucketeer/issues/1798

Use below condition to check if feature has rollout rule depend on another flags:

```mysql
JSON_CONTAINS(JSON_EXTRACT(rules,` '$[*].clauses[*].operator'), '11') = true
```


cre8ivejp (approved):
Thank you!

Issue #1856, opened 2025-06-18
fix: cannot create a fcm push notification when there are no flag tags


cre8ivejp:
Since the flag tags are not required, we must also make this optional when creating pushes.

PR #1857, opened 2025-06-18
fix: cannot create a fcm push notification when there are no flag tags

cre8ivejp:
Fix https://github.com/bucketeer-io/bucketeer/issues/1856


Ubisoft-potato (approved):
LGTM! 👍

PR #1860, opened 2025-06-18
fix: cannot delete all push tags

cre8ivejp:
Fix https://github.com/bucketeer-io/bucketeer/issues/1858


pkg/push/domain/push.go
```diff
@@ -157,6 +165,33 @@ func (p *Push) ExistTag(findTag string) bool {
...
 
+func (p *Push) AddTag(tag string) error {
+	if contains(tag, p.Tags) {
+		return nil
+	}
+	p.Tags = append(p.Tags, tag)
+	return nil
+}
+
+func (p *Push) RemoveTag(tag string) error {
```
cre8ivejp, line 176: To avoid compatibility issues with the current add and delete tag functions, I have added these new ones.
I will delete the old functions being used by the current console when the new one is released.

Ubisoft-potato (commented):
LGTM, Thanks! 😁

PR #1863, opened 2025-06-19
test: fix unknown field HasFeatureRule in the feature e2e test

hvn2k1:


cre8ivejp (approved):
👍 
nit: test: fix unknown field HasFeatureRule in the feature e2e test

Issue #1859, opened 2025-06-18
feat: create teams entity


cre8ivejp:
On the new console in the Members page, we have tags to be used as groups. It uses tags from the tag table.
The issue we have is that the tags are at the environment, not the organization level, and when creating a member, the admin is allowed to create in multiple environments.

In the current console, the members are at the environment level, not the organization, so it works fine.
So, when showing the tags, we don't know what environment we should list, and also, there could be the same tag in multiple environments.
To fix this, we should implement a new Teams entity to fix this properly.

PR #1867, opened 2025-06-19
feat: init proto and database for team entity

hvn2k1:
to resolve https://github.com/bucketeer-io/bucketeer/issues/1859


proto/account/service.proto
```diff
@@ -134,6 +135,7 @@ message UpdateAccountV2Request {
   google.protobuf.BoolValue disabled = 21;
   common.StringListValue tags = 22;
   ChangeAccountV2TagsCommand change_tags_command = 23 [deprecated = true];
+  common.StringListValue teams = 24;
```
cre8ivejp, line 138: We will need the granular changes implementation when updating.

Reference:
https://github.com/bucketeer-io/bucketeer/pull/1860/files#diff-aaaa1f00bfcb30d3f25e7c80acd9f1dcbf2cf186aae25f070880b0c1b0c4fa61R28-R39

hvn2k1: I've apply new granular pattern for this

PR #1878, opened 2025-06-21
feat(ui/dashboard): restrict the environment list when creating fcm and slack notifications

### LINKED ISSUE #1870: chore(ui/dashboard): restrict the environment list when creating fcm and slack notifications
For the FCM and Slack Create pages, I need to change the implementation of the environment list.
We must check the environment_roles from the local storage (retrieved by the get_me API), and create a list of environments where the user has the Editor role instead of calling the list_environments API, which includes all environments even if the user doesn't belong to them.

Currently, if the user tries to create using an environment he doesn't belong to, the request will fail.

steveninhle:
For the FCM and Slack Create pages, I need to change the implementation of the environment list.
We must check the environment_roles from the local storage (retrieved by the get_me API), and create a list of environments where the user has the Editor role instead of calling the list_environments API, which includes all environments even if the user doesn't belong to them.

Currently, if the user tries to create using an environment he doesn't belong to, the request will fail.

Fix #1870

### DIFF
--- a/ui/dashboard/src/@api/notification/notifications-fetcher.ts
+++ b/ui/dashboard/src/@api/notification/notifications-fetcher.ts
@@ -6,6 +6,7 @@ import { stringifyParams } from 'utils/search-params';
 
 export interface NotificationsFetcherParams extends CollectionParams {
   environmentId?: string;
+  environmentIds?: string[];
   organizationId: string;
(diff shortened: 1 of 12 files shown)


cre8ivejp (approved):
Thank you!

PR #1896, opened 2025-06-26
chore: improve debug log when validating events from the client

cre8ivejp:
When the validation fails, we don't have any information about which SDK the request came from.
So, I implemented logging more information in the event for better debugging.


Ubisoft-potato (approved):
Nice improvement!!!

PR #1942, opened 2025-07-04
fix: cannot update an existing member on the new console

cre8ivejp:
Fix https://github.com/bucketeer-io/bucketeer/issues/1943

Due to schema changes for the new console, the old members in the account table don't have the first and last names, which are being validated in the new API for the new console.
So, I changed the validation to allow the update if the `name` field is not empty.


Ubisoft-potato (approved):
LGTM!!n

PR #1944, opened 2025-07-05
feat(ui/dashboard): update translation for api keys role and invalidate cache for histories and segments.

steveninhle:


cre8ivejp (approved):
Thank you!

Issue #1947, opened 2025-07-07
fix: incorrect goal in used status


hvn2k1:
### Version

latest

### Describe the bug

When returning goals, there are goals not used by any experiment or operation, but in used status is true

### Relevant log output

```bash

```

PR #1948, opened 2025-07-07
fix: incorrect goal's in used status

hvn2k1:
Use JSON_CONTAINS in stead of LIKE in where condition to correctly filter in used goals
to fix https://github.com/bucketeer-io/bucketeer/issues/1947


cre8ivejp (approved):
Thank you!

PR #1955, opened 2025-07-07
fix(ui/dashboard): disabled archive option for current environment and invalidate environment list

steveninhle:


cre8ivejp (approved):
Thank you!

PR #1971, opened 2025-07-10
chore: add metrics registerer to event writers and persisters

nnnkkk7:
We can't collect BigQuery metrics now, so added metrics registerer to event writers and persisters.


cre8ivejp (approved):
👍

PR #1975, opened 2025-07-10
fix: flag only updating the updated_at field

cre8ivejp:
When the console sends in the flag update API, the `enabled` and `archived` fields, the new update logic still uses old functions from the old UI, which updates the `updated_at` when setting those fields. So, even if the boolean value is the same, because the `updated_at` was updated, MySQL saves with no issues, causing an incorrect state in the audit logs with only the `updated_at`.

If the console tries to send those two fields without changing the current state, it will return the following error, so we can know there is something wrong next time.

```sh
{"code":2, "message":"feature: unexpected affected rows", "details":[]}
```

### Problem
- Shallow Copy Bug: `copier.Copy` caused modifications to the updated feature to also modify the original feature, making copier.Copy always sees them as identical. Preventing version increments even when changes occurred
- Unnecessary Updates: Identical values (e.g., enabled: true when already enabled) incorrectly incremented version/timestamp
- Inconsistent Serialization: Empty slices showed as null vs [] between API responses and audit logs

### Solution
- Use copier.CopyWithOption(..., DeepCopy: true) for proper feature isolation
- Use proto.Equal() to detect actual changes and only update version/timestamp when needed
- Standardize empty slices as [] for consistent JSON serialization
- Fix audit log events to use updated.Version instead of feature.Version

### DIFF
--- a/pkg/feature/domain/feature_update.go
+++ b/pkg/feature/domain/feature_update.go
@@ -15,6 +15,8 @@
 package domain
 
 import (
+	"errors"
+	"slices"
 	"time"
 
 	"github.com/jinzhu/copier"
@@ -22,6 +24,10 @@ import (
 
 	"github.com/bucketeer-io/bucketeer/proto/common"
 	"github.com/bucketeer-io/bucketeer/proto/feature"
+)
+
+var (
+	ErrTagNotFound = errors.New("feature: tag not found")
 )
 
 // Update returns a new Feature with the updated values.
@@ -203,16 +209,16 @@ func (f *Feature) applyGeneralUpdates(
 	}
 	if enabled != nil {
 		if enabled.Value {
-			_ = f.Enable()
+			_ = f.updateEnable()
 		} else {
-			_ = f.Disable()
+			_ = f.updateDisable()
 		}
 	}
 	if archived != nil {
 		if archived.Value {
-			_ = f.Archive()
+			_ = f.updateArchive()
 		} else {
-			_ = f.Unarchive()
+			_ = f.updateUnarchive()
 		}
 	}
 	if defaultStrategy != nil {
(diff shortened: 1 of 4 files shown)


nnnkkk7 (approved):
Nice!

pkg/feature/domain/feature_update.go
```diff
@@ -15,15 +15,22 @@
...
 	"github.com/jinzhu/copier"
+	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/known/wrapperspb"
 
 	"github.com/bucketeer-io/bucketeer/proto/common"
 	"github.com/bucketeer-io/bucketeer/proto/feature"
 )
 
+var (
+	ErrTagNotFound = errors.New("feature: tag not found")
```
nnnkkk7, line 31: We don't use `ErrTagNotFound` now for error catching, so we can just return `errors.New("feature: tag not found")` .

pkg/feature/domain/feature_update.go
```diff
@@ -339,188 +350,282 @@ func (f *Feature) applyGranularChanges(
...
-		tagMap[tag] = struct{}{}
-	}
-	for _, tag := range other.Tags {
-		if _, exists := tagMap[tag]; !exists {
-			return true
-		}
-	}
+// New functions for Update method that handle timestamp updates intelligently
+// These functions only update the timestamp when actual changes occur
+// TODO: Remove these duplicate functions once old console is deprecated
```
nnnkkk7, line 355: It would be better if you could indicate to what extent this is a duplicate set of functions.

nnnkkk7 (commented):
Thanks!

hvn2k1 (dismissed):
The PR looks good to me!

pkg/feature/domain/feature_update.go
cre8ivejp: Done 08a9ab848441e04879ead5cfc3c62c04fc0444a1

pkg/feature/domain/feature_update.go
cre8ivejp: I have removed the comments because the functions that will be removed are in the `feature.go` file being used by the command logic.

PR #1978, opened 2025-07-11
fix: incorrect audit log state when creating or deleting entries

cre8ivejp:
Fix https://github.com/bucketeer-io/bucketeer/issues/1977

In many places, when creating entries, they were using different objects for the previous entry, causing other states when showing the audit logs on the console.

Also, I fixed that when deleting an entry, it was showing as an update.
I have fixed the console to support both cases for the data already saved in the DB for compatibility.


pkg/tag/api/api.go
```diff
@@ -131,36 +132,63 @@ func (s *TagService) CreateTag(
...
+			previousEntityData = existingTag
+			eventType = eventproto.Event_TAG_UPDATED
+			eventData = &eventproto.TagUpdatedEvent{
+				Id:            actualTag.Id,
+				Name:          actualTag.Name,
+				UpdatedAt:     actualTag.UpdatedAt,
+				EntityType:    actualTag.EntityType,
+				EnvironmentId: actualTag.EnvironmentId,
+			}
+		}
```
cre8ivejp, line 177: I found another bug reporting as CREATED even if we are updating.

hvn2k1 (approved):
thank you

PR #1995, opened 2025-07-15
chore: improve error log when saving audit logs

cre8ivejp:
I'm investigating the following error when saving the audit log.

```sh
ERROR 2025-07-15T02:18:13.876800485Z [resource.labels.containerName: subscriber] Failed to put audit logs
"error": "Error 1406 (22001): Data too long for column 'entity_data' at row 1"
```

### Resolution

The cause was that the user saved a very large list of user IDs in the individual targeting. So, when saving the entire feature object in the entity_data column exceeded the max limit of 64 KB.
I have asked the maintainer to use the user segment instead, and I have purged the message from PubSub since we are not going to support an entity larger than 64 KB at the moment.


Ubisoft-potato (approved):
LGTM!!

PR #1996, opened 2025-07-15
feat(ui/dashboard): save the data when the user is typing into browser and update the search placeholder

steveninhle:


cre8ivejp (approved):
Thank you!

PR #1998, opened 2025-07-15
chore: add interface for demo creation token

cre8ivejp:
Because the Demo organization is created before the user logs in, we need a different token to verify the request.
This will be used in the Environment service when adding the API.


hvn2k1 (approved):
Thank you! LGTM

Issue #1796, opened 2025-05-22
feat: auto create new organization and data retention in demo site


hvn2k1:
### Describe the feature you'd like

- When login, a new organization will be created for user
- Data will be deleted after a period of time, let say X days (with notice when user in the demo environment)

hvn2k1:
the retention will be split into different issue

PR #2004, opened 2025-07-16
feat: implement exchange demo token API for demo site

hvn2k1:
Part of https://github.com/bucketeer-io/bucketeer/issues/1796

This API create a different type of token to create demo organization in demo site


pkg/environment/api/api.go
```diff
@@ -73,22 +110,298 @@ func NewEnvironmentService(
...
+	// Create refresh token
+	refreshTokenTTL := 5 * time.Minute
+	if s.opts.refreshTokenTTL > 0 {
+		refreshTokenTTL = s.opts.refreshTokenTTL
+	}
+	refreshToken := &token.DemoCreationToken{
+		Email:    userEmail,
+		Expiry:   timeNow.Add(refreshTokenTTL),
+		IssuedAt: timeNow,
+	}
```
cre8ivejp, line 298: The demo creation token is not an access token, so it doesn't need a refresh token because it is a one-time token.

pkg/environment/api/api.go
```diff
@@ -73,22 +110,298 @@ func NewEnvironmentService(
...
+		}
+		return nil, dt.Err()
+	}
+
+	return &authproto.Token{
+		AccessToken:  signedAccessToken,
+		RefreshToken: signedRefreshToken,
+		TokenType:    "Bearer",
+		Expiry:       accessTokenTTL.Unix(),
+	}, nil
```
cre8ivejp, line 321: We should have a DemoCreationToken instead and pass only the necessary information. Which will be the token and the token type?

pkg/environment/api/api.go
```diff
@@ -73,22 +110,298 @@ func NewEnvironmentService(
...
+			Message: localizer.MustLocalize(locale.InternalServerError),
+		})
+		if err != nil {
+			return nil, auth.StatusInternal.Err()
+		}
+		return nil, dt.Err()
+	}
+
+	// check account exist in any organization
+	account, err := s.accountClient.GetMyOrganizationsByEmail(ctx, &accproto.GetMyOrganizationsByEmailRequest{
```
cre8ivejp, line 202: This request doesn't include archived organizations.
It would be safer if we use the GetAccountV2 API to ensure the user doesn't exist regardless of the organization status.
Otherwise, if the organization is archived, the user would be able to create another organization.

pkg/environment/api/api.go
```diff
@@ -73,22 +110,298 @@ func NewEnvironmentService(
...
+		dt, err := auth.StatusInternal.WithDetails(&errdetails.LocalizedMessage{
+			Locale:  localizer.GetLocale(),
+			Message: localizer.MustLocalize(locale.InternalServerError),
+		})
+		if err != nil {
+			return nil, auth.StatusInternal.Err()
+		}
+		return nil, dt.Err()
+	}
+
```
cre8ivejp, line 200: We should validate whether the email is verified (EmailVerified).

pkg/environment/api/error.go
```diff
@@ -56,4 +56,5 @@ var (
 	statusUnauthenticated           = gstatus.New(codes.Unauthenticated, "environment: unauthenticated")
 	statusPermissionDenied          = gstatus.New(codes.PermissionDenied, "environment: permission denied")
 	statusNotFound                  = gstatus.New(codes.NotFound, "environment: not found")
+	statusDemoSiteNotEnabled        = gstatus.New(codes.FailedPrecondition, "environment: demo site is not enabled")
```
cre8ivejp, line 59:
```suggestion
	statusDemoSiteDisabled        = gstatus.New(codes.FailedPrecondition, "environment: demo site is disabled")
```

pkg/environment/api/api.go
```diff
@@ -73,22 +110,298 @@ func NewEnvironmentService(
...
+	req *environmentproto.ExchangeDemoTokenRequest,
+) (*environmentproto.ExchangeDemoTokenResponse, error) {
+	localizer := locale.NewLocalizer(ctx)
+	if !s.opts.isDemoSiteEnabled {
+		s.logger.Error("Demo site is not enabled",
+			zap.Any("type", req.Type),
+			zap.String("code", req.Code),
+			zap.String("redirect_url", req.RedirectUrl),
+		)
+		dt, err := statusDemoSiteNotEnabled.WithDetails(&errdetails.LocalizedMessage{
```
cre8ivejp, line 148:
```suggestion
		s.logger.Error("Demo site is disabled",
			zap.Any("type", req.Type),
			zap.String("code", req.Code),
			zap.String("redirect_url", req.RedirectUrl),
		)
		dt, err := statusDemoSiteDisabled.WithDetails(&errdetails.LocalizedMessage{
```

pkg/environment/api/api.go
hvn2k1: I've removed that

pkg/environment/api/api.go
hvn2k1: Got it, I created new DemoCreationToken proto struct and use it instead

pkg/web/cmd/server/server.go
```diff
@@ -1022,6 +1028,45 @@ func (s *server) createAuthService(
...
+	config *auth.OAuthConfig,
+	logger *zap.Logger,
+) (rpc.Service, error) {
+	signer, err := token.NewSigner(*s.oauthPrivateKeyPath)
+	if err != nil {
+		return nil, err
+	}
+	serviceOptions := []environmentapi.Option{
+		environmentapi.WithLogger(logger),
+		environmentapi.WithRefreshTokenTTL(*s.refreshTokenTTL),
```
cre8ivejp, line 1046: We don't use the refresh token.

```suggestion
```

pkg/environment/api/api.go
```diff
@@ -73,22 +110,332 @@ func NewEnvironmentService(
...
+	organizations, _, _, err := s.orgStorage.ListOrganizations(ctx, &mysql.ListOptions{
+		Limit:  mysql.QueryNoLimit,
+		Offset: mysql.QueryNoOffset,
+		Filters: []*mysql.FilterV2{
+			{
+				Column:   "organization.archived",
+				Operator: mysql.OperatorEqual,
+				Value:    false,
+			},
+		},
```
cre8ivejp, line 267: I commented [here](https://github.com/bucketeer-io/bucketeer/pull/2004#discussion_r2209689759) that if the organization is archived in the future, the user can create another demo. That's why I suggested using the `GetAccountV2`, but checking the [query](https://github.com/bucketeer-io/bucketeer/blob/main/pkg/account/storage/v2/sql/account_v2/select_accounts_with_organization.sql#L33) it does not filter by archived, so the `GetMyOrganizationsByEmail` was correct.
Can you revert to using it?

cre8ivejp (approved):
Nice work!

PR #2005, opened 2025-07-17
chore: remove segment IsInUseStatus validation

### LINKED ISSUE #2000: chore: allow the user to update segment user list even if it is in use
Currently, the user must remove the segment from any flags before updating it because changing the user list could affect the experiment accuracy if the flag is being used for tests. However, there are other cases when the same segment could be used in many features, making it very hard to update the segment.

So, we should allow the users to update the user list but warn them that doing so could affect the experiment's accuracy, and let the users decide what works best for them.

nnnkkk7:
Remove validation that prevents segment user list updates when in use.
fix #2000

### DIFF
--- a/pkg/feature/api/segment_user.go
+++ b/pkg/feature/api/segment_user.go
@@ -400,16 +400,6 @@ func (s *FeatureService) BulkUploadSegmentUsers(
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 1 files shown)


cre8ivejp (approved):
Thanks!

PR #2010, opened 2025-07-17
build: implement cacher manager for dev container

cre8ivejp:
Fix https://github.com/bucketeer-io/bucketeer/issues/2013
This will help load much faster when reopening the dev container.


Ubisoft-potato (approved):
Nice! LGTM.

PR #2012, opened 2025-07-18
chore: remove email verified validation when authenticating

cre8ivejp:
We should not check the email, because most users are not verified. We could introduce this as optional in the future if required.


Ubisoft-potato (approved):
LGTM!

PR #2018, opened 2025-07-23
fix: showing false permission denied error

cre8ivejp:
When accessing the Organization settings with a non-system admin account, it showed permission denied error logs.
Since the `checkOrganizationRole` already checks for the system admin, I have removed the `checkSystemAdminRole` check to prevent false logs.


Ubisoft-potato (approved):
LGTM!

PR #2021, opened 2025-07-23
fix: different Redis instances  are used to store and retrieve user attribute keys

kakcy:


test/e2e/gateway/api_grpc_test.go
```diff
@@ -1058,14 +1054,19 @@ func TestGetUserAttributeKeys(t *testing.T) {
 	}
 
 	foundAttributes := make(map[string]bool)
```
cre8ivejp, line 1056: If the boolean is not used (It's always set to true), the `map[string]struct{}` is lighter and better.
For testing is fine, but keep this in mind when developing for production.

cre8ivejp (approved):
Thank you.

PR #2029, opened 2025-07-24
chore: change the ttl of UserAttributeKeys to config management

kakcy:
This pull request introduces a new feature to support configurable TTL (Time-To-Live) for user attributes in the caching layer and updates related configurations and tests. The main changes include adding a `UserAttributeTTLDay` parameter, modifying the caching logic to use this parameter, and updating tests to validate the new functionality.

This PR allows you to change the storage period of `UserAttributeKeys` depending on the environment, improving the performance of e2e tests.

### DIFF
--- a/pkg/cache/v3/user.go
+++ b/pkg/cache/v3/user.go
@@ -26,14 +26,14 @@ import (
 )
 
 const (
-	userAttributeKind     = "user_attr"
-	userAttributesMaxSize = int64(100)
-	userAttributeTTL      = time.Duration(30 * 24 * time.Hour)
+	userAttributeKind          = "user_attr"
+	userAttributesMaxSize      = int64(100)
+	defalutUserAttributeTTLDay = 30
 )
 
 type UserAttributesCache interface {
 	GetUserAttributeKeyAll(environmentId string) ([]string, error)
-	Put(userAttributes *userproto.UserAttributes) error
+	Put(userAttributes *userproto.UserAttributes, ttlDay int) error
 }
 
 type userAttributesCache struct {
@@ -82,7 +82,7 @@ func (u *userAttributesCache) GetUserAttributeKeyAll(environmentId string) ([]st
 	return attributeKeys, nil
 }
 
-func (u *userAttributesCache) Put(userAttributes *userproto.UserAttributes) error {
+func (u *userAttributesCache) Put(userAttributes *userproto.UserAttributes, ttlDay int) error {
 	if userAttributes == nil {
 		return errors.New("user attributes is nil")
 	}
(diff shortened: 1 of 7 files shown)


pkg/cache/v3/user.go
```diff
@@ -92,7 +92,11 @@ func (u *userAttributesCache) Put(userAttributes *userproto.UserAttributes) erro
 		for _, value := range attribute.Values {
 			pipe.SAdd(key, value)
 		}
-		pipe.Expire(key, userAttributeTTL)
+		setTTL := defalutUserAttributeTTL
+		if ttl > 0 {
+			setTTL = ttl
+		}
+		pipe.Expire(key, time.Duration(setTTL)*time.Second)
```
cre8ivejp, line 99: We don't need the default value because if the user wants to keep the keys permanently, they will set it to zero.
Also, we already set the default value in the config.

```suggestion
		pipe.Expire(key, time.Duration(ttl)*time.Second)
```

pkg/cache/v3/user.go
```diff
@@ -26,14 +26,14 @@ import (
 )
 
 const (
-	userAttributeKind     = "user_attr"
-	userAttributesMaxSize = int64(100)
-	userAttributeTTL      = time.Duration(30 * 24 * time.Hour)
+	userAttributeKind       = "user_attr"
+	userAttributesMaxSize   = int64(100)
+	defalutUserAttributeTTL = 7 * 24 * 60 * 60 // 7 days
```
cre8ivejp, line 31:
```suggestion
```

pkg/cache/v3/user.go
```diff
@@ -82,7 +81,7 @@ func (u *userAttributesCache) GetUserAttributeKeyAll(environmentId string) ([]st
 	return attributeKeys, nil
 }
 
-func (u *userAttributesCache) Put(userAttributes *userproto.UserAttributes) error {
+func (u *userAttributesCache) Put(userAttributes *userproto.UserAttributes, ttl int) error {
```
cre8ivejp, line 84: This way, the caller can use 5 * time.Minute, 30*time.Second, etc., without having to remember “this int is in seconds.

```suggestion
func (u *userAttributesCache) Put(userAttributes *userproto.UserAttributes, ttl time.Duration) error {
```

pkg/cache/v3/user.go
```diff
@@ -92,7 +91,7 @@ func (u *userAttributesCache) Put(userAttributes *userproto.UserAttributes) erro
 		for _, value := range attribute.Values {
 			pipe.SAdd(key, value)
 		}
-		pipe.Expire(key, userAttributeTTL)
+		pipe.Expire(key, time.Duration(ttl)*time.Second)
```
cre8ivejp, line 94:
```suggestion
		pipe.Expire(key, ttl)
```

pkg/cache/v3/user.go
kakcy: Thank you for the good suggestion!
I changed the ttl argument to time.Duration type.
https://github.com/bucketeer-io/bucketeer/pull/2029/commits/9fca7379a7456cef44d9597c224639a446caa300

cre8ivejp (approved):
Nice work 👍 
In another PR, having an e2e test that tests the expiration would be nice.

Issue #2034, opened 2025-07-28
fix: permission denied when account is organization owner


hvn2k1:
### Version

v1.3.0

### Describe the bug

If the account is organization owner, all API returned permission denied. This is due to the logic that checking if user is admin then skip environment role check

PR #2035, opened 2025-07-28
fix: permission denied when account is organization owner

hvn2k1:
To solve https://github.com/bucketeer-io/bucketeer/issues/2034

### DIFF
--- a/pkg/role/role.go
+++ b/pkg/role/role.go
@@ -128,7 +128,7 @@ func checkRole(
 	organizationRole accountproto.AccountV2_Role_Organization,
 	isAdmin bool,
 ) (*eventproto.Editor, error) {
-	if organizationRole == accountproto.AccountV2_Role_Organization_ADMIN {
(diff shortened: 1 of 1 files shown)


pkg/role/role.go
```diff
@@ -128,7 +128,7 @@ func checkRole(
 	organizationRole accountproto.AccountV2_Role_Organization,
 	isAdmin bool,
 ) (*eventproto.Editor, error) {
-	if organizationRole == accountproto.AccountV2_Role_Organization_ADMIN {
+	if organizationRole >= accountproto.AccountV2_Role_Organization_ADMIN {
```
kakcy, line 131: There are several other places where similar decisions have been made, but do they need to be revised?
https://github.com/bucketeer-io/bucketeer/blob/9a0727e07d7eba45637f3baf3b9aa979bde12f93/pkg/account/domain/account.go#L71

hvn2k1: that's a very nice catch, I've updated other places too

PR #2036, opened 2025-07-28
ci: fix node version in push image workflow

hvn2k1:
Reference: https://github.com/bucketeer-io/bucketeer/actions/runs/16560060717/job/46828083672


nnnkkk7 (approved):
Thanks!

PR #2045, opened 2025-08-01
fix(ui/dashboard): pagination list view when searching

steveninhle:
- Fixed searching doesn't work when changing pages
- Updated the auth buttons translation


cre8ivejp (approved):
Thank you!

PR #2068, opened 2025-08-18
build: implement minikube caching to speed up the start process

cre8ivejp:
Fix https://github.com/bucketeer-io/bucketeer/issues/2069

This cache system makes it 80% faster when rebuilding or reopening the dev container.
Even if we run `mikube delete`, it won't delete the image cache.

Before running `make start-minikube`, we can run the following command just once as a fresh start to cache or update the cache images.

```sh
make -C tools/dev pre-cache-docker-images
```

It uses a volume for the Docker images, so when reopening, the image will be there when starting the minikube, making it really fast.

In case we need to delete all the cache.

```sh
make -C tools/dev clean-cache
```

I also have added a light version for deleting only unused images.

```sh
make -C tools/dev clean-cache-light
```

### DIFF
--- a/tools/dev/MINIKUBE_CACHING.md
+++ b/tools/dev/MINIKUBE_CACHING.md
@@ -58,6 +58,15 @@ make -C tools/dev pre-cache-docker-images
 
 # Setup minikube with pre-cached images
 make -C tools/dev setup-minikube-cached
+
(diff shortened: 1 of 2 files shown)


Ubisoft-potato (approved):
Great work!!
It will save a lot of time when rebuilding.

PR #2095, opened 2025-08-29
feat: support not equal operator for evaluation

### LINKED ISSUE #2096: feat: support not equal operator for evaluation
### Describe the feature you'd like

Use `≠` operator for evaluation 

### Background

_No response_

### Alternative solutions

_No response_

nnnkkk7:
fix #2096

### Summary

Add NOT_EQUALS operator to feature flag targeting rules, allowing users to create conditions that match when attribute
 values are not equal to specified values.

### What's Changed logic

  The NOT_EQUALS operator returns:
  - true when the attribute value is not equal to any of the specified values
  - false when the attribute value equals at least one of the specified values

  Example:
  - Attribute: "100", Values: ["100", "200"], Operator:
  NOT_EQUALS → false (equals "100")
  - Attribute: "300", Values: ["100", "200"], Operator:
  NOT_EQUALS → true (not equal to any)

UI
スクリーンショット 2025-08-29 午後6 18 26

### DIFF
--- a/ui/web-v2/src/assets/lang/en.json
+++ b/ui/web-v2/src/assets/lang/en.json
@@ -295,7 +295,7 @@
   "feature.clause.operator.before": "before",
   "feature.clause.operator.endWith": "ends with",
   "feature.clause.operator.equal": "=",
-  "feature.clause.operator.notEqual": "≠ (not equals)",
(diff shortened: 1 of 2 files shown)


cre8ivejp (approved):
Nice work!

PR #2107, opened 2025-09-03
fix: change log target to cloned feature

### LINKED ISSUE #1768: fix: flag cloning is not create an audit log in the destination environment
Cloning a flag doesn't create an audit log in the destination environment.

hvn2k1:
To resolve #1768

Add source environment and target environment in the event message


cre8ivejp (approved):
Thank you!

PR #2118, opened 2025-09-10
deps: upgrade golangci-lint to v2

### LINKED ISSUE #2098: deps: upgrade golangci-lint to v2
golangci-lint v2 has released, so we should migrate it.
ref: https://golangci-lint.run/docs/product/migration-guide/

nnnkkk7:
fix #2098

- Execute `golangci-lint migrate` and update config. 
ref: https://golangci-lint.run/docs/product/migration-guide/
- Fix lint errors.

### DIFF
--- a/.github/workflows/pr-go.yaml
+++ b/.github/workflows/pr-go.yaml
@@ -178,6 +178,6 @@ jobs:
           go-version-file: 'go.mod'
           cache: true
       - name: Install golangci-lint
-        run: go install github.com/golangci/golangci-lint/cmd/golangci-lint@latest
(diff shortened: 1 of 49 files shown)


.github/workflows/pr-go.yaml
```diff
@@ -178,6 +178,6 @@ jobs:
           go-version-file: 'go.mod'
           cache: true
       - name: Install golangci-lint
-        run: go install github.com/golangci/golangci-lint/cmd/golangci-lint@latest
+        run: go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.4.0
```
nnnkkk7, line 181: It's better to pin the version.

Issue #1253, opened 2024-09-20
enhancement: move i18n logic to frontend


cre8ivejp:
Now, we have part of the i18n implementation on both sides, front and backend. Because the i18n implementation in all APIs made the backend code too hard to read, we decided to do this only on the front end.

t-kikuc:
TODO: 

- [x] Refactor error structs in backend
- [ ] Update frontend
- [ ] Delete old i18 logics in backend

t-kikuc:
revised TODO: 

- [ ] Refactor error structs in backend
    - [x] Define the new struct
    - [ ] delete all code depending on `localizer`
    - [ ] delete `localizer` itself (keep `locale` package including `locale.go`)
- [ ] Update frontend 
- [ ] Delete localizations in `DomainEvent` (migrations seem to be unnecessary)

t-kikuc:
revised TODO: 

- [x] refine the RFC #2205
- [ ] backend:
    - [ ] copy message template files to frontend #2259
    - [ ] refine the error struct #2207
    - [ ] update old errors that use the localizer (so many files)
- [ ] frontend: 
    - [ ] step1. temporal way using `message`(see [the RFC](https://github.com/bucketeer-io/bucketeer/blob/main/docs/rfcs/move-i18n-logic-to-frontend.md#frontend-updates))
    - [ ] step2. complete way

t-kikuc:
Completed!

PR #2125, opened 2025-09-19
refactor: change error definition in account PKG to Bucketeer error

kakcy:
This pull request refactors error handling in the account API service to standardize gRPC error responses using the new `api.NewGRPCStatus` helper. The previous approach of manually constructing localized error messages and attaching them to gRPC status objects is replaced with a simpler, unified method. This change improves consistency and maintainability across both main service code and associated tests.

I have also standardized the process of assigning ErrorInfo to GRPCStatus for each error via` api.NewGRPCStatus`.
This ErrorInfo will be used when moving i18n to the frontend(https://github.com/bucketeer-io/bucketeer/issues/1253), and we plan to generate error messages etc. based on this information on the frontend.


nnnkkk7 (dismissed):
Thanks you！

PR #2134, opened 2025-09-25
chore: add temporary logging for missing feature dependencies

nnnkkk7:
Add `printf` temporary debug log because it's difficult to add logger.


cre8ivejp (approved):
Thanks

PR #2170, opened 2025-10-27
chore: reduce info log verbosity in all services

cre8ivejp:


manifests/bucketeer/charts/batch/templates/cronjob.yaml
```diff
@@ -65,11 +65,8 @@ spec:
                   rm -f ${RESPONSE}
                   END_TIME=$(date +%s)
                   DURATION=$((END_TIME - START_TIME))
-                  # API returns empty {} on success, or error details on failure
-                  if [ "${RES}" = "200" ]; then
-                    # Success - use DEBUG to avoid verbose logs
-                    echo '{"severity":"DEBUG","message":"'${JOB_NAME}' job completed successfully","job":"'${JOB_NAME}'","jobId":"'${JOB_ID}'","statusCode":'${RES}',"durationSeconds":'${DURATION}'}'
```
cre8ivejp: Because the app log level doesn't affect the cronjob log level, even when setting it to debug, it was sending debug logs to production environments.
Since we only need the error log, I have removed the success message.

PR #2171, opened 2025-10-28
chore: suppress context.DeadlineExceeded errors in the api event publishing

cre8ivejp:


pkg/subscriber/processor/evaluation_events_evaluation_count_event_persister.go
```diff
@@ -361,11 +361,10 @@ func (p *evaluationCountEventPersister) cacheEnvLastUsedInfo(
 	p.envLastUsedCacheMutex.Lock()
 	defer p.envLastUsedCacheMutex.Unlock()
 	var clientVersion string
-	if event.User == nil {
-		p.logger.Warn("Failed to cache last used info. User is nil.",
-			zap.String("environmentId", environmentId))
```
Ubisoft-potato: We dont need the warning log when event user is nil, right?

PR #2174, opened 2025-10-28
fix: remove demoOrganizationCreationNotifier from ConfigMap when demoSiteEnabled is false

### LINKED ISSUE #2173: fix: subscriber service logs recurring "processor not found" errors
### Version

v2.1.1

## Problem

The subscriber service logs recurring "processor not found" errors when `demoSiteEnabled` is set to `false` (the default value)

## Root Cause

There is a mismatch between the Helm ConfigMap and the processor registration logic in the code:

1. **Helm Configuration** ([values.yaml:191-200](https://github.com/bucketeer-io/bucketeer/blob/main/manifests/bucketeer/charts/subscriber/values.yaml#L191-L200))
   - `demoOrganizationCreationNotifier` is always included in the `subscribers` list

2. **Default Flag** ([values.yaml:11](https://github.com/bucketeer-io/bucketeer/blob/main/manifests/bucketeer/charts/subscriber/values.yaml#L11))
   - `demoSiteEnabled: false` is the default value

3. **Processor Registration** ([server.go:558-568](https://github.com/bucketeer-io/bucketeer/blob/main/pkg/subscriber/cmd/server/server.go#L558-L568))
   - The processor is only registered when `demoSiteEnabled=true`
   - When `demoSiteEnabled=false`, the processor is not registered

4. **Runtime Behavior** ([server.go:437-446](https://github.com/bucketeer-io/bucketeer/blob/main/pkg/subscriber/cmd/server/server.go#L437-L446))
   - The service attempts to look up the processor from the ConfigMap
   - Fails to find it and logs an error (but continues execution)

## Solution

Modified the Helm template ([subscribers-configmap.yaml](https://github.com/bucketeer-io/bucketeer/blob/main/manifests/bucketeer/charts/subscriber/templates/subscribers-configmap.yaml)) to conditionally exclude `demoOrganizationCreationNotifier` based on `demoSiteEnabled`:

### Changes

**1. subscribers.json section:**
```yaml
subscribers.json: |-
  {{- $subscribers := .Values.subscribers }}
  {{- if not .Values.env.demoSiteEnabled }}
  {{- $subscribers = omit $subscribers "demoOrganizationCreationNotifier" }}
  {{- end }}
  {{- range $name, $config := $subscribers }}
  # ... rest of the configuration
  {{- end }}
  {{ toJson $subscribers }}
```

**2. processors.json section:**
```yaml
processors.json: |-
  {{- $processors := .Values.processors }}
  {{- if not .Values.env.demoSiteEnabled }}
  {{- $processors = omit $processors "demoOrganizationCreationNotifier" }}
  {{- end }}
  {{- range $name, $config := $processors }}
  # ... rest of the configuration
  {{- end }}
  {{ toJson $processors }}
```

## Design Pattern Compatibility

This solution uses the Helm `omit` function, which is consistent with existing patterns in the codebase:
- Existing templates already use conditional logic with `{{- if ... }}`
- The `omit` function is a standard Sprig template function
- Maintains the existing JSON generation approach with `toJson`

_No response_

### Expected behavior

_No response_

### Relevant log output

```bash

```

### Additional information

_No response_

nnnkkk7:
fix #2173

I tested this change on minikube and didn't find this error.

### DIFF
--- a/manifests/bucketeer/charts/subscriber/templates/subscribers-configmap.yaml
+++ b/manifests/bucketeer/charts/subscriber/templates/subscribers-configmap.yaml
@@ -11,7 +11,11 @@ metadata:
     heritage: {{ .Release.Service }}
 data:
   subscribers.json: |-
-    {{- range $name, $config := .Values.subscribers }}
+    {{- $subscribers := .Values.subscribers }}
+    {{- if not .Values.env.demoSiteEnabled }}
+    {{- $subscribers = omit $subscribers "demoOrganizationCreationNotifier" }}
+    {{- end }}
+    {{- range $name, $config := $subscribers }}
     {{- $_ := set $config "pubSubType" $.Values.global.pubsub.type }}
     {{- $_ := set $config "redisServerName" $.Values.global.pubsub.redis.serverName }}
     {{- $_ := set $config "redisAddr" $.Values.global.pubsub.redis.addr }}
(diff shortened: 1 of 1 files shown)


hvn2k1 (approved):
Nice catch, thank you!

Issue #1894, opened 2025-06-26
feat: implement api key last seen


cre8ivejp:
Some users want to know when the API key was used on the admin console.

hvn2k1:
I propose 2 solutions for this:

## A. Use Pub/sub to receive last used info event
- This is the same flow when we save feature last used at events: https://github.com/bucketeer-io/bucketeer/blob/main/pkg/subscriber/processor/evaluation_events_evaluation_count_event_persister.go
- API gateway receive request and publish API usage event. The api key last used info subscriber receive event and cache it in-memory, another go routine is spawned and read from in-memory cache to write into database
<img width="1262" height="780" alt="Image" src="https://github.com/user-attachments/assets/4c8b44cb-396d-4987-a8de-451a1dfef430" />

### Pros
- Persistent, we can control if the event is ack or not
### Cons
- Can be costly because the number of events go through pub/sub

## B. Use Redis to cache events in API gateway and save periodically
- API gateway receive request then save into Redis,  is spawned and read from Redis to write into database

<img width="982" height="644" alt="Image" src="https://github.com/user-attachments/assets/1d846f4e-2a2f-466e-a6c6-494a25dbc867" />

### Pros
- Lower pricing than Pub/sub
### Cons
- Not persistent as Pub/sub

PR #2175, opened 2025-10-28
feat: implement api key last used at

hvn2k1:
To resolve https://github.com/bucketeer-io/bucketeer/issues/1894

### DIFF
--- a/pkg/account/api/error.go
+++ b/pkg/account/api/error.go
@@ -48,5 +48,4 @@ var (
 	statusSearchFilterIDIsEmpty                  = api.NewGRPCStatus(pkgErr.NewErrorInvalidArgEmpty(pkgErr.AccountPackageName, "search filter ID is empty", "search_filter_ID"))
 	statusSearchFilterIDNotFound                 = api.NewGRPCStatus(pkgErr.NewErrorNotFound(pkgErr.AccountPackageName, "search filter not found", "search_filter"))
 	statusInvalidListAPIKeyRequest               = api.NewGRPCStatus(pkgErr.NewErrorInvalidArgEmpty(pkgErr.AccountPackageName, "invalid list api key request", "list_api_key_request"))
-	statusInvalidLastUsedAt                      = api.NewGRPCStatus(pkgErr.NewErrorInvalidArgNotMatchFormat(pkgErr.AccountPackageName, "last used at is invalid", "last_used_at"))
 )


pkg/account/domain/api_key.go
```diff
@@ -112,6 +113,11 @@ func (a *APIKey) Update(
 	if disabled != nil {
 		updated.Disabled = disabled.Value
 	}
+	if lastUsedAt != nil {
+		if a.LastUsedAt < lastUsedAt.Value {
+			updated.LastUsedAt = lastUsedAt.Value
+		}
+	}
```
cre8ivejp: Just to let you know, we can't allow the developer to update this field. This must be updated only by the backend.

pkg/api/api/save_api_key_last_used_at.go
```diff
@@ -0,0 +1,117 @@
...
+)
+
+type lastUsedInfo struct {
+	envAPIKey  *account.EnvironmentAPIKey
+	lastUsedAt int64
+}
+
+type apikeyLastUsedInfoCache map[string]lastUsedInfo
+
+type envAPIKeyLastUsedInfoCache map[string]apikeyLastUsedInfoCache
```
cre8ivejp: Can't we have just one map?
We just need the api_key_id and the timestamp.

pkg/api/api/save_api_key_last_used_at.go
```diff
@@ -0,0 +1,117 @@
...
+
+func (s *grpcGatewayService) cacheAPIKeyLastUsedAt(
+	envAPIKey *account.EnvironmentAPIKey,
+	lastUsedAt int64,
+) {
+	s.envAPIKeyLastUsedInfoMutex.Lock()
+	defer s.envAPIKeyLastUsedInfoMutex.Unlock()
+
+	if cache, ok := s.envAPIKeyLastUsedInfoCacher[envAPIKey.Environment.Id]; ok {
```
cre8ivejp: Why do we need to save this in Redis?
Since this data is not critical, we just need to keep it in memory and periodically save it in the DB.
Just like we do when saving the feature flag last used info.

hvn2k1: We don't save it in redis here, it's just a normal in memory map

pkg/api/api/save_api_key_last_used_at.go
hvn2k1: I've switch to sync.Map

pkg/account/domain/api_key.go
hvn2k1: https://github.com/bucketeer-io/bucketeer/pull/2175/files#diff-1236610833a324b8af6d6273437bdeb2fb35012b5ae3b3374760927a8427c204R777

pkg/api/api/save_api_key_last_used_at.go
```diff
@@ -0,0 +1,117 @@
...
+type apikeyLastUsedInfoCache map[string]lastUsedInfo
+
+type envAPIKeyLastUsedInfoCache map[string]apikeyLastUsedInfoCache
+
+func (s *grpcGatewayService) cacheAPIKeyLastUsedAt(
+	envAPIKey *account.EnvironmentAPIKey,
+	lastUsedAt int64,
+) {
+	s.envAPIKeyLastUsedInfoMutex.Lock()
+	defer s.envAPIKeyLastUsedInfoMutex.Unlock()
```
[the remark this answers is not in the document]
hvn2k1: I've switch to sync.Map https://github.com/bucketeer-io/bucketeer/pull/2175/files#diff-193246ee4b1236d00412b96fc8b36407d8c06569aabedd52f67725a2a6f7c438R174

pkg/account/domain/api_key.go
hvn2k1: Using the account storage interface directly would require us to init a mysql client inside the api gateway service, then we'll have to update the helmchart, configuration, passing it into the constructor, etc... while a small internal API took less work than that

cre8ivejp: Let's keep it as it is, then.

pkg/api/api/save_api_key_last_used_at.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+	s.envAPIKeyLastUsedInfoCacher[envAPIKey.Environment.Id] = cache
+}
+
+func (s *grpcGatewayService) writeAPIKeyLastUsedAtCacheToDatabase(ctx context.Context) {
+	ticker := time.NewTicker(time.Minute)
+	for {
+		select {
+		case <-ctx.Done():
+			return
```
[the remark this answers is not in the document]
hvn2k1: I've switch to the app context https://github.com/bucketeer-io/bucketeer/pull/2175/files#diff-8ce90509dcb5910a6395cd6add277d6d21f6676649bc74683478f779db9a8b92R461-R462

pkg/account/api/api_key.go
```diff
@@ -773,3 +773,72 @@ func (s *AccountService) UpdateAPIKey(
...
+			return nil, dt.Err()
+		}
+		s.logger.Error(
+			"Failed to update api key last used at",
+			log.FieldsFromIncomingContext(ctx).AddFields(
+				zap.Error(err),
+				zap.String("id", req.ApiKeyId),
+			)...,
+		)
+		return nil, api.NewGRPCStatus(err).Err()
```
cre8ivejp: There is an inconsistency here.
When the error is not found, the localizer logic is used. Since this API is for internal use only, we don't need to use the localizer. It's also deprecated.
WDYT?

https://github.com/bucketeer-io/bucketeer/pull/2175/files#diff-1236610833a324b8af6d6273437bdeb2fb35012b5ae3b3374760927a8427c204R813-R815

pkg/account/api/api_key.go
```diff
@@ -773,3 +773,72 @@ func (s *AccountService) UpdateAPIKey(
...
+	}
+	return &proto.UpdateAPIKeyLastUsedAtResponse{}, nil
+}
+
+func validateUpdateAPIKeyLastUsedAtRequest(
```
cre8ivejp: We should also check the lastUsedAt to ensure it is not zero.

pkg/account/api/api_key.go
```diff
@@ -773,3 +773,72 @@ func (s *AccountService) UpdateAPIKey(
...
+
+	err = s.mysqlClient.RunInTransactionV2(ctx, func(contextWithTx context.Context, _ mysql.Transaction) error {
+		envAPIKey, err := s.accountStorage.GetEnvironmentAPIKey(contextWithTx, req.ApiKeyId)
+		if err != nil {
+			return err
+		}
+		apiKey := &domain.APIKey{
+			APIKey: envAPIKey.ApiKey,
+		}
+		apiKey.UsedAt(req.LastUsedAt)
```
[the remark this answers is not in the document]
hvn2k1: I think we can ignore the updated_at field when we update the last_used_at field as the API and the function here is **not designed to Update the APIKey main metadata**

pkg/api/api/api_grpc.go
```diff
@@ -1472,7 +1480,7 @@ func (s *grpcGatewayService) checkTrackRequest(
 		s.logger.Error("Failed to check environment API key",
 			log.FieldsFromIncomingContext(ctx).AddFields(
 				zap.Error(err),
-				zap.Any("envAPIKey", envAPIKey),
+				zap.Any("apikey", envAPIKey),
```
cre8ivejp: The value is the envAPIKey, not only the apiKey.

pkg/api/api/api_grpc.go
```diff
@@ -1499,11 +1507,14 @@ func (s *grpcGatewayService) checkRequest(
...
```
cre8ivejp: Same here.

pkg/account/api/api_key.go
cre8ivejp: Additionally, by passing the `err` without formatting it, an unknown error will be generated.
https://github.com/bucketeer-io/bucketeer/blob/main/pkg/api/api/grpc_status.go#L41

Since we already have the `statusInternal`, let's use it.

Reference: https://github.com/bucketeer-io/bucketeer/blob/main/pkg/account/storage/v2/api_key.go#L94-L103

pkg/api/api/save_api_key_last_used_at.go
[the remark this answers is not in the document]
hvn2k1: https://github.com/bucketeer-io/bucketeer/pull/2175/files#diff-0972635bb3a20f989b8f81f3c9aca778c7e53e819c45a1691fc4a084ffe003b4R44-R45 I have added write for graceful

PR #2185, opened 2025-10-31
fix: update last seen when using system admin

### LINKED ISSUE #2192: fix: system admin's last login not displayed in organization member page
### Version

v2.1.1

### Describe the bug

When a system administrator creates a new Organization using their email as the owner, the member page incorrectly shows that they have "never logged in to the console," even though they are currently logged in.

## Current Behavior

1. A system admin creates a new Organization with their email as the owner
2. The system admin logs into the console
3. On the Organization's member page, the owner account shows "never" in the "Last Seen" column
4. This happens because the `last_seen` field remains `0` in the database

## Expected Behavior

When a system admin who owns an Organization logs into the console, their "Last Seen" status should be updated and displayed correctly on the member page, showing the actual login time instead of "never."

nnnkkk7:
fix #2192 

## Summary

Fixes an issue where system administrators who create and own Organizations show "never" in the "Last Seen" column on the member page, despite being actively logged in.

## Problem

When a system admin creates a new Organization using their email as the owner and logs into the console, the Organization's member page incorrectly displays that they have "never logged in." This occurs because:

1. System admin accounts belong to a special system admin Organization
2. When creating a new Organization, an owner account is created in that Organization with `last_seen = 0`
3. The `GetMe` endpoint only updates the account in the system admin's Organization, not the account in the Organization being viewed

## Solution

This PR implements a selective update strategy that updates the `last_seen` timestamp for the relevant accounts:

1. Always update the account in the Organization currently being viewed
2. If the viewed Organization differs from the system admin's Organization, also update the system admin's Organization account
3. Properly handle cases where the system admin doesn't have an account in the Organization being viewed

### DIFF
--- a/pkg/account/api/admin_account.go
+++ b/pkg/account/api/admin_account.go
@@ -114,8 +114,9 @@ func (s *AccountService) GetMe(
 	}
 	if sysAdminAccount != nil && !sysAdminAccount.Disabled {
 		adminEnvRoles := s.getAdminConsoleAccountEnvironmentRoles(environments, projects)
+		lastSeen := sysAdminAccount.LastSeen
 
-		if err := s.updateLastSeen(ctx, sysAdminAccount.Email, req.OrganizationId); err != nil {
+		if updated, err := s.updateLastSeen(ctx, sysAdminAccount.Email, req.OrganizationId); err != nil {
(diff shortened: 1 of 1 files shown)


cre8ivejp (approved):
Thank you!

PR #2203, opened 2025-11-12
fix: align migration imagePullSecrets with other services

cre8ivejp:
Fix https://github.com/bucketeer-io/bucketeer/issues/2202

Use `global.image.imagePullSecrets` for the migration Job, consistent with api, 
web, batch, and subscriber services. Adds optional override capability via 
migration.imagePullSecrets for special cases.

Fixes 403 errors when pulling migration images from private registries.


nnnkkk7 (approved):
Thank you!
LGTM!

Issue #2073, opened 2025-08-21
chore: remove old console and its references


cre8ivejp:
- [ ] Remove old console, including protobuf files

- [ ] Remove the old command layer

PR #2209, opened 2025-11-14
refactor: remove command layer from notification API

hvn2k1:
Part of https://github.com/bucketeer-io/bucketeer/issues/2073


proto/notification/command.proto
```diff
@@ -42,33 +42,4 @@ message DeleteAdminSubscriptionCommand {}
 
 message RenameAdminSubscriptionCommand {
```
cre8ivejp: Are we still using this?

cre8ivejp (approved):
👍

PR #2244, opened 2025-11-21
fix: api key last seen writer is not executing correctly after graceful shutdown

hvn2k1:
Part of https://github.com/bucketeer-io/bucketeer/issues/1894

Use dependency injection to allow calling the last seen info write action when shutting down the service gracefully


hvn2k1:
I decided to split the context only, instead of adding the logic in the cmd layer

PR #2246, opened 2025-11-21
docs: add rfc for auto-archive feature

nnnkkk7:
Add rfc for environment-level auto-archive system.

### DIFF
--- a/docs/rfcs/0044-feature-auto-archive.md
+++ b/docs/rfcs/0044-feature-auto-archive.md
@@ -54,6 +54,24 @@ CREATE INDEX idx_environment_auto_archive_enabled
   ON environment_v2 (auto_archive_enabled);
 ```
 
+#### Design Decision: Configuration Storage Approach
+
+We evaluated multiple approaches for storing auto-archive configuration.
+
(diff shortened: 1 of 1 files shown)


docs/rfcs/0044-feature-auto-archive.md
```diff
@@ -0,0 +1,230 @@
...
+
+### Database Schema Changes
+
+Add three columns to the existing `environment_v2` table to store auto-archive configuration:
+
+```sql
+ALTER TABLE environment_v2
+  ADD COLUMN auto_archive_enabled BOOLEAN NOT NULL DEFAULT FALSE,
+  ADD COLUMN auto_archive_unused_days INT NOT NULL DEFAULT 90,
+  ADD COLUMN auto_archive_require_no_code_refs BOOLEAN NOT NULL DEFAULT TRUE;
```
cre8ivejp: As environmental options increase, perhaps we should consider designing a proper structure for this? WDYT?

cre8ivejp: Sounds good!
Let's merge it after adding the decision section.

cre8ivejp (approved):
Thank you :tada:

PR #2249, opened 2025-11-25
fix: api key last seen writer is not executing correctly after graceful shutdown

hvn2k1:
Part of #1894


pkg/api/api/save_api_key_last_used_at.go
```diff
@@ -54,15 +54,17 @@ func (s *grpcGatewayService) cacheAPIKeyLastUsedAt(
 
 func (s *grpcGatewayService) writeAPIKeyLastUsedAtCacheToDatabase(ctx context.Context) {
 	ticker := time.NewTicker(time.Minute)
+	defer ticker.Stop()
+
+	writeCtx := context.Background()
```
cre8ivejp: Why not use `context.WithCancel`?

hvn2k1: I've switch to context.WithCancel

cre8ivejp (approved):
👍

Issue #1625, opened 2025-03-27
chore: improve security in the list api keys api


cre8ivejp:
Because the Client SDK API keys are read-only and can be found easily using debugging tools, we don't obfuscate the key in the response.

However, because we are implementing the Public API and those keys have editor and admin roles, a user with a Viewer role shouldn't be able to see them.

So, we must obfuscate all the keys, no matter the type, when returning the response.

We have already implemented this in the Flag Trigger API.
https://github.com/bucketeer-io/bucketeer/blob/main/pkg/feature/api/flag_trigger.go#L1225-L1235

PR #2250, opened 2025-11-25
chore: change to use uuid for the api key id

hvn2k1:
To resolve https://github.com/bucketeer-io/bucketeer/issues/1625

**NOTE: before deploy this, things need to be done:**
- Announce this to every client that their API key will be obfuscated

### DIFF
--- a/pkg/account/api/api_key.go
+++ b/pkg/account/api/api_key.go
@@ -35,7 +35,8 @@ import (
 )
 
 const (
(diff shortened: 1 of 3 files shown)


hvn2k1:
> @hvn2k1, how about the keys created recently? Don't we need a DB migration for this?

I'm thinking about splitting it into another pr!

Issue #2268, opened 2025-12-01
fix: e2e test failed due to saving API key id in create-api-key hack script


hvn2k1:
### Version

v2.1.1

### Describe the bug

create-api-key has yet being updated to switch to api_key column instead of id column

### Relevant log output

```bash

```

PR #2269, opened 2025-12-01
chore: fix to use the correct APIKey field when creating api keys

hvn2k1:
To resolve https://github.com/bucketeer-io/bucketeer/issues/2268


cre8ivejp (approved):
👍

Issue #2103, opened 2025-09-03
feat: retention for bucketeer


hvn2k1:
### Describe the feature you'd like

After a period of time, demo organizations will be deleted with its related data

PR #2271, opened 2025-12-02
feat: implement delete org and env data API

hvn2k1:
To resolve #https://github.com/bucketeer-io/bucketeer/issues/2103


proto/environment/service.proto
```diff
@@ -411,6 +411,18 @@ message CreateDemoOrganizationResponse {
...
+}
+
+message DeleteOrganizationDataResponse {}
+
+message DeleteEnvironmentDataRequest {
+  repeated string environment_ids = 1;
+}
+
+message DeleteEnvironmentDataResponse {}
+
```
Ubisoft-potato: The environment belongs to organization, we have a `organization_id` field in `environment_v2` table, so I think it make more sense to delete all organization realted data in one request, instead of spliting into two request.

Also we got foreign key: `environment_v2_foreign_project_id` in `environment_v2` table, when we delete data from `project`, related environment will be deleted.

hvn2k1: The DeleteOrganizationData API is to delete all the data inside organizations, include every of its environment
The DeleteEnvironmentData API is to delete all the data inside environments only

pkg/environment/storage/v2/sql/organization/delete_organization_data.sql
```diff
@@ -0,0 +1 @@
+DELETE FROM %s %s
```
cre8ivejp: Same here.

pkg/environment/api/organization.go
```diff
@@ -1470,3 +1492,133 @@ func (s *EnvironmentService) validateConvertTrialOrganizationRequest(
...
+	if err != nil {
+		s.logger.Error("Could not list environments", zap.Error(err))
+		return nil, err
+	}
+
+	err = s.mysqlClient.RunInTransactionV2(ctx, func(ctxWithTx context.Context, _ mysql.Transaction) error {
+		// 2. Delete all target entities from the environments
+		for _, environment := range environments {
+			for _, target := range targetEntitiesInEnvironment {
+				err := s.environmentStorage.DeleteTargetFromEnvironmentV2(ctxWithTx, environment.Id, target)
```
cre8ivejp: This is running a transaction per environment. If something goes wrong, it won't be rolled back, resulting in an incorrect database state.

E.g.

```go
func (s *EnvironmentService) DeleteOrganizationData(...) {    
    // SINGLE TRANSACTION for everything
    err = s.mysqlClient.RunInTransactionV2(ctx, func(ctxWithTx context.Context, _ mysql.Transaction) error {
        environments, _, _, err := s.environmentStorage.ListEnvironmentsV2(ctxWithTx, options)
        
        // Delete all entities from all environments
        for _, environment := range environments {
            for _, target := range targetEntitiesInEnvironment {
                err := s.environmentStorage.DeleteTargetFromEnvironmentV2(ctxWithTx, environment.Id, target)
            }
        }
        
        // Delete all environments
        err = s.environmentStorage.DeleteEnvironmentV2(ctxWithTx, whereParts)
        
        // Delete all projects
        err = s.projectStorage.DeleteProjects(ctxWithTx, whereParts)
        
        // Delete all organizations (calls deleteOrganizations with ctxWithTx)
        err = s.deleteOrganizations(ctxWithTx, req.OrganizationIds)
        
        return nil  // All-or-nothing
    })
}

func (s *EnvironmentService) deleteOrganizations(ctx context.Context, organizationIDs []string) error {
    // NO NEW TRANSACTION - uses parent's ctxWithTx
    for _, target := range targetEntitiesInOrganization {
        err := s.orgStorage.DeleteOrganizationData(ctx, target, whereParts)  // Uses existing transaction
    }
    err := s.orgStorage.DeleteOrganizations(ctx, whereParts)  // Uses existing transaction
}
```

test/e2e/environment/environment_v2_test.go
```diff
@@ -156,6 +158,58 @@ func TestUpdateEnvironmentV2(t *testing.T) {
...
+	// 3. delete env
+	_, err = envc.DeleteEnvironmentData(ctx, &environmentproto.DeleteEnvironmentDataRequest{
+		EnvironmentIds: []string{envID},
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// 4. verify data is deleted
+	_, err = ftc.GetFeature(ctx, &featureproto.GetFeatureRequest{
```
cre8ivejp: Same here. We must check all related data to the environment.

E.g.

```go
// Verify all data is deleted
testCases := []struct{
    name string
    check func() error
}{
    {"environment", func() error { return envc.GetEnvironmentV2(...) }},
    {"feature", func() error { return ftc.GetFeature(...) }},
    {"project", func() error { return envc.GetProject(...) }},
    {"organization", func() error { return envc.GetOrganization(...) }},
    // ... etc
}
for _, tc := range testCases {
    if err := tc.check(); err == nil {
        t.Errorf("%s should be deleted but still exists", tc.name)
    }
}
```

pkg/environment/storage/v2/sql/environment/delete_target_from_environment.sql
```diff
@@ -0,0 +1,4 @@
+DELETE FROM
+    %s
```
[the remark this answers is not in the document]
hvn2k1: Thank you for pointing out! let me try the whitelist method

pkg/environment/api/organization.go
```diff
@@ -41,6 +42,29 @@ import (
...
+		"goal",
+		"experiment",
+		"tag",
+		"ops_progressive_rollout",
+		"flag_trigger",
+		"code_reference",
+		"feature",
+		"api_key",
+		"audit_log",
+	}
```
t-kikuc: [ask] Doesn't this need to contain `feature_last_used_info`?

pkg/environment/api/environment_v2.go
```diff
@@ -656,6 +656,235 @@ func (s *EnvironmentService) ArchiveEnvironmentV2(
...
+			isRunning, err := s.checkRunningFeaturesInEnvironments(ctxWithTx, req.EnvironmentIds)
+			if err != nil {
+				s.logger.Error("Could not check running features in environments",
+					zap.Error(err),
+					zap.Strings("environmentIDs", req.EnvironmentIds),
+				)
+				return err
+			}
+			if isRunning {
+				return statusCannotDeleteOrganization.Err()
```
t-kikuc:
```suggestion
				return statusCannotDeleteEnvironment.Err()
```

is this correct?

Issue #2274, opened 2025-12-03
fix(ui/dashboard): add documentation link to code refs enable state


cre8ivejp:
Add documentation link to the code references empty state when there's no public API key with read/write permissions, matching the pattern already used when there are no code references. Updates both English and Japanese translations.

PR #2275, opened 2025-12-03
fix(ui/dashboard): add documentation link to code refs enable state

cre8ivejp:
Fix https://github.com/bucketeer-io/bucketeer/issues/2274

Add documentation link to the code references empty state when there's no public API key with read/write permissions, matching the pattern already used when there are no code references. Updates both English and Japanese translations.


steveninhle (approved):
LGTM!
Thank you!

Issue #2299, opened 2025-12-10
feat: support for PostgreSQL as data warehouse option


hvn2k1:
### Describe the feature you'd like

Implement Postgres data warehouse

MySQL reference: https://github.com/bucketeer-io/bucketeer/pull/1707

PR #2300, opened 2025-12-10
feat: implement postgreSQL as data warehouse option

hvn2k1:
1. Implement postgres for DWH
- [x] Docker compose
- [x] Dev container minikube

2. Other:
- Upgrade mysql helm chart to compatible with postgres helm chart, using old mysql chart cause common chart conflict
- Replace deprecated bitnami images

To resolve issue https://github.com/bucketeer-io/bucketeer/issues/2299
MySQL as DWH for reference: https://github.com/bucketeer-io/bucketeer/pull/1707

### DIFF
--- a/pkg/subscriber/storage/postgres_event_test.go
+++ b/pkg/subscriber/storage/postgres_event_test.go
@@ -1,0 +1,298 @@
+// Copyright 2026 The Bucketeer Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package storage
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/mock/gomock"
+
+	storagev2 "github.com/bucketeer-io/bucketeer/v2/pkg/subscriber/storage/v2"
+	mockv2 "github.com/bucketeer-io/bucketeer/v2/pkg/subscriber/storage/v2/mock"
+	epproto "github.com/bucketeer-io/bucketeer/v2/proto/eventpersisterdwh"
+)
+
+func TestPostgresEvalEventWriterAppendRows(t *testing.T) {
+	t.Parallel()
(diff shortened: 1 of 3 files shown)


hack/create-postgres-event-tables/command.go
```diff
@@ -0,0 +1,220 @@
...
+	if err != nil {
+		logger.Error("Failed to execute SQL statements", zap.Error(err))
+		return err
+	}
+
+	logger.Info("Successfully completed PostgreSQL event tables setup")
+	return nil
+}
+
+func (c *command) checkTablesExist(ctx context.Context, client postgres.Client, logger *zap.Logger) ([]string, error) {
```
nnnkkk7: Instead of issuing one query per table, we could fetch existing tables in a single query and check inclusion on the application side.
This would reduce database round-trips and scale better if we add more tables in the future.

hvn2k1: can you explain the single query idea? I don't understand

pkg/subscriber/cmd/server/server.go
```diff
@@ -113,6 +120,11 @@ func RegisterCommand(r cli.CommandRegistry, p cli.ParentCommand) cli.Command {
 		mysqlPort:        cmd.Flag("mysql-port", "MySQL port.").Required().Int(),
 		mysqlDBName:      cmd.Flag("mysql-db-name", "MySQL database name.").Required().String(),
 		mysqlDBOpenConns: cmd.Flag("mysql-db-open-conns", "MySQL open connections.").Required().Int(),
+		postgresUser:     cmd.Flag("postgres-user", "Postgres user.").Required().String(),
+		postgresPass:     cmd.Flag("postgres-pass", "Postgres password.").Required().String(),
+		postgresHost:     cmd.Flag("postgres-host", "Postgres host.").Required().String(),
+		postgresPort:     cmd.Flag("postgres-port", "Postgres port.").Required().Int(),
+		postgresDBName:   cmd.Flag("postgres-db-name", "Postgres database name.").Required().String(),
```
cre8ivejp: Since PostgreSQL is optional, we can't make these flags required.
The Data warehouse configurations must be passed via configMap only.

pkg/subscriber/cmd/server/server.go
```diff
@@ -213,6 +225,11 @@ func (s *server) Run(ctx context.Context, metrics metrics.Metrics, logger *zap.L
 	if err != nil {
 		return err
 	}
+	postgresClient, err := s.createPostgresClient(ctx, registerer, logger)
+	if err != nil {
+		// TODO: postgres client is optional for now to support mysql only setup, handle this more gracefully
+		logger.Warn("Failed to create Postgres client", zap.Error(err))
+	}
```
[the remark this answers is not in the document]
hvn2k1: I've removed the client creation in cmd layer

pkg/subscriber/cmd/server/server.go
hvn2k1: I've removed these flags and use config map only

PR #2331, opened 2026-01-08
ci: add migration ci for checking hash and validation

nnnkkk7:
Before merging pr and executing migration, we should check the atlas hash and validate migration file.

### DIFF
--- a/.github/workflows/pr-migration.yaml
+++ b/.github/workflows/pr-migration.yaml
@@ -27,16 +27,6 @@ jobs:
 
       - name: Validate migration files
         run: make migration-validate
-
-  hash-check:
-    runs-on: ubuntu-latest
-    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-migration.yaml
```diff
@@ -0,0 +1,42 @@
...
+    steps:
+      - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
+
+      - name: Setup Atlas
+        uses: ariga/setup-atlas@1ffa6444093025c9791791f0de623560359acd02 # v0
+        with:
+          version: ${{ env.ATLAS_VERSION }}
+
+      - name: Check atlas.sum is up to date
+        run: make migration-hash-check
```
cre8ivejp: This isn't necessary, but since this is not a heavy task, you could simplify by running it in just one job.

```suggestion
        run: make migration-validate
        
      - name: Check atlas.sum is up to date
        run: make migration-hash-check
```

Issue #1626, opened 2025-03-27
chore: define the id for environments with empty environment_id


cre8ivejp:
## Environment ID Migration: Empty String to UUID

There is an environment in Production with an empty `environment_id`. We must set a UUID for it, but because the keys saved in Redis also contain the environment_id in the key, we will need a data migration in Redis, too.

We must solve this asap because this could introduce other bugs when validating the environment_id in the requests.

### Problem
An organization has an environment with an empty string (`""`) as its ID. This causes Redis keys to be stored without an environment prefix, making them indistinguishable from admin-level keys.

### Approach
Migrate MySQL and Redis separately to avoid data loss:
- **MySQL**: Few thousand rows - can be migrated with a single transaction
- **Redis**: ~1.3M keys - requires a 31-day double-write period before MySQL migration

#### Phase 1: Enable Double-Write (Day 0)
Deploy subscriber with migration enabled:

```
env:
  migration:
    environmentId:
      enabled: "true"
      from: ""
      to: "<new-uuid>"The subscriber will write evaluation/goal counts to **both** old (`ec:...`) and new (`<uuid>:ec:...`) 
```

Redis key formats.

#### Phase 2: Wait Period (Days 1-31)
Keep double-write running for 31 days. This ensures all queryable data (within retention window) exists in both key formats.

#### Phase 3: MySQL Migration (Day 32)
Run SQL migration to update environment ID across all tables:

```
UPDATE environment_v2 SET id = '<new-uuid>' WHERE id = '';
UPDATE feature SET environment_id = '<new-uuid>' WHERE environment_id = '';
-- ... repeat for all tables with environment_id
```

#### Phase 4: Disable Double-Write (Day 32+)

Set `enabled: "false"` and redeploy subscriber. Old Redis keys will expire naturally.

#### Phase 5: Enable Environment ID Validation
Update all APIs to enforce environment ID validation. Previously, validation was skipped to accommodate the empty ID. After migration, enable strict UUID validation for all environment ID parameters across API endpoints.

PR #2341, opened 2026-01-13
chore: implement double-write for empty environment id migration

cre8ivejp:
Part of https://github.com/bucketeer-io/bucketeer/issues/1626

This PR will double-write all events and unique counts for evaluation and goal events for 31 days.

### DIFF
--- a/pkg/subscriber/processor/evaluation_events_evaluation_count_event_persister.go
+++ b/pkg/subscriber/processor/evaluation_events_evaluation_count_event_persister.go
@@ -642,10 +642,8 @@ func (p *evaluationCountEventPersister) writeUserAttributes() {
 func (p *evaluationCountEventPersister) upsertUserAttributes(
 	userAttributes *userproto.UserAttributes,
 ) error {
-	if err := p.userAttributesCacher.Put(
-		userAttributes,
-		time.Duration(p.evaluationCountEventPersisterConfig.UserAttributeKeyTTL)*time.Second,
-	); err != nil {
+	ttl := time.Duration(p.evaluationCountEventPersisterConfig.UserAttributeKeyTTL) * time.Second
+	if err := p.userAttributesCacher.Put(userAttributes, ttl); err != nil {
 		p.logger.Error("Failed to save user attributes to cache",
 			zap.Error(err),
 			zap.String("environmentId", userAttributes.EnvironmentId),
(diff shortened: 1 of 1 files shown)


manifests/bucketeer/charts/subscriber/values.yaml
```diff
@@ -18,6 +18,17 @@ env:
   mysqlPort:
   mysqlDbName:
   mysqlDbOpenConns: 50
+  # Migration settings for environment ID migration
+  # Set these when migrating an environment from an old ID (e.g., empty string) to a new UUID
+  # After 31 days of double-write, set enabled to false and run the MySQL migration
+  migration:
+    environmentId:
+      # Enable the migration double-write (must be explicitly set to "true")
+      enabled: "false"
```
cre8ivejp: The migration is disabled by default.
So, we can enable only in the clusters with an empty environment ID.

Ubisoft-potato (dismissed):
LGTM! Thanks!

pkg/subscriber/processor/migration.go
```diff
@@ -0,0 +1,141 @@
...
+
+var (
+	migrationConfig     *EnvironmentIDMigration
+	migrationConfigOnce sync.Once
+)
+
+// GetEnvironmentIDMigration returns the migration configuration.
+// It reads from environment variables on first call and caches the result.
+// If enabled, it validates that ToEnvironmentID is a valid UUID v4.
+func GetEnvironmentIDMigration() *EnvironmentIDMigration {
```
nnnkkk7: I think this method is used only in this package, so we can make it private.

nnnkkk7 (commented):
Thanks!
I left a few comments.

pkg/subscriber/processor/migration.go
cre8ivejp: @nnnkkk7, fixed https://github.com/bucketeer-io/bucketeer/pull/2341/commits/90d398885f86d1217a7c5eed5135c97933b4c9d4.

@Ubisoft-potato, I agree. At the time I read your comment, I had already changed it to private.
For now, only the `LogMigrationConfig` is public.

Ubisoft-potato: @nnnkkk7 It make sense for me, now let's just make `LogMigrationConfig()` public.

nnnkkk7 (approved):
Thank you!

PR #2347, opened 2026-01-15
test: add e2e test for feature auto archiver

nnnkkk7:
Add e2e test for feature auto archiver job.
I ran the added tests and they all passed.


cre8ivejp (approved):
Thanks!

PR #2352, opened 2026-01-19
refactor: remove command layer in segment pkg

hvn2k1:
Part of https://github.com/bucketeer-io/bucketeer/issues/2073


cre8ivejp:
How about the segment commands in the [command.proto](https://github.com/bucketeer-io/bucketeer/blob/6dfeb4c44f77d6208250f9577f1cf04f6327a49c/proto/feature/command.proto#L186-L214)?

hvn2k1:
> How about the segment commands in the [command.proto](https://github.com/bucketeer-io/bucketeer/blob/6dfeb4c44f77d6208250f9577f1cf04f6327a49c/proto/feature/command.proto#L186-L214)?

@cre8ivejp after remove every command layer in all api in feature package, I'll remove the command.proto file and the https://github.com/bucketeer-io/bucketeer/tree/main/pkg/feature/command file

PR #2363, opened 2026-01-21
fix: segment e2e failed after switch to no command API

hvn2k1:
Root cause:
- In the old delete segment using command pattern, it's the soft delete.
- The API with no command use hard delete hence cause nil pointer in the assertion.

Part of #2073


cre8ivejp (approved):
👍

PR #2365, opened 2026-01-22
fix: resolve tar race condition vulnerability (CVE-2026-23950)

cre8ivejp:
Fix https://github.com/bucketeer-io/bucketeer/security/dependabot/310


Ubisoft-potato (approved):
Thanks! 🙏

PR #2396, opened 2026-02-09
fix(ui/dashboard): fix auto-archive custom days value being reset on modal reopen

nnnkkk7:
## Summary

  - Fix a bug where custom auto-archive days (e.g., 10) were reset to 7 when reopening the environment settings modal.
  - Unify `PRESET_DAYS` as the single source of truth for preset day options.

## Problem

1. Form initial value: `autoArchiveUnusedDays` checked if the saved value was in `PRESET_DAYS` ([7, 14, 30]) and fell back to `7` if not — discarding any custom value
2. Custom input visibility: `isCustomDays` was always initialized as `useState(false)`, so even when the server returned a custom value, the custom input field was never shown


cre8ivejp (approved):
👍

Issue #2397, opened 2026-02-09
fix: default value count not being incremented when using detailed error types


cre8ivejp:
Fix evaluation event count persister so all error reason types (not only deprecated `Reason_CLIENT`) map to the default variation counter.

`getVariationID` only treated `Reason_CLIENT` as an error reason. New error types (`Reason_ERROR_FLAG_NOT_FOUND`, `Reason_ERROR_EXCEPTION`, etc.) were not handled, so events with empty `variation_id` were counted under an empty-string bucket instead of the default variation.

PR #2398, opened 2026-02-09
fix: default value count not being incremented when using detailed error types

cre8ivejp:
Fix https://github.com/bucketeer-io/bucketeer/issues/2397

Fix evaluation event count persister so all error reason types (not only deprecated `Reason_CLIENT`) map to the default variation counter.

Currently, the only SDK using this is the Go SDK from 1.6.0.

## Problem
`getVariationID` only treated `Reason_CLIENT` as an error reason. New error types (`Reason_ERROR_FLAG_NOT_FOUND`, `Reason_ERROR_EXCEPTION`, etc.) were not handled, so events with empty `variation_id` were counted under an empty-string bucket instead of the default variation.

## Solution
- Add `isErrorReason` helper aligned with `grpc_validation.go`, covering all error reason types
- Update `getVariationID` to use `isErrorReason` for deciding when to use the default variation ID

### DIFF
--- a/pkg/api/api/api_grpc.go
+++ b/pkg/api/api/api_grpc.go
@@ -1412,6 +1412,20 @@ func (s *grpcGatewayService) RegisterEvents(
 				continue
 			}
 			evaluationMessages = append(evaluationMessages, event)
+			// Report evaluation events with error reasons for monitoring.
+			if evValidator, ok := validator.(*eventEvaluationValidator); ok &&
+				evValidator.lastUnmarshaledEvent != nil &&
+				isEvaluationEventErrorReason(evValidator.lastUnmarshaledEvent.Reason) {
+				ev := evValidator.lastUnmarshaledEvent
+				evaluationEventErrorReasonCounter.WithLabelValues(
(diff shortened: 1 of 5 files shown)


pkg/subscriber/processor/evaluation_events_evaluation_count_event_persister_test.go
```diff
@@ -375,3 +377,217 @@ func TestCacheUserAttributes(t *testing.T) {
...
+			got := isErrorReason(tt.reason)
+			assert.Equal(t, tt.expectTrue, got, "isErrorReason must stay in sync with grpc_validation.go")
+		})
+	}
+}
+
+// TestIsErrorReasonCoversAllProtoErrorTypes fails when a new error reason type is added to the proto
+// (CLIENT or ERROR_* naming) but isErrorReason in the persister hasn't been updated.
+// This forces us to update isErrorReason and grpc_validation.go when adding new error types.
+func TestIsErrorReasonCoversAllProtoErrorTypes(t *testing.T) {
```
cre8ivejp: Same here.

PR #2400, opened 2026-02-13
chore: remove double error check in get evaluation

hvn2k1:
remove redundant error check


t-kikuc (approved):
Nice!


~Please re-run `pr-go` for sometimes. maybe it's due to a flaky test~

Issue #2411, opened 2026-02-19
docs: add technical document for experiment calculation


hvn2k1:
Document to explain the logic behind the calculation codebase to make it easier to maintain the feature

PR #2412, opened 2026-02-19
docs: add technical explanation for experiment calculation

hvn2k1:
To resolve https://github.com/bucketeer-io/bucketeer/issues/2411 This pr brings:
- Clear explanation on mathematical side
- Mapping the theory with the logic in the code


t-kikuc (commented):
[ask] Are these RFCs? Since they seem like user docs to me, I felt they should be moved to another place

PR #2424, opened 2026-02-25
chore: drop unused mau and mau_summary tables

cre8ivejp:
Fix https://github.com/bucketeer-io/bucketeer/issues/2423


nnnkkk7 (approved):
Thanks!

Issue #2151, opened 2025-10-03
feat: Insights Dashboard


t-kikuc:
## TODO

- [x] RFC #2371 
- [x] Data Generation:
  - [x] Add Prometheus Metrics #2389
  - [x] Add Redis PFADD to Backend (DAU counting) #2391 -> #2437 
  - [x] RDB: Create table #2392
- [x] Batch:
  - [x] MAU: PFMERGE, PFCOUNT, UPSERT (redis+rdb) #2443 
  - [x] Requests: UPSERT (prom+rdb) #2444 
- [x] Backend:
  - [x] Add proto definitions #2390
  - [x] API > Monthly usage + prometheus timeseries x4 #2448 
- [x] Frontend:
  - [x] Graphs
  - [x] CSV Export
- [x] local dev:  Add prometheus to local manifests

### Describe the feature you'd like

Let's display some metrics like the number of requests, MAU, and latency as a dashboard.

### UI Idea 

1. Overview: 
    <img width="977" height="640" alt="Image" src="https://github.com/user-attachments/assets/aed0e150-9063-4048-b0d2-17c7703df71d" />

2. Breakdown for each API
    <img width="977" height="638" alt="Image" src="https://github.com/user-attachments/assets/da190eab-1912-4fb0-ac5d-99bdf0baa17f" />

3. Latency
    <img width="974" height="638" alt="Image" src="https://github.com/user-attachments/assets/109171d9-0c0d-4489-9926-96a5a33f47c2" />

4. Monthly Summary
    <img width="975" height="636" alt="Image" src="https://github.com/user-attachments/assets/eabb0413-c4b4-4f17-a9e5-3bb9ee435bbe" />

### Background

to be described

t-kikuc:
UI idea v2: 

<img width="2880" height="3656" alt="Image" src="https://github.com/user-attachments/assets/819b9217-f49c-4ca8-bf25-8ebc5369a8e5" />

t-kikuc:
[UI] We should add several [i] buttons.


### 1. MAU card 
  `Estimated number of unique users who sent requests to Bucketeer.`
  - The first month may have slight inaccuracies due to the data collection start date.
  - Today's data is not included (aggregated up to the previous day).
  - The number count is based on UTC
  - The number might have around 0.81% errors due to the algorithm.

### 2.   Monthly Requests card 
`Total number of API requests from SDKs processed by Bucketeer.`
[...]

PR #2437, opened 2026-03-03
feat(insights): count DAU in evaluation_count_event_persister

t-kikuc:
Part of #2151 

This fixes #2429 by using the persistent Redis cluster to avoid heavy load on the cache Redis cluster.

### DIFF
--- a/docker-compose/config/subscriber-config/processors.json
+++ b/docker-compose/config/subscriber-config/processors.json
@@ -9,6 +9,7 @@
     "flushSize": 100,
     "flushInterval": 10,
     "writeCacheInterval": 1,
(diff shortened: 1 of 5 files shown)


pkg/cache/v3/mau.go
cre8ivejp: The public function records DAU, but the file name is MAU. It would be better to rename the file. WDYT?

t-kikuc: That's because the file will also contain MAU logic.
If it is confusing, I will fix it.

cre8ivejp (approved):
Nice work!
Thank you!

PR #2463, opened 2026-03-18
feat: obfuscate API key in the UI console

### LINKED ISSUE #1625: chore: improve security in the list api keys api

hvn2k1:
To resolve #1625

This is the preview example of the UI

image

The copy button in the API key page is also removed

**Note: announce the obfuscation to users before deployment**

### DIFF
--- a/ui/dashboard/src/@locales/en/message.json
+++ b/ui/dashboard/src/@locales/en/message.json
@@ -28,10 +28,10 @@
   },
   "copied": "Copied",
   "api-key-created-title": "API key created",
-  "api-key-secret-warning": "This is the only time you can view this full API key. Copy it now and store it somewhere safe — you will not be able to see it again.",
+  "api-key-secret-warning": "Copy and store the API Key. Once you leave this page, the key will be hidden, and you will not be able to see it again.",
   "api-key-secret-label": "Your API key",
   "copy-api-key": "Copy API key",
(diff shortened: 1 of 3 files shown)


ui/dashboard/src/@locales/en/message.json
```diff
@@ -27,6 +27,11 @@
     "name-at-least-characters": "The {{name}} you have provided must have at least {{count}} characters"
   },
   "copied": "Copied",
+  "api-key-created-title": "API key created",
+  "api-key-secret-warning": "This is the only time you can view this full API key. Copy it now and store it somewhere safe — you will not be able to see it again.",
```
cre8ivejp:
```suggestion
  "api-key-secret-warning": "Copy and store the API Key. Once you leave this page, the key will be hidden, and you will not be able to see it again.",
```

Issue #2475, opened 2026-03-24
refactor: replace %v with %w in fmt.Errorf for proper error wrapping


nnnkkk7:
### Summary

Several places in the codebase use `fmt.Errorf("...: %v", err)` instead of `fmt.Errorf("...: %w", err)`. This prevents callers from using `errors.Is()` and `errors.As()` to inspect the underlying error, breaking the standard Go error chain.

### Example

```go
// Before
return nil, fmt.Errorf("malformed jwt: %v", err)

// After
return nil, fmt.Errorf("malformed jwt: %w", err)
```

### Notes

- This is a mechanical change but requires checking each call site to confirm the argument is actually an `error`
- Some `fmt.Errorf` calls use `%v` intentionally for non-error values — leave those as-is
- No behavior change is expected; existing tests should continue to pass

PR #2476, opened 2026-03-24
refactor: replace %v with %w in fmt.Errorf for proper error wrapping

isomorfisma:
For issue: #2475 

This commit updates fmt.Errorf calls to use the %w verb instead of %v when the argument is an error type. This ensures the error chain is preserved, allowing callers to inspect underlying errors using errors.Is() and errors.As().

Note:
- Only modified fmt.Errorf calls where the argument is definitively an error.
- Ignored non-error values (e.g., %v for structs/integers).
- Ignored testing.Fatalf and grpc/status.Errorf as they do not support %w.


t-kikuc (approved):
LGTM
Thank you!

Issue #2473, opened 2026-03-24
chore: remove logging of sensitive information


nnnkkk7:
### Describe the bug

There are some parts where API keys and other sensitive information are being output to the logs, so we want to prevent that for security reasons.
ex):
https://github.com/bucketeer-io/bucketeer/blob/e659999759f97463eb4bc5e1283e4b85c0318d94/pkg/api/api/api_grpc.go#L1444-L1447

### Relevant log output

```bash

```

PR #2495, opened 2026-03-31
chore: remove logging of sensitive information

officialasishkumar:
For issue: #2473

Prevent API keys and refresh tokens from being output to logs in clear text, which poses a security risk.

### Changes

**pkg/api/api/api_grpc.go**
- Obfuscate raw API key values in `Track` request validation and check error logs using `obfuscateString()`
- Obfuscate `ApiKey.Id` in `GetFeatureFlags`, `GetSegmentUsers`, and `RegisterEvents` error logs
- Replace inconsistent `fmt.Sprintf("%s*****", ...)` obfuscation with the standard `obfuscateString()` function
- Obfuscate raw API key in `checkTrackRequest` error log
- Replace `zap.Any("apikey", envAPIKey)` (which logs the entire EnvironmentAPIKey object including the secret) with `zap.String("apiKey", obfuscateString(...))` in both `checkTrackRequest` and `checkRequest`

**pkg/account/api/api_key.go**
- Obfuscate API key value in `GetEnvironmentAPIKey` error log using the existing `obfuscateAPIKey()` function

**pkg/batch/jobs/cacher/apikeycacher.go**
- Replace `zap.Any("envAPIKey", envAPIKey)` (which logs the entire EnvironmentAPIKey object) with individual non-sensitive fields (`apiKeyName`, `environmentId`) using nil-safe proto getters

**pkg/auth/api/api.go**
- Remove raw refresh token string (`req.RefreshToken`) from `RefreshToken` validation and organization fetch error logs
- Remove parsed refresh token object from token generation error log (email is already logged separately)

### DIFF
--- a/pkg/auth/api/api.go
+++ b/pkg/auth/api/api.go
@@ -283,6 +283,7 @@ func (s *authService) RefreshToken(
 	if err := validateRefreshTokenRequest(req); err != nil {
 		s.logger.Error("Failed to validate refresh token request",
 			zap.Error(err),
+			zap.String("refresh_token", obfuscateString(req.RefreshToken)),
 		)
 		return nil, err
 	}
@@ -837,4 +838,15 @@ func (s *authService) PrepareDemoUser() {
 		}
 	}
 	s.logger.Info("Demo environment prepared successfully")
+}
+
+const obfuscateVisibleChars = 4
+
+// obfuscateString obfuscates the input string by showing the first and last
+// obfuscateVisibleChars characters, replacing the middle with dots.
+func obfuscateString(input string) string {
+	if len(input) > obfuscateVisibleChars*2 {
+		return input[:obfuscateVisibleChars] + "...." + input[len(input)-obfuscateVisibleChars:]
+	}
+	return input
 }


pkg/auth/api/api.go
```diff
@@ -283,21 +283,19 @@ func (s *authService) RefreshToken(
...
-		s.logger.Error("Refresh token is invalid", zap.Any("refresh_token", refreshToken))
+		s.logger.Error("Refresh token is invalid", zap.Error(err))
 		return nil, statusUnauthenticated.Err()
 	}
 	organizations, err := s.getOrganizationsByEmail(ctx, refreshToken.Email)
 	if err != nil {
 		s.logger.Error("Failed to get organizations by email",
 			zap.Error(err),
 			zap.String("email", refreshToken.Email),
-			zap.String("refresh_token", req.RefreshToken),
```
hvn2k1: Instead of removing the log, how about use the obfuscateString(req.RefreshToken) function? WDYT, this still hide sensitive info and help with the debugging

pkg/batch/jobs/cacher/apikeycacher.go
```diff
@@ -89,7 +89,9 @@ func (c *apiKeyCacher) putCache(envAPIKey *accproto.EnvironmentAPIKey) int {
 				// Log the error, but do not stop the other goroutines
 				c.logger.Error("Failed to cache environment api key",
 					zap.Error(err),
-					zap.Any("envAPIKey", envAPIKey),
+					zap.String("apiKeyId", obfuscateAPIKey(envAPIKey.GetApiKey().GetId())),
```
hvn2k1: We don't have to hide the ID as it cannot be use to bypass authorization

hvn2k1 (dismissed):
Thank you!

PR #2510, opened 2026-04-06
ci: add hack dir to the dependabot update rules

cre8ivejp:
I'm also ignoring automatic major updates to avoid compatibility issues. This should be done manually after doing proper tests.


Ubisoft-potato (approved):
Thanks！

PR #2517, opened 2026-04-07
feat: support rules reorder

### LINKED ISSUE #2172: feat: support flag custom rule ordering
The UI is ready, but he backend is not.
The user can move up/down the custom rules on the targeting tab to change the order condition, so the user doesn't need to delete and recreate the rules again when ordering.

Ubisoft-potato:
Fixed #2172
                                                                                      
## Summary                                                                                                                                                                                            
  - Add `rule_order` field to `UpdateFeatureRequest` and `ScheduledChangePayload` (proto field 19 / 14)                                                                                                   
  - Wire `ChangeRulesOrder` into the `Update()` domain flow — applied after all rule CREATE/UPDATE/DELETE operations                                                                                      
  - Fix UI swap handler: remove UUID-reassignment hack that caused reorders to be sent as incorrect DELETE+CREATE pairs                                                                                   
  - Add `computeRuleOrder` util and include `ruleOrder` in both immediate update and scheduled change payloads


docker-compose/static-files/js/env.js
```diff
@@ -2,5 +2,6 @@ window.env = {
   "GOOGLE_TAG_MANAGER_ID": "",
   "DEMO_SIGN_IN_ENABLED": true,
   "DEMO_SIGN_IN_EMAIL": "demo@bucketeer.io",
-  "DEMO_SIGN_IN_PASSWORD": "demo"
+  "DEMO_SIGN_IN_PASSWORD": "demo",
+  "SCHEDULED_FLAG_CHANGES_ENABLED": true
```
cre8ivejp: This must be false because it is still a work in progress.

steveninhle (dismissed):
Frontend code looks good! Thank you!

ui/dashboard/src/pages/feature-flag-details/targeting/index.tsx
```diff
@@ -366,19 +366,11 @@ const TargetingPage = ({
...
   const handleSwapSegmentRule = useCallback(
     (indexA: number, indexB: number) => {
-      segmentRulesUpdate(indexA, {
-        ...segmentRulesWatch[indexA],
-        id: uuid()
-      });
-      segmentRulesUpdate(indexB, {
-        ...segmentRulesWatch[indexB],
-        id: uuid()
-      });
```
cre8ivejp, line 376: @steveninhle cc: @Ubisoft-potato 
Is it okay to delete this? It seems it's not related to this PR.

ui/dashboard/yarn.lock
```diff
@@ -1457,7 +1457,7 @@
 
 "@types/luxon@^3.7.1":
   version "3.7.1"
-  resolved "https://registry.yarnpkg.com/@types/luxon/-/luxon-3.7.1.tgz#ef51b960ff86801e4e2de80c68813a96e529d531"
+  resolved "https://registry.npmmirror.com/@types/luxon/-/luxon-3.7.1.tgz#ef51b960ff86801e4e2de80c68813a96e529d531"
```
cre8ivejp: @steveninhle cc:  @Ubisoft-potato 
Can you confirm why we are seeing this diff?

pkg/feature/api/scheduled_feature_change.go
```diff
@@ -962,6 +962,39 @@ func (s *FeatureService) validateScheduledChangePayload(
...
+			case ftproto.ChangeType_DELETE:
+				delete(expectedRuleIDs, rc.Rule.Id)
+			}
+		}
+		if len(payload.RuleOrder) != len(expectedRuleIDs) {
+			return statusInvalidRuleOrder.Err()
+		}
+		for _, id := range payload.RuleOrder {
+			if id == "" {
+				return statusInvalidRuleOrder.Err()
```
cre8ivejp: I think the error message could be confusing when the ID is empty.
Is it possible for the rule ID to be empty at this point?

pkg/feature/domain/feature_update.go
```diff
@@ -728,3 +735,124 @@ func (f *Feature) updateAddVariationToRolloutStrategy(strategy *feature.RolloutS
...
+				return errTagRequired
+			}
+		default:
+			return errUnknownChangeType
+		}
+	}
+	return nil
+}
+
+func validateRuleOrderShape(ruleOrder []string) error {
```
cre8ivejp: To maintain consistency with other validation functions.

```suggestion
func (f *Feature) validateRuleOrderShape(ruleOrder []string) error {
```

pkg/feature/api/scheduled_feature_change.go
```diff
@@ -962,6 +962,39 @@ func (s *FeatureService) validateScheduledChangePayload(
...
+			}
```
cre8ivejp: I think the error message could be confusing when the ID is empty.
I don't think we need to check it.

```suggestion
```

pkg/feature/api/scheduled_feature_change.go
```diff
@@ -962,6 +962,39 @@ func (s *FeatureService) validateScheduledChangePayload(
...
+		expectedRuleIDs := make(map[string]struct{}, len(feature.Rules)+len(payload.RuleChanges))
+		for _, rule := range feature.Rules {
+			if rule != nil && rule.Id != "" {
+				expectedRuleIDs[rule.Id] = struct{}{}
+			}
+		}
+		for _, rc := range payload.RuleChanges {
+			if rc.Rule == nil || rc.Rule.Id == "" {
+				continue
+			}
```
cre8ivejp: It's being checked already [here](https://github.com/bucketeer-io/bucketeer/pull/2517/changes#diff-0ec837ed4ef1a6206b1c998760ed92a62674dcdf6dbc51d478ad23c414bd178bR954-R957).
```suggestion
```

proto/feature/service.proto
```diff
@@ -161,6 +161,7 @@ message UpdateFeatureRequest {
   repeated TargetChange target_changes = 16;
   repeated TagChange tag_changes = 17;
   google.protobuf.StringValue maintainer = 18;
+  repeated string rule_order = 19;
```
cre8ivejp: The current name could be misunderstood as the index rather than the rule ID.
WDYT?

```suggestion
  repeated string ordered_rule_ids = 19;
```

ui/dashboard/yarn.lock
steveninhle: This is the same package & hash, only the registry differs yarnpkg( is the default registry used by yarn) vs npmmirror( is a mirror of the npm registry, commonly used in China for faster). Likely caused by different local registry configs.

Ubisoft-potato: It's my local npm url config: registry.npmmirror.com (a Chinese npm mirror, commonly used in China for speed), yarn updated it when I build fronted, I'll revert it.

proto/feature/service.proto
Ubisoft-potato: Agree! It's ambiguous, let me change the name.

ui/dashboard/src/pages/feature-flag-details/targeting/index.tsx
Ubisoft-potato: The deletion is intentional and related to the PR — the old `id: uuid()` pattern was a workaround that reassigned new IDs on swap, which is incompatible with server-side  
  reorder (the server needs stable rule IDs to know what's being reordered, not replaced). The new `segmentRulesSwap` handles this correctly by preserving IDs. 

cc: @steveninhle, can you please double review it if it not effect other page logic?

pkg/feature/domain/feature_update.go
Ubisoft-potato: Fixed: [478c8e4](https://github.com/bucketeer-io/bucketeer/pull/2517/commits/478c8e49dd252f2acd80991ba51fbdafc8363524)

pkg/feature/api/scheduled_feature_change.go
Ubisoft-potato: Fixed: [3b644e5](https://github.com/bucketeer-io/bucketeer/pull/2517/commits/3b644e52d336ca6dcad9ccd11af615329ae12668)

pkg/feature/api/scheduled_feature_change.go
Ubisoft-potato: Fixed: [3b644e5](https://github.com/bucketeer-io/bucketeer/pull/2517/commits/3b644e52d336ca6dcad9ccd11af615329ae12668)

proto/feature/service.proto
Ubisoft-potato: Fixed: [3b644e5](https://github.com/bucketeer-io/bucketeer/pull/2517/commits/3b644e52d336ca6dcad9ccd11af615329ae12668)

ui/dashboard/src/pages/feature-flag-details/targeting/index.tsx
steveninhle: It's okay for now. Its goal is to force a UI rerender when swapping rule. But RHF's useFieldArray with keyName already regenerates new keys for swapped entries automatically.

cre8ivejp (approved):
Thank you!

PR #2526, opened 2026-04-09
chore: ignore sdk cache not found warning logs

cre8ivejp:
When deploying a service using the server SDK, the first cache will always be empty; logging "cache not found" warning logs is unnecessary.
I'm changing to ignore it.

### DIFF
--- a/pkg/api/api/metrics_event.go
+++ b/pkg/api/api/metrics_event.go
@@ -16,6 +16,7 @@ package api
 
 import (
 	"errors"
+	"strings"
 
 	"github.com/golang/protobuf/ptypes"
 	"go.uber.org/zap"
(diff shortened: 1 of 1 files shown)


Ubisoft-potato (approved):
Thanks!

pkg/api/api/metrics_event.go
```diff
@@ -386,20 +387,28 @@ func (s *grpcGatewayService) saveInternalSdkError(event *eventproto.MetricsEvent
...
-			zap.String("apiId", ev.ApiId.String()),
-			zap.String("sdkVersion", event.SdkVersion),
-			zap.String("sourceId", event.SourceId.String()),
-			zap.String("tag", ev.Labels["tag"]),
-			zap.String("errorMessage", ev.Labels["error_message"]),
-			zap.Any("labels", ev.Labels),
-		)
+		errorMessage := ev.Labels["error_message"]
+		// Skip logging cache not found errors
+		if errorMessage != "" && !isCacheNotFoundError(errorMessage) {
```
t-kikuc: The `errorMessage != ""` condition also suppresses logging for SDK internal errors where `error_message` label is absent or empty. The original code logged all errors when `ev.Labels != nil`. This silently drops errors beyond the stated intent (cache not found only).

```suggestion
		if !isCacheNotFoundError(errorMessage) {
```

PR #2527, opened 2026-04-09
feat: bulk flag duplication across envs

### LINKED ISSUE #2211: feat: bulk flag duplication across multiple environments
## Problem Statement

Currently, when users need to copy a feature flag from one environment (e.g., dev) to other environments (e.g., staging, production), they must manually duplicate the flag to each environment individually. This process is time-consuming and inefficient, especially when the same flag configuration needs to be replicated across multiple environments.

## Current Behavior

- Users must copy flags one environment at a time
- The same flag definition must be manually recreated multiple times
- No way to select multiple target environments for duplication in a single operation

## Desired Behavior

Users should be able to duplicate a feature flag to multiple environments simultaneously through a bulk copy operation. This would allow:

1. Selecting a source flag from one environment (e.g., dev)
2. Selecting multiple target environments (e.g., staging, production) at once
3. Copying the flag configuration to all selected environments in a single action

## Use Case

A common workflow is:
1. Create and test a feature flag in the dev environment
2. Once validated, promote the same flag configuration to staging for further testing
3. Finally, deploy to production with the same configuration

Having to repeat the flag creation process for each environment slows down the deployment pipeline and increases the risk of configuration inconsistencies.

### Background

_No response_

### Alternative solutions

_No response_

Ubisoft-potato:
Part of https://github.com/bucketeer-io/bucketeer/issues/2211

## Summary

Implements `BulkCloneFeature` RPC (`POST /v1/feature/bulk_clone`) to resolve #2211 — allowing users to copy a feature flag to multiple target environments in a single API call.

- **Proto**: Added `BulkCloneFeatureRequest`, `BulkCloneFeatureResult`, `BulkCloneFeatureResponse`, and `BulkCloneFeatureStatus` enum; registered `BulkCloneFeature` RPC with HTTP gateway binding
- **Handler** (`pkg/feature/api/feature.go`): Checks EDITOR permission on **all** target environments upfront before cloning any; iterates targets, clones each independently, collects per-env results
- **Validation** (`pkg/feature/api/validation.go`): Rejects empty id/environment_id, empty target list, duplicate targets, or source == target
- **Error codes** (`pkg/feature/api/error.go`): Added `statusMissingEnvironmentID`, `statusMissingTargetEnvironments`, `statusDuplicateTargetEnvironments`
- **Mock** (`pkg/feature/client/mock/client.go`): Regenerated to include new method

## Behavior design

| Scenario | Behavior |
|---|---|
| Permission denied on any target env | Abort entire request — no cloning happens |
| Flag already exists in a target env | Skip that env, continue others, report `ALREADY_EXISTS` |
| DB error on a target env | Skip that env, continue others, report `FAILED` with message |
| Prerequisites | Cleared to `[]` — variation IDs of other flags do not transfer across envs |
| Rules referencing feature flags or segments | Entire rule dropped — those entities may not exist in the target env |

Each call to `Clone()` regenerates variation UUIDs independently per target environment and remaps all internal references (off-variation, targets, default strategy, rule strategies).

Closes #2211


cre8ivejp (approved):
Thank you! Nice improvement!

PR #2535, opened 2026-04-14
fix(insights): exclude zero-request SDKs from source ID filter

t-kikuc:
Follows up on #2530. The batch job upserts `monthly_summary` rows for every SDK even when there is no traffic (`mau=0, request_count=0`), so the previous row-existence check showed all SDKs regardless of actual usage.

- Filter `availableSourceIds` by checking `requests > 0` in at least one month instead of just checking series existence

### DIFF
--- a/ui/dashboard/src/pages/insights/page-loader.tsx
+++ b/ui/dashboard/src/pages/insights/page-loader.tsx
@@ -88,8 +88,8 @@ const PageLoader = () => {
       enabled: hasEnvironments
     });
 
-  // Discover which SDKs have data by fetching monthly summary for all source IDs.
-  // The DB only returns rows for SDKs that actually have records.
+  // Discover which SDKs have actual traffic by fetching monthly summary for
+  // all source IDs and filtering out zero-request entries.
   // Using ALL_SOURCE_IDS also lets React Query share the cache when the SDK
   // filter is set to "All", avoiding a duplicate request.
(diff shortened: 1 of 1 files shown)


cre8ivejp (approved):
Thank you!

steveninhle (commented):
I found an issue with the SDK filter not resetting when switching environments

**Steps to reproduce:**

1. Open the Insights dashboard
2. Select an environment that has data for a specific SDK (e.g. go)
3. Select that SDK from the SDK filter dropdown
4. Switch to a different environment that has no data for that SDK

**Expected behavior:**

-    The SDK filter resets to "All", the URL updates accordingly, and API requests use   sourceId=ALL.

**Actual behavior:**

The SDK dropdown visually shows only "All" (since the selected SDK has no data in the new environment and is filtered out of the options), but:
- The URL still contains the old sourceId (e.g. sourceId=go)
- The filter state still holds the stale sourceId
- API requests still use the old SDK as a parameter
Suggest a fix:
ui/dashboard/src/pages/insights/page-content.tsx
   ```
         > const handleEnvironmentChange = useCallback(
            >     (value: string) => {
            >       onFiltersChange({ ...filters, environmentId: value, sourceId: ALL });
            >     },
            >     [filters, onFiltersChange]
            >   );
```

t-kikuc:
@steveninhle 
Thanks, could you please open another new PR for that?

PR #2556, opened 2026-04-24
refactor: use isCallerContextErr consistently at all metric/log sites

cre8ivejp:
## Summary

Follow-up to the cascading-cancellation fix in #<original-PR>. After watching
production logs, a few sites still emitted noisy ERROR logs and a few
predicate inconsistencies were brittle. This PR cleans those up without
changing the core fix's behavior.

### 1. Suppress noisy ERROR logs for caller-cancel propagation

The original PR added `errors.Is(err, errCallerCanceled)` branches at the
public-API entry points so caller-cancellation no longer bumps
`codeInternalError` / `apiErrorCounter`. But several intermediate helpers
were still logging the propagated error at ERROR severity, polluting the
log stream even though the metric attribution was already correct.

Fixed at three sites:

- `getSegmentUsersMap` — `"Failed to list segments"` was firing on every
  SDK disconnect mid-evaluation. Now suppressed when err matches
  `isCallerContextErr`.
- `checkTrackRequest` — `"Failed to get environment API key"` was firing
  on every Track-path SDK disconnect. **Removed entirely** rather than
  suppressed: the public `Track` API already logs the propagated error
  with full suppression, so the helper-level log was double-logging on
  any non-cancel failure too. This brings `checkTrackRequest` in line
  with `checkRequest`, which has never logged this error.
- `Track` public API — `"Failed to check Track request"` had no
  suppression chain. Brought into the same
  `!isCallerContextErr(err) && !errors.Is(err, ErrInvalidAPIKey) &&
  !errors.Is(err, ErrMissingAPIKey)` pattern the other 5 public APIs use.

### 2. Preserve `DeadlineExceeded` in handler early-out

`checkRequest` and `checkTrackRequest` open with a fast-path early return
for already-terminated contexts. The previous helper `isContextCanceled`
matched only `context.Canceled`, so deadline-expired requests fell through
into the work and lost the correct gRPC status code at the early-out.

Replaced with `ctxAlreadyDoneErr(ctx) error` that returns the right
sentinel for both terminations:

| `ctx.Err()` | returns |
|---|---|
| `context.Canceled` | `ErrContextCanceled` (`codes.Canceled`) |
| `context.DeadlineExceeded` | `ErrContextDeadlineExceeded` (`codes.DeadlineExceeded`) |
| nil (live ctx) | `nil` |

`isContextCanceled` is left intact because the legacy REST `gatewayService`
in `api.go` (deprecated, no production callers) still uses it with its
own `errContextCanceled` sentinel — out of scope for this PR.

### 3. Unify caller-context error detection on `isCallerContextErr`

Across the file there were two predicates being used to detect "this error
is a caller-context termination":

- `errors.Is(err, errCallerCanceled)` — narrow; matches only the raw
  wrapper produced by `singleflightFetch`.
- `isCallerContextErr(err)` — broad; matches the wrapper plus the two
  translated public-facing sentinels (`ErrContextCanceled`,
  `ErrContextDeadlineExceeded`).

The narrow check was used at 7 metric-attribution sites (cache-fetch error
paths in 4 public APIs); the broad check at 6 log-suppression sites (auth
paths). Today both are correct given the exact error-propagation paths,
but the split is brittle — if anyone ever adds a translation step on a
cache-fetch path, the narrow check silently misses caller-cancel and
starts bumping `codeInternalError` again, exactly the regression the
original PR fixes.

`isCallerContextErr` is a strict superset, so unifying on it preserves
all current behavior and future-proofs the metric attribution. All 7
metric-attribution sites converted; the 2 narrow checks remaining are in
`isCallerContextErr`'s own definition and inside
`translateCallerCanceledErr` (which by design only translates the raw
wrapper).

## Tests

- Updated `TestIsCallerContextErr` (already added in the original PR) with
  two new cases for the raw `errCallerCanceled` wrapper to lock in the
  superset semantic.
- New `TestCtxAlreadyDoneErr` — table-driven, 3 cases covering live ctx,
  canceled ctx, and deadline-expired ctx.

All previously added tests for the singleflight helper continue to pass:

- `TestSingleflightFetch` (5 cases)
- `TestSingleflightFetchDetachesCallerCancellation`
- `TestSingleflightFetchCollapsesConcurrentCallers` (with the documented
  scheduler caveat)
- `TestTranslateCallerCanceledErr` (4 cases)

## Behavior change matrix

| Scenario | Before this PR | After this PR |
|---|---|---|
| SDK disconnects mid-evaluation while waiting on segment users | ERROR log `"Failed to list segments"` per disconnect | No log (metric already correct) |
| SDK disconnects mid-Track | Two ERROR logs (`"Failed to get environment API key"` + `"Failed to check Track request"`) per disconnect | No log |
| Real downstream failure on Track path | Two ERROR logs | One ERROR log (Track only) |
| Caller deadline expires before handler runs | Returns `codes.Canceled` (wrong) | Returns `codes.DeadlineExceeded` |
| Future code adds a translation on a cache-fetch error path | Metric attribution silently regresses to `codeInternalError` | Metric attribution stays correct via `isCallerContextErr` superset |

## Test plan

- [x] `go build ./pkg/api/api/...`
- [x] `go vet ./pkg/api/api/...` (no new warnings)
- [x] `go test ./pkg/api/api/... -count=1`
- [x] `go test ./pkg/api/api/... -run 'TestSingleflightFetch|TestCtxAlreadyDoneErr|TestIsCallerContextErr|TestTranslateCallerCanceledErr' -count=10 -race`

### DIFF
--- a/pkg/api/api/api_grpc.go
+++ b/pkg/api/api/api_grpc.go
@@ -316,17 +316,19 @@ func (s *grpcGatewayService) Track(ctx context.Context, req *gwproto.TrackReques
 	}
 	envAPIKey, err := s.checkTrackRequest(ctx, req.Apikey)
 	if err != nil {
-		s.logger.Error("Failed to check Track request",
-			log.FieldsFromIncomingContext(ctx).AddFields(
-				zap.Error(err),
-				zap.String("apiKey", obfuscateString(req.Apikey, obfuscateAPIKeyLength)),
-				zap.String("tag", req.Tag),
-				zap.Any("userId", req.Userid),
-				zap.String("goalId", req.Goalid),
-				zap.Int64("timestamp", req.Timestamp),
-				zap.Float64("value", req.Value),
-			)...,
-		)
+		if !isCallerContextErr(err) && !errors.Is(err, ErrInvalidAPIKey) && !errors.Is(err, ErrMissingAPIKey) {
+			s.logger.Error("Failed to check Track request",
+				log.FieldsFromIncomingContext(ctx).AddFields(
+					zap.Error(err),
+					zap.String("apiKey", obfuscateString(req.Apikey, obfuscateAPIKeyLength)),
+					zap.String("tag", req.Tag),
+					zap.Any("userId", req.Userid),
+					zap.String("goalId", req.Goalid),
+					zap.Int64("timestamp", req.Timestamp),
+					zap.Float64("value", req.Value),
+				)...,
+			)
+		}
 		return nil, err
 	}
 	requestTotal.WithLabelValues(
@@ -465,7 +467,7 @@ func (s *grpcGatewayService) GetEvaluations(
 	})
 	spanGetFeatures.End()
 	if err != nil {
-		if errors.Is(err, errCallerCanceled) {
+		if isCallerContextErr(err) {
 			evaluationsCounter.WithLabelValues(
 				environmentId, envAPIKey.Environment.UrlCode, req.Tag, codeCanceled, sourceID).Inc()
 			return nil, status.FromContextError(ctx.Err()).Err()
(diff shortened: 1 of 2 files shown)


Ubisoft-potato (approved):
Thanks!

PR #2571, opened 2026-05-11
fix(ui): replace nested button with span in tooltip triggers on create feature flag

steveninhle:
Fixed console error at create feature flag:
```
installHook.js:1 <button> cannot contain a nested <button>.
See this log for the ancestor stack trace.
```


cre8ivejp (approved):
Thank you!

PR #2573, opened 2026-05-12
fix(ui): use variation name as label in evaluation chart for YAML/JSON type

steveninhle:
## Summary

- Evaluation chart legend was showing raw YAML/JSON blob content as the
  variation label, making it unreadable for those variation types
- Fix: resolve variation label using `variation.name || variation.value`,
  which prefers the human-readable name when set and falls back to value
  otherwise
- This aligns the chart legend with the evaluation table, which already
  uses the same `name || value` fallback pattern


cre8ivejp (approved):
Thank you!

Issue #2499, opened 2026-04-01
feat: support json schema for flag variations


cre8ivejp:
## Summary

Support schema-based validation for variation values based on the variable type, and validate values before creating or updating a variation.

Supported schema types:

- **Enum** for **String** or **Number** variables
- **Regex** for **String** variables
- **JSON** for **JSON** variables

If validation fails, the variation must not be saved, and the UI should display validation errors so the user can resolve them before saving.

## How this feature would work

When a user edits a variation, the system validates the value against the schema defined for that flag/variable before saving.

### Supported schema types

- **Enum**: supported for **String** and **Number** variables
- **Regex**: supported for **String** variables
- **JSON Schema**: supported for **JSON** variables

### Example: JSON variation validation

When a user edits a JSON variation, for example:

```json
{
  "theme": "dark",
  "maxItems": 20,
  "showPromo": true
}
```

the system validates the value against the schema defined for that flag/variable, such as:

```json
{
  "type": "object",
  "properties": {
    "theme": {
      "type": "string",
      "enum": ["light", "dark"]
    },
    "maxItems": {
      "type": "integer",
      "minimum": 1,
      "maximum": 100
    },
    "showPromo": {
      "type": "boolean"
    }
  },
  "required": ["theme", "maxItems"],
  "additionalProperties": false
}
```

Before saving, the system validates that:

- required fields are present
- field types are correct
- enum values are allowed
- additional properties are rejected when not allowed by the schema

If the user enters an invalid value, for example:

```json
{
  "theme": "blue",
  "maxItems": "20",
  "extraField": "x"
}
```

The UI should display validation errors such as:

- `theme` must be one of: `light`, `dark`
- `maxItems` must be an integer
- `extraField` is not allowed

The user should not be able to create or update the variation until all validation errors are resolved.

## Expected behavior

- Variation values can be validated against a defined schema based on the variable type.
- Validation is performed before creating or updating a variation.
- Supported schema types are:
  - **Enum** for **String** or **Number** variables
  - **Regex** for **String** variables
  - **JSON Schema** for **JSON** variables
- If validation fails, the variation cannot be saved.
- Validation errors are shown in the UI so the user can correct them before saving.

## Notes

- Validation should run before both **create** and **update** actions.
- Error messages should clearly indicate what is invalid.
- For JSON variables, validation errors should point to the relevant field when possible.

pradhankukiran:
[...]

PR #2578, opened 2026-05-14
feat: add variation value schema persistence

pradhankukiran:
Adds the variation value schema model and persistence plumbing for #2499.

This PR does not implement validation behavior yet. It only introduces the schema shape across proto/API/storage layers so the backend validation PR can build on top of it.

  ## Changes

  - Add `VariationValueSchema` to feature protos
  - Add create/update API fields for setting and clearing the schema
  - Persist the schema on features in MySQL and PostgreSQL
  - Update feature storage SQL queries and scans
  - Add DB migrations and update Atlas checksums
  - Update generated Go/OpenAPI artifacts
  - Add dashboard TypeScript types for the new schema field


pkg/feature/api/feature.go
```diff
@@ -783,6 +785,18 @@ func (s *FeatureService) updateFeatureWithinTransaction(
...
+		schemaChanged = updated.VariationValueSchema != nil
+		updated.VariationValueSchema = nil
+	} else if req.VariationValueSchema != nil {
+		schemaChanged = !goproto.Equal(updated.VariationValueSchema, req.VariationValueSchema)
+		updated.VariationValueSchema = req.VariationValueSchema
+	}
+	if schemaChanged && updated.Version == feature.Version {
+		updated.Version++
+		updated.UpdatedAt = time.Now().Unix()
+	}
```
cre8ivejp: Changes to the flag object should be made inside [feature.Update](https://github.com/bucketeer-io/bucketeer/pull/2578/changes#diff-2d0a2c8153246509abb80eca89b6ca25f95f7ea8e75640dae22408489b8b0eafR768) to maintain consistency.

pkg/feature/domain/feature_update.go
```diff
@@ -43,6 +48,7 @@ func (f *Feature) Update(
 	tagChanges []*feature.TagChange,
 	maintainer *wrapperspb.StringValue,
 	ruleOrder []string,
+	variationValueSchemaUpdates ...*VariationValueSchemaUpdate,
```
Ubisoft-potato: The variadic `...*VariationValueSchemaUpdate` parameter is misleading — `applyVariationValueSchemaUpdate` only ever reads `updates[0]` and silently drops anything beyond that. Variadic signals "you can pass many and all will be processed," but that's not what happens here.

A plain nullable pointer is the correct type:

```go
// instead of:
variationValueSchemaUpdates ...*VariationValueSchemaUpdate

// use:
schemaUpdate *VariationValueSchemaUpdate
```

Then simplify the apply function accordingly:

```go
func (f *Feature) applyVariationValueSchemaUpdate(update *VariationValueSchemaUpdate) {
    if update == nil {
        return
    }
    if update.Clear.GetValue() {
        f.VariationValueSchema = nil
        return
    }
    if update.Schema != nil {
        f.VariationValueSchema = update.Schema
    }
}
```

The call site in `pkg/feature/api/feature.go` already passes a single `&domain.VariationValueSchemaUpdate{...}`, so no change is needed there.

pkg/feature/storage/v2/mysql/feature.go
```diff
@@ -60,10 +63,54 @@ type featureStorage struct {
...
+}
+
 func NewFeatureStorage(qe mysqlstorage.QueryExecer) v2fs.FeatureStorage {
 	return &featureStorage{qe: qe}
 }
 
+func variationValueSchemaJSONValue(schema *proto.VariationValueSchema) interface{} {
+	if schema == nil {
+		return nil
+	}
```
cre8ivejp: We should check for a zero-value {} schema to avoid becoming two distinct states.
Please update the unit tests to cover both cases and for PostgreSQL storage.

E.g.
```suggestion
	if schema == nil || goproto.Equal(schema, &proto.VariationValueSchema{}) {
		return nil
	}
```

cre8ivejp:
@pradhankukiran, can you check the comments when you have time?
Once you push the changes, we can merge.

docker-compose/init-db/mysql_dump.sql
```diff
@@ -87,6 +87,7 @@ CREATE TABLE `feature` (
   `maintainer` varchar(255) CHARACTER SET utf8mb4 COLLATE utf8mb4_bin NOT NULL,
   `environment_namespace` varchar(255) CHARACTER SET utf8mb4 COLLATE utf8mb4_bin NOT NULL,
   `variation_type` int NOT NULL,
+  `variation_value_schema` json DEFAULT NULL,
```
Ubisoft-potato: This line must be removed. 
The init-db dump is the initial project DB schema and should not be modified. The `variation_value_schema` column is already added by the migration file `migration/mysql/20260514000000_update_feature_variation_value_schema.sql`. 

Having it in both places will cause the Docker Compose migration service to fail on a fresh setup with a duplicate column error, since the dump creates the table with the column already present before Atlas runs the migration.

image

docker-compose/init-db/postgres_dump.sql
```diff
@@ -145,6 +145,7 @@ CREATE TABLE feature (
     tags JSONB NOT NULL,
     maintainer VARCHAR(255) NOT NULL,
     variation_type INTEGER NOT NULL,
+    variation_value_schema JSONB,
```
Ubisoft-potato: This line must be removed. The init-db dump is the initial project DB schema and should not be modified. The `variation_value_schema` column is already added by the migration file `migration/postgres/20260514000000_update_feature_variation_value_schema.sql`. Having it in both places will cause the Docker Compose migration service to fail on a fresh setup with a duplicate column error, since the dump creates the table with the column already present before Atlas runs the migration.

Ubisoft-potato (commented):
LGTM!
Thanks for contributing.

cre8ivejp (approved):
Thank you so much!

Issue #2306, opened 2025-12-11
feat: support for PostgreSQL as database option for storage layer


hvn2k1:
### Describe the feature you'd like

Implement PostgreSQL as a main database option, support all the functional queries in console

hvn2k1:
These packages will be updated as they have storage implementation to query the database:
- [x] account
- [x] auditlog
- [x] autoops
- [x] coderef
- [x] environment
- [x] eventcounter
- [x] experiment
- [x] experiment calculator
- [x] feature
- [x] feature last used info
- [x] flag trigger
- [x] insights
- [x] notification
- [x] opsevent
- [x] push
- [x] tag
- [x] team
- [x] scheduled flag change
- [x] segment
- [x] segment user
- [x] persister

hvn2k1:
All the packages are implemented, the postgres operational database alternative is ready to be tested

PR #2583, opened 2026-05-18
feat: implement postgres for tag svc

hvn2k1:
Part of #2306

Summary
- Refactored tag storage layer to support both MySQL and PostgreSQL by extracting the monolithic pkg/tag/storage implementation into DB-specific sub-packages (mysql/ and postgres/)
- Made TagStorage interface DB-agnostic with ListTagsParams replacing MySQL-specific ListOptions in the interface signature
- Updated all tag API consumers (tag/api, feature/api, account/api, batch/jobs/deleter) to accept injected TagStorage instead of creating it inline from mysqlClient
- Replaced mysqlClient.RunInTransactionV2 with dbClient.RunInTransactionV2 in TagService for DB-agnostic transactions
- Wired postgres/mysql tag storage branching in both web/cmd/server and batch/cmd/server based on operationalDatabaseType flag
- Added unit tests for the PostgreSQL tag storage implementation


pkg/tag/storage/tag.go
```diff
@@ -17,33 +17,21 @@ package storage
...
-	countTagsSQL string
-	//go:embed sql/delete_tag.sql
-	deleteTagSQL string
+	ErrTagNotFound               = pkgerr.NewErrorNotFound(pkgerr.TagPackageName, "tag not found", "tag")
+	ErrTagUnexpectedAffectedRows = pkgerr.NewErrorUnexpectedAffectedRows(
+		pkgerr.TagPackageName,
+		"tag unexpected affected rows",
+	)
+	ErrInvalidListTagsCursor  = errors.New("tag storage: invalid list tags cursor")
+	ErrInvalidListTagsOrderBy = errors.New("tag storage: invalid list tags order by")
```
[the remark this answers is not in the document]
hvn2k1: Updated! https://github.com/bucketeer-io/bucketeer/pull/2583/commits/cbdf8e1730d6d59120cf71b460e2e3a697f7ddc8

pkg/feature/api/tag.go
```diff
@@ -89,6 +73,12 @@ func (s *FeatureService) ListTags(
 				zap.String("environmentId", req.EnvironmentId),
 			)...,
 		)
+		if errors.Is(err, tagstorage.ErrInvalidListTagsCursor) {
```
Ubisoft-potato: Since `s.logger.Error("Failed to list tags", ...)` fires before this check, a client sending an invalid cursor like `"foo"` will generate an ERROR-level log entry — which could look like a real storage failure and trigger false alerts.

The old code caught this with `strconv.Atoi` before calling storage and returned `statusInvalidCursor` quietly. Worth doing the same here so ops noise stays clean:

```go
cursor := req.Cursor
if cursor == "" {
    cursor = "0"
}
if _, err := strconv.Atoi(cursor); err != nil {
    return nil, statusInvalidCursor.Err()
}
```

With that in place, the `ErrInvalidListTagsCursor` branch below becomes unreachable and can be dropped.

pkg/tag/api/api.go
```diff
@@ -311,6 +256,12 @@ func (s *TagService) ListTags(
 				zap.String("environmentId", req.EnvironmentId),
 			)...,
 		)
+		if errors.Is(err, tagstorage.ErrInvalidListTagsCursor) {
```
Ubisoft-potato: Same situation as `pkg/feature/api/tag.go` — the `s.logger.Error("Failed to list tags", ...)` on line 253 runs before this check, so an invalid cursor from a client ends up as an ERROR log. Worth validating the cursor before calling `tagStorage.ListTags` to avoid that noise.

pkg/feature/api/tag.go
hvn2k1: https://github.com/bucketeer-io/bucketeer/pull/2583/commits/adbfe759f67824c09bbc9cbed49b388f016c2161 nice catch, I updated to use the correct error

pkg/tag/api/api.go
hvn2k1: https://github.com/bucketeer-io/bucketeer/pull/2583/commits/adbfe759f67824c09bbc9cbed49b388f016c2161 ⬆️

Ubisoft-potato (approved):
The rest LGTM! 👍

PR #2585, opened 2026-05-19
fix(insights): include `UNKNOWN` sdk in monthly usages

t-kikuc:
When calculating the request counts for billing, the `UNKNOWN` SDK should also be included.
So the insights dashboard should show the req counts of `UNKNOWN` too.

After releasing this fix, the count for this month will be updated correctly; the previous months will not be updated.


hvn2k1 (approved):
Thank you!

PR #2591, opened 2026-05-22
fix: improve Codespaces dev container setup

Rtosshy:
## Summary

Add Docker-in-Docker support to the Codespaces dev container to fix container-to-container communication issues when running `make docker-compose-up` and `make start-minikube`.

Also adds a post-create setup script to initialize Codespaces permissions for Go cache, the home directory, and MySQL config files.

## Changes
- Add the Docker-in-Docker dev container feature
- Add the SSHD dev container feature
- Add a post-create setup script for Codespaces permission initialization
- Update MySQL config file permissions during post-create setup

### Docker-in-Docker

Without Docker-in-Docker, Docker workloads started from the Codespaces dev container could not use the expected container networking behavior.

For `make docker-compose-up`, this caused container-to-container communication failures.

For `make start-minikube`, this caused Minikube addons to fail because required services were not reachable.

Docker-in-Docker makes the Docker daemon and its container networks run inside the Codespaces environment, which provides the expected networking behavior for these workflows.

### SSHD

SSHD is added so developers can connect to Codespaces with `gh cs ssh` and work directly inside the dev container, for example by using Neovim or other terminal-based tools installed in the environment.

### Permission initialization

Codespaces-mounted files can have permissions that are not suitable for some tools running inside containers.

MySQL ignores world-writable config files, so the post-create setup updates the MySQL config directory permissions to make sure local MySQL settings are loaded correctly.

The setup also initializes ownership for the Go package cache and the Codespaces user's home directory so development tools can write to them without permission errors.

## How to verify

- Rebuild the Codespaces dev container
- Run `make docker-compose-up`
- Run `make start-minikube`

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -15,7 +15,12 @@
   "features": {
     "ghcr.io/devcontainers/features/sshd:1": {
       "version": "latest"
-    }
+    },
+    "ghcr.io/devcontainers/features/docker-in-docker:2": {
+        "version": "latest",
+        "moby": true,
+        "enableNonRootDocker": true
+    },
   },
   "mounts": [
     // Persistent Go modules cache
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -16,11 +16,6 @@
     "ghcr.io/devcontainers/features/sshd:1": {
       "version": "latest"
     },
-    "ghcr.io/devcontainers/features/docker-in-docker:2": {
-        "version": "latest",
-        "moby": true,
-        "enableNonRootDocker": true
-    },
   },
   "mounts": [
     // Persistent Go modules cache
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -86,6 +86,6 @@
     "memory": "8gb",
     "storage": "32gb"
   },
-  "postCreateCommand": "bash .devcontainer/post-create-command.sh",
+  "postCreateCommand": "bash .devcontainer/setup-permissions.sh",
   "postAttachCommand": "bash .devcontainer/setup.sh"
 }
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -12,6 +12,11 @@
   "runArgs": [
     "--cgroupns=host"
   ],
+  "features": {
+    "ghcr.io/devcontainers/features/sshd:1": {
+      "version": "latest"
+    }
+  },
   "mounts": [
     // Persistent Go modules cache
     "source=bucketeer-go-mod-cache,target=/go/pkg/mod,type=volume",
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -86,5 +86,6 @@
     "memory": "8gb",
     "storage": "32gb"
   },
+  "postCreateCommand": "bash .devcontainer/post-create-command.sh",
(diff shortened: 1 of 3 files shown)


Ubisoft-potato:
Hey @Rtosshy, thanks for the PR!

Two things to address:

**1. `docker-in-docker` feature is not needed**
Docker is already installed in our base image (`.github/.devcontainer/Dockerfile`) and the daemon is started automatically at the end of `setup.sh` (the `postAttachCommand`):
```bash
nohup sudo dockerd > /tmp/dockerd.log 2>&1 < /dev/null &
```
Adding the `docker-in-docker` devcontainer feature would download and install Docker a second time on top of an image that already has it, adding unnecessary Codespaces startup time. Docker should already be working — if you hit networking issues, it's worth checking whether `setup.sh` completed successfully and `dockerd` is running.

**2. Features should go in `.github/.devcontainer/devcontainer.json`**
`.github/.devcontainer/` is the canonical devcontainer configuration for this project. Features declared there get baked into the pre-built image that is pushed to `ghcr.io/bucketeer-io/bucketeer-devcontainer:latest` via our CI workflow. That image is then pulled by `.devcontainer/devcontainer.json` as its base.

Adding features to `.devcontainer/devcontainer.json` means they are downloaded and installed at Codespaces runtime on every new environment, which defeats the purpose of the pre-built image. Any devcontainer features like SSHD should be added to `.github/.devcontainer/devcontainer.json` so they are baked in once and reused.

Rtosshy:
Hi! @Ubisoft-potato, thanks for the detailed review!

> 1. `docker-in-docker` feature is not needed

I understand that Docker is already installed in the base image and that dockerd is expected to be started by setup.sh. However, in my environment, after connecting to the Codespace with `gh cs ssh`, `setup.sh` does not seem to run automatically. I also tried running `setup.sh` manually after SSHing into the container, but even after that, both make docker-compose-up and make start-minikube still fail.

I haven’t fully identified the difference yet between the current pre-built image setup and the behavior when using the docker-in-docker feature, so I’ll keep investigating this before deciding whether the feature is actually necessary.

> 2. Features should go in `.github/.devcontainer/devcontainer.json`

Thank you for the clarification. I agree that putting features in .github/.devcontainer/devcontainer.json is the right direction for this project, since they can be baked into the pre-built image.

However, I’m not fully sure yet whether simply moving the SSHD feature there will make the CI correctly include the SSH functionality in the pre-built image. So I’ll verify this on a separate branch first, and then open a PR once I confirm the expected behavior.

Ubisoft-potato:
Hi @Rtosshy!

1. You can also try start dev container in your local machine using VSCode, it will be more convenient for debuing.

2. after add feature to `.github/.devcontainer` you can run the pre-built CI workflow, usually that's should be fine, since we just add a package to this image.

cre8ivejp:
@Rtosshy, hi!
Are there any updates on the changes that need to be made?

Rtosshy:
@Ubisoft-potato @cre8ivejp 
Hi, sorry for the long delay.

For now, I’m going to stop working on this PR and close it.

I’d like to separate the two concerns:

1. For SSH support, I will first verify that adding the SSHD feature to `.github/.devcontainer/devcontainer.json` works correctly with the pre-built image and `gh cs ssh`. Once I have confirmed the behavior, I will submit a separate PR focused only on SSHD support.

2. For the Docker Compose and Minikube issues, I will investigate them separately instead of trying to solve them by adding the `docker-in-docker` feature in this PR.

Thank you for the reviews and for following up.

Issue #2150, opened 2025-10-03
feat: add ai recommend chat


nnnkkk7:
#### Problem
Bucketeer is a feature-rich platform offering feature flags, A/B testing, segments, progressive rollouts, flag triggers, and more. However, user research indicates that many powerful features remain underutilized:
- Feature Discovery Problem: Users often don't know about features like Flag Triggers, Progressive Rollout, or Prerequisite Rules until they encounter specific pain points
- Documentation Fatigue: Users rarely have time to read comprehensive documentation to discover relevant features
- Context-Sensitive Needs: Users need guidance that's relevant to their current task, not generic feature lists

#### What we’d like
We will implement an AI Chat Assistant feature that helps users discover and utilize Bucketeer's powerful but underused features. The assistant provides contextual suggestions and guidance through a conversational interface, improving feature discoverability without requiring users to read extensive documentation.

#### Notes
- We should estimate the costs of ai api at first.

PR #2597, opened 2026-05-29
feat: map synonym to improve correction

hvn2k1:
An extension of https://github.com/bucketeer-io/bucketeer/issues/2150

RAG keyword search misses docs when users phrase questions with non-canonical terms (e.g. "A/B test" vs. the `experimentation/` folder, "webhook" vs. `trigger.mdx`). The LLM then refuses with "the provided reference documents do not contain information…".

Adds a small synonym map applied to query tokens before scoring.

**Before:**
image

**After:**

https://github.com/user-attachments/assets/d0f54dfb-7e7b-40ae-9a31-e7ed1f161f7b


hvn2k1:
The branch was checked out from the old pr so the commits are messed up

PR #2601, opened 2026-05-29
fix: system admin can't see other envs if it's a member

### LINKED ISSUE #2600: fix: system admin can't see other environments or projects if it's a member
### Version

v2.2.0

### Describe the bug

This error happened to GetMe API

### What we are having in the GetMe response:
* System admin account can only see the projects and environments that the account is added, and can write if that account has permission

* We have a non-system admin organization, have multiple projects
* System admin account is also added to it as a a member on specific environments, other environments are not added

### Expected behavior
* System admin account can still see every projects and environments inside that organization
* System admin account can only write on specific environments that account has editor member (or that account become the organization admin)

hvn2k1:
To resolve https://github.com/bucketeer-io/bucketeer/issues/2600

Fixed by switch the logic for system admin in case it's a member, show every environment in the org (system admins always have read access), but apply the explicit member role where one exists so write permissions are respected.

### DIFF
--- a/pkg/account/api/admin_account.go
+++ b/pkg/account/api/admin_account.go
@@ -153,7 +153,12 @@ func (s *AccountService) GetMe(
 			if orgAccount.OrganizationRole >= accountproto.AccountV2_Role_Organization_ADMIN {
 				envRoles = s.getAdminConsoleAccountEnvironmentRoles(environments, projects)
 			} else {
-				envRoles = s.getConsoleAccountEnvironmentRoles(orgAccount.EnvironmentRoles, environments, projects)
+				// System admin who joined as a non-admin member: show every environment in
+				// the org (system admins always have read access), but apply the explicit
+				// member role where one exists so write permissions are respected.
+				envRoles = s.getSystemAdminMemberEnvironmentRoles(
+					orgAccount.EnvironmentRoles, environments, projects,
+				)
 			}
 		case errors.Is(orgErr, v2as.ErrAccountNotFound):
 			s.logger.Warn(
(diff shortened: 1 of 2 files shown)


Ubisoft-potato (dismissed):
Thanks! 👍

pkg/account/api/admin_account.go
```diff
@@ -328,6 +333,38 @@ func (s *AccountService) getConsoleAccountEnvironmentRoles(
 	return environmentRoles
 }
 
+func (s *AccountService) getSystemAdminMemberEnvironmentRoles(
+	roles []*accountproto.AccountV2_EnvironmentRole,
+	environments []*environmentproto.EnvironmentV2,
+	projects []*environmentproto.Project,
+) []*accountproto.ConsoleAccount_EnvironmentRole {
+	roleByEnv := make(map[string]accountproto.AccountV2_Role_Environment, len(roles))
+	for _, r := range roles {
```
cre8ivejp, line 344:
```suggestion
    // If the system-admin user belongs to multiple non-system organizations,
    // we must use the configured environment roles instead of configuring it to `VIEWER`.
	for _, r := range roles {
```

pkg/account/api/admin_account.go
```diff
@@ -328,6 +333,38 @@ func (s *AccountService) getConsoleAccountEnvironmentRoles(
...
+		if e.Archived {
+			continue
+		}
+		p, ok := projectSet[e.ProjectId]
+		if !ok || p.Disabled {
+			continue
+		}
+		role, ok := roleByEnv[e.Id]
+		if !ok {
+			role = accountproto.AccountV2_Role_Environment_VIEWER
```
cre8ivejp, line 360:
```suggestion
             // Only set to `VIEWER` if the system-admin user doesn't belong to the target organization
			role = accountproto.AccountV2_Role_Environment_VIEWER
```

pkg/account/api/admin_account.go
hvn2k1: https://github.com/bucketeer-io/bucketeer/pull/2601/commits/533bd11586f77797aa1efae91ac0dc19ec099c4d comment added

pkg/account/api/admin_account.go
hvn2k1: https://github.com/bucketeer-io/bucketeer/pull/2601/commits/533bd11586f77797aa1efae91ac0dc19ec099c4d comment added

cre8ivejp (approved):
Thank you!

Issue #2619, opened 2026-06-04
feat: add slack alerts when job/subscriber failed


hvn2k1:
When a batch job or subscriber job failed, alert slack channel

PR #2620, opened 2026-06-04
feat: alert slack when job/subscriber failed

hvn2k1:
Part of https://github.com/bucketeer-io/bucketeer/issues/2619

## Summary

Post a Slack alert when a batch job or subscriber consumer fails (follow-up to #2610, where such failures were silent). Posts via `chat.postMessage` with a bot token; throttled per job/consumer (30m) to avoid spam. Disabled unless a token + channel are set.

## Changes

- **Batch**: alert in `ExecuteBatchJob` on real failures (skips `ResourceExhausted`).
- **Subscriber**: alert on terminal failures in both consumers; suppressed on graceful shutdown.
- **notifier**: new `FailureAlerter` (Web API + in-memory throttle) + `JobFailure` proto type/render.
- **Config**: `--failure-alert-slack-token` / `--failure-alert-slack-channel` flags, wired through Helm, `values.dev.yaml`, and docker-compose.
- List of alerts:
  - processor.PushSenderName
  - processor.SegmentUserPersisterName
  - processor.DomainEventInformerName
  - batch.BatchJob_DatetimeWatcher
  - batch.BatchJob_EventCountWatcher
  - batch.BatchJob_ProgressiveRolloutWatcher
  - batch.BatchJob_ScheduledFlagChangeExecutor
  - batch.BatchJob_ExperimentCalculator


pkg/notification/sender/notifier/failure_alert.go
```diff
@@ -0,0 +1,171 @@
...
+		)
+		return
+	}
+	attachment := jobFailureAttachment(&senderproto.JobFailureNotification{
+		ServiceType:  serviceType,
+		JobName:      name,
+		ErrorMessage: failureErr.Error(),
+	})
+	if _, _, err := a.poster.PostMessageContext(
+		ctx,
```
Ubisoft-potato, line 128: ctx here is the caller's run context (batch job or subscriber). If the job ran close to its deadline this context may have little time left, causing the Slack HTTP call to time out before delivery. Consider `context.WithTimeout(context.Background(), 5*time.Second)` instead.

hvn2k1: https://github.com/bucketeer-io/bucketeer/pull/2620/commits/24377d927e72aaba750bc92ed3cb8a8b5d19bb60 I've splited the context

hvn2k1:
We'll use prometheus to alert

PR #2635, opened 2026-06-15
feat: implement postgres for insight svc

hvn2k1:
Part of #2306

## Summary
- Refactor `pkg/insights/storage/v2` to be DB-agnostic and add PostgreSQL support
- Split interface from impl: parent package keeps the `MonthlySummaryStorage` interface + `MonthlySummaryRecord` / `ListMonthlySummaryResult` types; MySQL impl moves to `mysql/`, new PostgreSQL impl under `postgres/` (SQL relocated to `mysql/sql` and `postgres/sql`)
- `pkg/web/cmd/server` and `pkg/batch/cmd/server` wire mysql or postgres monthly-summary storage based on `--storage-type`
- Add storage unit tests for both the MySQL and PostgreSQL implementations

## PostgreSQL notes
- Batch upsert uses numbered placeholders (`($1..$7), ($8..)`) and `ON CONFLICT (environment_id, yearmonth, source_id) DO UPDATE` instead of MySQL's `ON DUPLICATE KEY UPDATE`
- The "keep existing request_count when the incoming value is 0" rule is expressed as `CASE WHEN EXCLUDED.request_count = 0 THEN monthly_summary.request_count ELSE EXCLUDED.request_count END`
- `ListMonthlySummaries` builds sequential `$n` placeholders across the environment and source ID `IN (...)` lists

## Note
The insights API layer already received `MonthlySummaryStorage` via injection, so no API-layer changes were needed to remove the MySQL dependency.


Ubisoft-potato (approved):
Nice work!!🚀

PR #2199, opened 2025-11-11
feat: support for 3 or more variation when using progressive rollout

Ubisoft-potato:
## Summary

Fixed: #2193
This PR adds control and target variation selection for progressive rollouts, allowing users to explicitly choose which variations to use when rolling out changes.

### Backend
- Added `control_variation_id` and `target_variation_id` fields to `ProgressiveRolloutManualScheduleClause` and `ProgressiveRolloutTemplateScheduleClause`
- The old `variation_id` field is now deprecated but still supported for backward compatibility
- Added validation to ensure both control and target variations are provided and exist in the feature
- Updated rollout execution to use only the selected control and target variations

### Tests
- Updated e2e tests to use the new control and target variation fields
- Fixed test assertions to verify both variation IDs
- Updated expected rollout strategies to reflect correct weight distribution (target gets the specified weight, control gets the remainder)

## Motivation

Previously, progressive rollouts only allowed selecting a single variation, which was ambiguous about whether it represented the starting point or the target. This made it unclear which variation would receive the rollout traffic.

With explicit control and target variations:
- **Control variation**: The baseline variation users currently see
- **Target variation**: The new variation being rolled out
- The rollout gradually increases traffic to the target variation while decreasing the control

This gives users full control over progressive rollout behavior and makes the intent clear.

PR #2641, opened 2026-06-19
test: fix e2e test feature variation id

Ubisoft-potato:
## Problem

PR #2199 added support for 3+ variations in progressive rollout, which introduced required `control_variation_id` and `target_variation_id` fields. The old `variation_id` field alone is no longer accepted, so two E2E tests started failing:

- `TestEnableFeature`
- `TestListFeaturesFilterAutoOps`

```
rpc error: code = InvalidArgument desc = autoops:control variation id must be specified for a progressive rollout
```

## Fix

Updated the two failing tests to use the new `ControlVariationId` + `TargetVariationId` fields instead of the old deprecated `VariationId` field when creating a `ProgressiveRolloutManualScheduleClause`.

### DIFF
--- a/test/e2e/feature/feature_test.go
+++ b/test/e2e/feature/feature_test.go
@@ -249,8 +249,9 @@ func TestEnableFeature(t *testing.T) {
 		aoCLient,
 		req.Id,
 		&aoproto.ProgressiveRolloutManualScheduleClause{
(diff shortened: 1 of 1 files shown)


cre8ivejp (approved):
Thank you!

PR #2642, opened 2026-06-19
feat: implement postgres for scheduled feature change svc

hvn2k1:
Part of #2306

Adds the PostgreSQL implementation for the scheduled feature change service so it can run on either MySQL or Postgres.

## Changes
- Add Postgres storage layer for scheduled feature changes (`storage/v2/postgres`), with dedicated SQL files mirroring the MySQL queries.
- Move the MySQL implementation into `storage/v2/mysql` and split its inline queries into separate `.sql` files.
- Wire the storage selection into the `batch` and `web` servers based on the configured backend.
- Simplify the API handler and conflict-detection logic to rely on the shared storage interface.
- Add Postgres storage tests and a parity test to keep both backends in sync.


Ubisoft-potato (approved):
Thanks! 👍

PR #2643, opened 2026-06-19
fix(ui): correct split-equally weights and show variation allocation bar

steveninhle:
## Fix "Split equally" weights & show variation allocation bar

**Problems**
1. "Split equally between all variations" produced weights with floating-point drift so they didn't sum to 100 and triggered a false *"should be 100%"* error.
2. The variation allocation bar never showed, because its visibility relied on a stale memo (weight edits mutated the form array in place).

**Changes**
- Round split weights to 2 decimals and tolerate summation drift in all total checks (`handleCheckError` + both Yup validators + `isEquallyVariations`).
- Derive the bar's count locally and render from the live field value so it stays in sync.
- Build a new array on weight edits so the bar re-renders per keystroke.


cre8ivejp (approved):
Thank you!

PR #2650, opened 2026-06-22
feat: implement postgres for subscribers and refactor dwh layer

hvn2k1:
Part of #2306

Reorganize the subscriber storage layer by **purpose** and move all data-warehouse client/config initialization to the server, so the processor depends only on storage interfaces.

### Why
Evaluation/goal events go to the **data warehouse** (its own client/adapter, selectable as bigquery/mysql/postgres), while experiment/auto-ops counts read the **operational database** (the main `--storage-type` client). These were mixed together, and the operational counts were wrongly running through the DWH client.

### Changes
- **New storage layout** under `pkg/subscriber/storage/`:
  - `dwhstorage/` — evaluation + goal event inserts, with `bigquery/`, `mysql/`, `postgres/` implementations behind shared interfaces.
  - `operationalstorage/` — experiment + auto-ops-rule counts, with `mysql/` + `postgres/`.
  - (`v2/` keeps the legacy MAU storage untouched.)
- **DWH client + config moved to `cmd/server/server.go`**: the server reads the data-warehouse config, builds the DWH client (separate from the operational client), and injects ready storage interfaces into the persisters. The processor no longer references any DB client, dialect, or DWH config.
- **Fixed** the operational experiment/auto-ops counts to use the operational client instead of the DWH client.
- Added unit tests for every new mysql/postgres/bigquery implementation.


pkg/subscriber/cmd/server/server.go
```diff
@@ -690,6 +715,156 @@ func (s *server) createPostgresClient(
...
+			bqquerier.WithMetrics(registerer),
+		)
+		if err != nil {
+			return nil, nil, nil, nil, err
+		}
+		return evalWriter,
+			goalWriter,
+			ecbigquery.NewBigQueryEventStorage(eventQuerier, dataset, logger),
+			func() { eventQuerier.Close() },
+			nil
```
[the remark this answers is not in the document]
hvn2k1: https://github.com/bucketeer-io/bucketeer/pull/2650/commits/db4ff220807c7604dfccab51fe93b8d3cc867acd now bigquery client will close gracefully

pkg/subscriber/cmd/server/server.go
```diff
@@ -934,23 +1128,50 @@ func (s *server) registerPubSubProcessorMap(
...
 			persistentRedisClient,    // use persistent redis instance here for goal retry events
 			exClient,
 			ftClient,
 			processor.EvaluationCountEventDWHPersisterName,
-			registerer,
 			logger,
 		)
 		if err != nil {
-			return nil, err
+			return nil, nil, err
```
Ubisoft-potato, line 1171: if `NewEventsDWHPersister` fails here, `dwhCleanup` was already set at line 1156 but this returns `nil` for it — the DWH connections opened in `initDataWarehouseStorages` will leak since the caller never receives the cleanup func

pkg/subscriber/cmd/server/server.go
```diff
@@ -690,6 +716,171 @@ func (s *server) createPostgresClient(
...
+		project := dwhConfig.BigQuery.Project
+		dataset := dwhConfig.BigQuery.Dataset
+		batchSize := dwhConfig.BatchSize
+		evalWriter, err := dwhbigquery.NewEvaluationEventWriter(ctx, logger, project, dataset, batchSize, registerer)
+		if err != nil {
+			return nil, nil, nil, nil, err
+		}
+		goalWriter, err := dwhbigquery.NewGoalEventWriter(ctx, logger, project, dataset, batchSize, registerer)
+		if err != nil {
+			return nil, nil, nil, nil, err
```
Ubisoft-potato, line 783: if `NewGoalEventWriter` fails here, `evalWriter` is already open (holds a live BigQuery managed stream) but gets dropped with no cleanup — same for both writers if `bqquerier.NewClient` fails on line 780

pkg/subscriber/cmd/server/server.go
```diff
@@ -690,6 +716,171 @@ func (s *server) createPostgresClient(
...
+			client = dedicated
+			cleanup = func() { dedicated.Close() }
+		}
+		return dwhstorage.NewEvalEventWriter(dwhmysql.NewEvaluationEventStorage(client)),
+			dwhstorage.NewGoalEventWriter(dwhmysql.NewGoalEventStorage(client)),
+			ecmysql.NewMySQLEventStorage(client, logger),
+			cleanup,
+			nil
+	case "postgres":
+		client, err := createDWHPostgresClient(ctx, &dwhConfig.Postgres, logger)
```
Ubisoft-potato, line 761: `PostgresConfig` has a `UseMainConnection` field just like MySQL, but it's never checked here — this always creates a dedicated client. The MySQL branch above correctly branches on that flag, probably worth doing the same for postgres

hvn2k1: Hmm I did not intent to have use main connection for postgres, let me remove the UseMainConnection for postgres

pkg/subscriber/cmd/server/server.go
hvn2k1: https://github.com/bucketeer-io/bucketeer/pull/2650/commits/e433ab8bad3e42042a44d0638681c4ecf6fb5224 I add a check here

pkg/subscriber/cmd/server/server.go
hvn2k1: https://github.com/bucketeer-io/bucketeer/pull/2650/commits/e433ab8bad3e42042a44d0638681c4ecf6fb5224 fail check updated

Ubisoft-potato (approved):
Thanks! LGTM!

PR #2653, opened 2026-06-22
chore: remove user event persisters (upsert maus)

hvn2k1:
The mau is deprecated and replaced with insight dashboard so these codes are deprecated

Removes:
- Storage mau implementation (table already removed in migration)
- User event persister, it was used to upsert mau data
- Helm and docker compose config


Ubisoft-potato (approved):
Thanks for cleanup!

Issue #2652, opened 2026-06-22
test: e2e test for postgres operational database and data warehouse


hvn2k1:
### Describe the feature you'd like

To ensure the credibility of Postgres implementation for operational database (and data warehouse), an e2e flow running on Postgres deployment is needed

### Background

Replicate the same flow of e2e test for mysql x big query

PR #2659, opened 2026-06-24
chore: implement delete e2e data for postgres

hvn2k1:
Part of https://github.com/bucketeer-io/bucketeer/issues/2652

## What

Adds a PostgreSQL counterpart to the existing `delete-e2e-data-mysql` tooling so E2E test data can be cleaned up when the operational database is Postgres.

## Changes

- **New `hack/delete-e2e-data-postgres` CLI tool** — mirrors `hack/delete-e2e-data-mysql`, but uses the `pkg/storage/v2/postgres` client and `$n` numbered placeholders instead of MySQL's `?`. It deletes rows scoped to the `e2e` environment (and `e2e` organization for `account_v2`), optionally filtered by `--test-id`, across the same set of tables. The `name != 'localenv'` guard is preserved so the e2e service account is not removed.
  - Includes `command.go`, `main.go`, `Makefile`, `Dockerfile`, `README.md`, `.gitignore`, and `go.mod`/`go.sum`.
- **Root `Makefile`** — new targets:
  - `delete-e2e-data-postgres` — builds and runs the tool with `--postgres-*` flags.
  - `delete-dev-container-postgres-data` — Minikube wrapper (`$(minikube ip):32100`).
  - `docker-compose-delete-data-postgres` — Docker Compose wrapper (`localhost:5432`).
  - `docker-compose-delete-redis-retry-keys` — Docker Compose variant of the redis retry-key cleanup (`localhost:6379`), since the original `delete-redis-retry-keys` targets the Minikube redis nodeport.
- **`DEVELOPMENT.md`** — documents the "Delete E2E data" flow for both MySQL and Postgres, in both the Minikube and Docker Compose sections, keyed off the operational database type.

## How to use

```shell
# Minikube
make delete-dev-container-postgres-data

# Docker Compose
make docker-compose-delete-data-postgres
```


Ubisoft-potato:
The MySQL cleanup flow documents two steps — `delete-dev-container-mysql-data` **and** `delete-mysql-data-warehouse-data` — because event/goal data lives in a separate data-warehouse store. The new Postgres section only documents `delete-dev-container-postgres-data`, and there's no `delete-postgres-data-warehouse-data` Makefile target in this PR.

If the project is configured to use Postgres as the data warehouse (`dataWarehouse.type` in `values.dev.yaml`), the event-counter and analytics tables never get cleaned after E2E runs, which causes stale data to accumulate and corrupts subsequent experiment-result counts.

Could you add an equivalent `hack/delete-postgres-data-warehouse` tool (mirroring `hack/delete-mysql-data-warehouse`) and a corresponding Makefile target, or clarify if Postgres-as-data-warehouse is intentionally out of scope for this PR?

Ubisoft-potato (dismissed):
Thanks! 👍

hvn2k1:
@Ubisoft-potato the other pr for delete dwh data is merged, thank you!

Ubisoft-potato (approved):
Thanks! 👍

PR #2662, opened 2026-06-25
test: use decisive 50-user value fixture and assert metric-aware winner

cre8ivejp:
## Summary

Makes the experiment-result e2e fixture statistically meaningful (Follow-up F) and adds the first end-to-end assertion of the metric-aware best-variations ("winner") logic (Follow-up E). Test-only change — no production code touched.

## Problem

`TestExperimentResult` / `TestGrpcExperimentResult` used only n=2–3 users per arm. At that size the value-metric posterior is **prior-dominated**, so the goldens (`GoalValueSumPerUserProbBest` / `ProbBeatBaseline`) reflected prior arithmetic rather than real signal — a regression in the NIG math could pass these tests unnoticed.

Separately, the metric-aware best-variations lists from Follow-up E (`Summary.BestVariations` for CVR, `Summary.BestVariationsValue` for value) had **no end-to-end coverage** — only the in-process `TestPickBestVariations` unit test.

## Approach

- **Fixture: 50 users/arm (100 total), value-decisive, CVR-tied.**
  - 1 evaluation + 1 goal per user (100% conversion both arms → CVR is a
    deliberate tie, so no CVR winner).
  - Per-user value: baseline ~10 vs treatment ~15, with a deterministic
    `+0/+0.1` alternating spread so `VAR_SAMP > 0` (the calculator skips value
    metrics when per-variation variance is exactly zero — see
    `experiment_calculator.go`). The spread is tiny vs the 10-vs-15 separation
    and stays below the winsorization cap, so it tests the model, not the cap.
  - At n=50 the empirical-Bayes prior weight is ~2%, so the value goldens now
    reflect posterior signal.
- **Decisive value goldens:** treatment `GoalValueSumPerUserProbBest` /
  `ProbBeatBaseline` = 1.0, baseline = 0.0; `ExperimentCount.ValueSum`
  A=502.5, B=752.5 (deterministic from the fixture). CVR goldens updated to
  the ~0.50 tie.
- **Metric-aware winner assertion (`checkExperimentSummary`):**
  `BestVariationsValue` = exactly `[treatment]` with `IsBest=true` and
  probability > 0.95; `BestVariations` (CVR) empty. Exercises the
  `best_variations_value` proto field through calculator → storage → API.
- **Robustness fixes:** `waitAndCheckExperimentResult` looks up variation
  results by `VariationId` (not slice index) and waits for both the linked
  counts and value-metric convergence (`B.ProbBest ≥ 0.5`) before asserting,
  avoiding races on partial calculator output.
- **Helpers:** batched event registration (gRPC + HTTP, 50/chunk) and bulk
  split individual-targeting, so the 200-event fixture is ~4 round-trips rather
  than 200.

## Files changed

- `test/e2e/eventcounter/eventcounter_test.go` — fixture, helpers, goldens,
  and the `checkExperimentSummary` assertion.

No production code, protos, or UI changed.

## Test plan

- [x] `TestExperimentResult` — PASS in the dev container (~45s).
- [x] `TestGrpcExperimentResult` — PASS in the dev container (~55s).
- Both assert the decisive value goldens, the metric-aware best-variations
  (`BestVariationsValue=[B]`, CVR list empty), and SRM `SKIPPED` (FIXED default
  strategy).
- `go vet` + `gofmt` clean.

## Notes

- The fixture is a **clean-separation** winner (tiny variance) — it proves the pipeline produces the correct decisive verdict and the metric-aware split,
  not discrimination under realistic noise (that belongs in the seeded  `normal_inverse_gamma_test.go` unit tests, which avoid e2e flakiness).
- CVR is intentionally a tie to isolate the value path; CVR-winner detection   isn't exercised here (possible small follow-up: make one arm convert less).

### DIFF
--- a/test/e2e/eventcounter/eventcounter_test.go
+++ b/test/e2e/eventcounter/eventcounter_test.go
@@ -2285,6 +2285,10 @@ func addFeatureIndividualTargetingBulkSplit(
 }
 
 func experimentResultCountsReady(vsA, vsB *ecproto.VariationResult) bool {
+	if vsA.EvaluationCount == nil || vsA.ExperimentCount == nil ||
+		vsB.EvaluationCount == nil || vsB.ExperimentCount == nil {
+		return false
+	}
 	n := int64(experimentResultUsersPerVariation)
 	return vsA.EvaluationCount.EventCount == n &&
 		vsA.EvaluationCount.UserCount == n &&
@@ -2294,6 +2298,37 @@ func experimentResultCountsReady(vsA, vsB *ecproto.VariationResult) bool {
 		vsA.ExperimentCount.UserCount == n &&
 		vsB.ExperimentCount.EventCount == n &&
 		vsB.ExperimentCount.UserCount == n
+}
+
+// experimentResultProbFieldsReady reports whether the calculator has populated
+// the CVR and value-metric distribution summaries for a variation.
+func experimentResultProbFieldsReady(vr *ecproto.VariationResult) bool {
+	return vr.CvrProbBest != nil &&
+		vr.CvrProbBeatBaseline != nil &&
+		vr.GoalValueSumPerUserProbBest != nil &&
+		vr.GoalValueSumPerUserProbBeatBaseline != nil
+}
+
+func requireDistributionSummary(
+	t *testing.T,
+	variationValue, fieldName string,
+	ds *ecproto.DistributionSummary,
+) *ecproto.DistributionSummary {
+	t.Helper()
+	if ds == nil {
+		t.Fatalf("variation: %s: %s should be populated, got nil", variationValue, fieldName)
+	}
+	return ds
+}
+
+func requireVariationCounts(t *testing.T, variationValue string, vr *ecproto.VariationResult) {
+	t.Helper()
+	if vr.EvaluationCount == nil {
+		t.Fatalf("variation: %s: evaluation count should be populated, got nil", variationValue)
+	}
+	if vr.ExperimentCount == nil {
+		t.Fatalf("variation: %s: experiment count should be populated, got nil", variationValue)
+	}
 }
 
 func getVariationResult(vrs []*ecproto.VariationResult, id string) *ecproto.VariationResult {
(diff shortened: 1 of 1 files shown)


test/e2e/eventcounter/eventcounter_test.go
```diff
@@ -2322,6 +2211,173 @@ func addFeatureIndividualTargetingBulk(
...
+		if vsB.GoalValueSumPerUserProbBest == nil ||
+			vsB.GoalValueSumPerUserProbBest.Mean < 0.5 {
+			probBest := 0.0
+			if vsB.GoalValueSumPerUserProbBest != nil {
+				probBest = vsB.GoalValueSumPerUserProbBest.Mean
+			}
+			t.Logf("Retry %d/%d: value-metric probs not ready yet (B probBest=%f)",
+				i+1, retryTimes, probBest)
+			continue
+		}
```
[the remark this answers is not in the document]
cre8ivejp: Good catch on conflating readiness vs outcome — I'll consider a follow-up that nil-checks value-metric fields instead of thresholding on Mean.

Not applying this suggestion as-is: value NIG summaries don't set Rhat (only Mean), and Summary is always non-nil. The ≥ 0.5 gate is a practical readiness heuristic for this fixture; final outcomes are asserted in checkExperimentVariationResultB and checkExperimentSummary.

test/e2e/eventcounter/eventcounter_test.go
```diff
@@ -2823,37 +2879,36 @@ func createUserID(t *testing.T, uuid string) string {
...
+	if vsA.ExperimentCount.UserCount != n {
 		t.Fatalf("variation: %s: experiment user count is not correct: %d", variationValue, vsA.ExperimentCount.UserCount)
 	}
-	if diff := cmp.Diff(vsA.ExperimentCount.ValueSum, 1.0, compareFloatOpt); diff != "" {
+	if diff := cmp.Diff(vsA.ExperimentCount.ValueSum, valueSum, compareFloatOpt); diff != "" {
 		t.Fatalf("variation: %s: experiment value sum is not correct: %f", variationValue, vsA.ExperimentCount.ValueSum)
 	}
-	// cvr prob best
-	if diff := cmp.Diff(vsA.CvrProbBest.Mean, 0.57, compareFloatBayesian); diff != "" {
+	if diff := cmp.Diff(vsA.CvrProbBest.Mean, 0.50, compareFloatBayesian); diff != "" {
```
Ubisoft-potato: `vsA.CvrProbBest`, `vsA.GoalValueSumPerUserProbBest`, etc. are proto pointer fields that can be `nil` — dereferencing them directly here will panic instead of failing cleanly. 

The guard at line 2363 only nil-checks the `vsB` side. Same issue in `checkExperimentVariationResultB` at lines 2950/2968. Worth adding a nil check for `vsA` before entering these functions.

cre8ivejp: I have added nil-guard.
https://github.com/bucketeer-io/bucketeer/pull/2662/commits/a833c77769856eec58bf816ed37db829f2896f1d

Ubisoft-potato (approved):
Thanks!👍

PR #2663, opened 2026-06-25
fix(rest): skip logging for HTTP redirect responses

cre8ivejp:
## Summary

- Stop `LogServerMiddleware` from logging 3xx redirect responses
- Redirects are expected HTTP behavior (e.g. `/assets` → `/assets/`, scanner probes like `GET /temp/`) and were incorrectly logged at ERROR severity
- Add unit tests covering 200, 301, and 404 logging behavior

## Context

The web service dashboard REST server was emitting many ERROR logs for requests like `GET /temp/` with status 301. These are not application failures — they come from Go's `FileServer` redirect behavior combined with the logging middleware treating any non-2xx/4xx response as an error.

## Test plan

- [x] `go test ./pkg/rest/ -run TestLogServerMiddleware`

### DIFF
--- a/pkg/rest/log.go
+++ b/pkg/rest/log.go
@@ -40,7 +40,7 @@ func LogServerMiddleware(logger *zap.Logger) middleware {
 						body:           new(bytes.Buffer),
 					}
 					next.ServeHTTP(rr, r)
-					// Skip logging for successful and redirect responses.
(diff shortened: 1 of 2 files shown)


Ubisoft-potato (approved):
Thanks!

Issue #2215, opened 2025-11-17
feat: notification center in dashboard


nnnkkk7:
### Description
Add a notification center feature to the Bucketeer admin dashboard to inform users about platform updates, announcements, and important changes.

### Proposed Solution
Implement a notification icon (bell icon) in the dashboard header that:
- Displays a badge when there are unread notifications
- Opens a dropdown/panel when clicked to show recent updates and announcements
- Allows users to view platform updates, new features, and important notices

### Use Case
Currently, users may miss important platform updates or new features. A notification center would:
- Keep users informed about Bucketeer updates and new releases
- Improve user engagement with new features
- Provide a centralized place for important announcements
- Enhance the overall user experience

hvn2k1:
## Notification Center — Feature Description

A notification center that lets **system admins broadcast platform-wide messages** to users, viewable from the dashboard. It gives users a centralized inbox to read announcements and updates from the platform.

### What It's Used For

Keep users informed about platform-related communications, such as:
- Announcements
- Maintenance windows
- Bug reports / known issues
- Platform updates

### Content Format

- Messages support **rich content**: a mix of text and links, and images via external links for now.
- The message editor will use **Markdown** (HTML support to be evaluated).

### Permissions

- **Only system admins** can create, edit, and publish notifications.
- All users can view notifications addressed to them.

### Console Behavior

**For system admins (authoring):**
- A **message editor** to draft notifications.
- **Save as draft** option, with drafts persisted in the backend.
- Ability to publish drafts when ready.

**For all users (viewing):**
- An **inbox list** showing received notifications.
- A **dedicated panel or page** to view the full notification message.
- A view that separates **read and unread** messages into different **tabs**.

### Components to Build

1. **Editor** — draft, save draft (backend-persisted), and publish messages. _(Sys admin only)_
2. **Inbox list** — browse received notifications.
3. **Detail view** — panel or new page showing the full message content.
4. **Read/Unread view** — tabbed separation of read vs. unread messages.

t-kikuc:
@hvn2k1 
Can we close this issue as finished?

PR #2680, opened 2026-07-03
docs: add RFC for system notification center

hvn2k1:
Part of #2215

## What this PR does

Adds an RFC for the system notification center: system admins publish platform-wide announcements (Markdown), and every console user gets an inbox with unread/read tabs, a bell badge, search, date filtering, sorting, and pagination.

## Points

- The existing `notification` domain (Slack/FCM event subscriptions) will be renamed to `subscription` first, since its RPCs, REST paths, and tables already use that word. The new feature then takes the `notification` name. The rename changes nothing external (REST paths and tables are already `subscription`-named).
- Read state is a per-user marker row: unread = published notification without a marker for the viewer's email. No fan-out at publish time, so publishing is O(1) regardless of user count.
- Published notifications cannot be edited in v1 (delete and republish instead); this is listed as an open question for later.


docs/rfcs/0047-system-notification-center.md
```diff
@@ -0,0 +1,193 @@
...
+```
+
+- Registered in `pkg/web/cmd/server/server.go` as a new gRPC service plus a grpc-gateway handler (`RegisterNotificationServiceHandlerFromEndpoint`), exposed under `/v1/notification*` on the web gateway.
+- Read state is a per-user marker row; **unread = published notification without a marker row for the viewer's email**. No fan-out rows are written at publish time, so publishing is O(1) regardless of user count.
+
+## 4. Design Details
+
+### 4.1 Database Design
+
+Two tables, added to both `migration/mysql/` and `migration/postgres/` (Atlas, `YYYYMMDDHHMMSS_create_notification_tables.sql`).
```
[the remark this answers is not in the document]
hvn2k1: https://github.com/bucketeer-io/bucketeer/pull/2680/commits/7917fe1eb71c56d4781d74c14b4cdaf110bcfcda updated

docs/rfcs/0047-system-notification-center.md
```diff
@@ -0,0 +1,193 @@
...
+    CONSTRAINT `fk_notification_read`
+        FOREIGN KEY (`notification_id`)
+        REFERENCES `notification` (`id`) ON DELETE CASCADE
+);
+```
+
+Notes:
+
+- **`tags` as JSON** keeps the schema simple; the tag palette (name + color pairs shown in the mockup) is a fixed set defined in the console for v1. If tags later need server-side management, a `notification_tag` master table can be added without touching this schema.
+- **`email` as viewer identity**: accounts are keyed by email across organizations (`account_v2.email`), and the JWT carries `Email`, so read state naturally spans organizations for the same person.
```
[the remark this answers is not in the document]
hvn2k1: https://github.com/bucketeer-io/bucketeer/pull/2680/commits/7917fe1eb71c56d4781d74c14b4cdaf110bcfcda address the account read correctly

docs/rfcs/0047-system-notification-center.md
```diff
@@ -0,0 +1,193 @@
...
+
+- **`tags` as JSON** keeps the schema simple; the tag palette (name + color pairs shown in the mockup) is a fixed set defined in the console for v1. If tags later need server-side management, a `notification_tag` master table can be added without touching this schema.
+- **`email` as viewer identity**: `account_v2` is keyed by `(email, organization_id)`, so one person has one account row per organization — but all of them share the same email, and the access token carries a single `Email` for the person. Keying read state by email therefore gives one shared read state across organizations for the same person.
+- **Growth**: `notification` grows by human-authored rows (tens per month). `notification_read` grows by at most `users × notifications` and only when a user actually reads — negligible at Bucketeer's console-user scale.
+- Deleting a notification cascades its read markers. (If we prefer avoiding FKs like the `team` table does, the storage layer deletes both rows in one transaction instead.)
+
+### 4.2 API List
+
+**Viewer APIs** - any authenticated console user (identity = `token.Email` from the access token):
+
```
t-kikuc, line 109: `ListNotifications` is listed under Viewer APIs, but there is no separate admin endpoint for listing drafts. `GetNotification` explicitly states "Drafts are visible to system admins only," but `ListNotifications` has no equivalent clause.

Please clarify: (a) does this endpoint accept a `status` filter (DRAFT / PUBLISHED)? (b) are drafts automatically excluded for non-admin callers?

Without this, an implementer won't know how a system admin is supposed to find and manage their drafts.

docs/rfcs/0047-system-notification-center.md
```diff
@@ -0,0 +1,193 @@
...
+    participant S as NotificationService
+    participant DB as MySQL/Postgres
+
+    rect rgb(230, 240, 255)
+    Note over C,DB: 1. Bell badge + dropdown
+    C->>GW: GET /v1/notifications/unread_count
+    GW->>S: GetNotificationUnreadCount
+    S->>DB: COUNT published LEFT JOIN read WHERE read.email IS NULL
+    S-->>C: {count: 3}
+    C->>GW: GET /v1/notifications?read_status=UNREAD&page_size=5
```
t-kikuc, line 175: The second request flow (`GET /v1/notifications?read_status=UNREAD`) and the third (`GET /v1/notification?id=X`) omit the `GW->>S` arrow, jumping from `C->>GW` directly to `S->>DB`. The first and fourth flows include it. Adding the missing arrows would keep the diagram consistent.

Ubisoft-potato:
One thing I'm wondering about the unread model: since unread is just "published row without a marker", a brand-new user will see every announcement ever published as unread. After a year or so of tens per month, someone joining gets a bell badge like 200+ on their first login, and the only way to clear it is mark all as read.

The list view is somewhat protected by the `Last 30 days` default filter, but `GetNotificationUnreadCount` has no such bound.

A few options:
- only count notifications published after the user's earliest `account_v2.created_at`
- cap the badge display on the client (e.g. `99+`) and just accept it
- leave it as is

Any of these is fine with me, but I think it's worth deciding now and adding a line to the Open Questions section, since the count query changes depending on the answer.

Ubisoft-potato:
Another gap I noticed while going through the API list against the mockups: how does the "Publish notification" tab list existing drafts?

The design shows a draft opened in the detail panel with a `Draft` badge and an "Edit Draft" button, so admins clearly need to browse drafts somewhere. But `ListNotifications` only mentions keyword search, read/unread filter, date range, sort and pagination — there's no status filter. `GetNotification` says drafts are admin-only, but nothing says whether `ListNotifications` excludes drafts for regular users.

I think we need:
- an admin-only `status` filter on `ListNotifications` (or a separate admin list RPC)
- an explicit note that non-admins only ever get `PUBLISHED` rows from the list

One more detail hiding under this: drafts have `published_at = 0`, so the "Newest" sort and the `(status, published_at)` index don't work for them — the draft listing would need to order by `created_at` / `updated_at` instead. Worth a sentence when the `order_by` field gets defined in the proto.

Since phases 2 and 3 both build on the proto contract, I'd rather settle this in the RFC than discover it mid-implementation.

hvn2k1:
@Ubisoft-potato thank you for your review. I made some changes to resolve your concern:
Let's split into 2 API for list notifications:
* ListNotifications so any account can use that, we call 2 times in the notification page: first filter by unread, return list of unread notis and total count, then call read notis it return list of unread and total count, the FE switch tab call that twice and load every thing, each tab 
* ListDraftNotification only allow for system admin, return list draft notifications and total count.

This would make the implementation easier for both backend and frontend

Ubisoft-potato:
> @Ubisoft-potato thank you for your review. I made some changes to resolve your concern: Let's split into 2 API for list notifications:
> 
> * ListNotifications so any account can use that, we call 2 times in the notification page: first filter by unread, return list of unread notis and total count, then call read notis it return list of unread and total count, the FE switch tab call that twice and load every thing, each tab
> * ListDraftNotification only allow for system admin, return list draft notifications and total count.
> 
> This would make the implementation easier for both backend and frontend

Hi! @hvn2k1 

Looks great for now! 
Thanks for your update system design!

t-kikuc (commented):
Thank you

t-kikuc (commented):
Thank you

docs/rfcs/0047-system-notification-center.md
```diff
@@ -0,0 +1,195 @@
...
+- Per-user read/unread state, unread count for the bell badge and tab counters.
+- Announcements are **global** (all organizations, all environments).
+- MySQL and PostgreSQL support (same as other `storage/v2` domains).
+- Rename the legacy domain to `subscription` first (see Background).
+
+**Non-Goals.**
+
+- Per-organization / per-environment targeting, maybe in the future
+- Scheduled publishing (publish-at-future-time).
+
```
t-kikuc, line 37: I want an easy mock UI to easily understand what we're going to develop.

maybe you need [1] badge sidebar for viewers [2] notification center page [3] admin page
(+ popup dialog)

docs/rfcs/0047-system-notification-center.md
```diff
@@ -0,0 +1,195 @@
...
+**Issue**: [https://github.com/bucketeer-io/bucketeer/issues/2215](https://github.com/bucketeer-io/bucketeer/issues/2215)
+
+## 1. Background
+
+Users currently have no in-product way to learn about platform updates, maintenance windows, known issues, or new features. The existing `notification` domain (`pkg/notification`, `proto/notification`) is **not** this feature: it manages Slack/FCM *subscriptions* that forward domain events (feature changed, experiment finished, …) to external recipients.
+
+This feature is a **broadcast inbox**:
+
+- **System admins** draft, edit, publish, and delete announcements from a "Publish notification" tab.
+- **All authenticated console users** see published announcements in Unread / Read tabs, a bell dropdown with the latest unread items, and a detail panel rendering Markdown.
```
t-kikuc, line 14: Basically, this badge UI seems really good.

This assumes users open the console frequently enough to notice announcements. 
For time-sensitive messages  (maintenance windows, breaking changes), users who don't visit for days will miss them. 
Consider acknowledging  this limitation and scoping an external delivery channel (email / Slack / RSS, etc.) as a future phase.

docs/rfcs/0047-system-notification-center.md
hvn2k1: This is just future suggestion, open question, so there should be no UI for this for now

docs/rfcs/0047-system-notification-center.md
hvn2k1: No we have List draft notifications API only for sysadmin

docs/rfcs/0047-system-notification-center.md
hvn2k1: https://github.com/bucketeer-io/bucketeer/pull/2680/changes/11a296940cb0b47077f72d28f082fc927e44761b updated

docs/rfcs/0047-system-notification-center.md
hvn2k1: I added it to non goal

docs/rfcs/0047-system-notification-center.md
```diff
@@ -0,0 +1,198 @@
+# System Notification Center
```
t-kikuc, line 1: I noticed we need i18n.

Whenever we implement it, we should consider the data structure and UI for i18n in advance.

One of my ideas is that an admin optionally writes a message for each language, and shows the English version if the message for the user's language is not written.

hvn2k1: I see, I add the i18n support https://github.com/bucketeer-io/bucketeer/pull/2680/commits/2e9ae4c6606fa7bdfb5cb25871f932798ce43813

t-kikuc (approved):
Thank you!

Issue #2152, opened 2025-10-03
feat: Massive real-time flag synchronization for clients


t-kikuc:
### Describe the feature you'd like

Enable synchronization of flags in real-time for a massive number of clients, more than 3M concurrent clients.

We need to consider some ways, such as WebSocket.
This feature might cost so much.

### Background

Currently, the real-time flag synchronization for clients is supported via Firebase Cloud Messaging (FCM).
https://docs.bucketeer.io/integration/pushes

However, the synchronization pace is not so fast for massive clients due to the FCM quota.

t-kikuc:
Tasks: 

- [x] **Proto**: Define SSE payload types #2592 
- [x] **Backend**:
  - [x] Domain topic subscription, connection pooling #2630 
  - [x] Evaluation, Metrics #2665 
  - [x] Graceful shutdown #2675 
  - [x] tag propagation #2679 
  - [x] max conns limitation #2676 
- [x] **Infra**:
  - [x] Gateway config for allowing SSE
  - [x] Resize memory of API pods
  - [x] HPA on connection count
- [x] **Load Test**
- [ ] **Client SDKs**: 
  - [ ] JavaScript
  - [ ] iOS
  - [ ] Android
  - [ ] later: 
      - [ ] Flutter, 
      - [ ] React, 
      - [ ] React Native

t-kikuc:
note:  we need to add safety against the cache gap (like batch per 1min)

 #2630

t-kikuc:
### TODO: SDK Documentation

The JavaScript client SDK PR (https://github.com/bucketeer-io/javascript-client-sdk/pull/333) adds three new config properties for SSE streaming. Documentation is needed in [bucketeer-docs](https://github.com/bucketeer-io/bucketeer-docs):

| Property | Type | Default | Description |
|---|---|---|---|
| `enableStreaming` | `boolean` | `false` | `false` = polling, `true` = SSE streaming |
| `streamingFallbackToPolling` | `boolean` | `true` | Fall back to polling when SSE fails |
| `eventSource` | `EventSourceLike` | built-in | Optional custom SSE transport (POST-capable). Omit to use built-in `FetchEventSource` |

Platform notes to document:
- React Native: may need a `ReadableStream` polyfill or a custom `eventSource` injection
- All platforms: existing users keep polling with no config change; add `enableStreaming: true` to switch

t-kikuc:
**TODO: Remove SSE circuit breaker configuration**

[...]

PR #2684, opened 2026-07-06
fix(stream): prevent buffer overflow in SSE streaming middleware

t-kikuc:
Part of #2152, Follows #2679

## What this PR does

Prevents unbounded memory growth in the streaming middleware's `responseRecorder` during long-lived SSE connections.

## Background

The `responseRecorder` middleware copies every `Write()` into an internal `bytes.Buffer` for logging. For short-lived HTTP requests this is fine, but SSE connections are long-lived — heartbeats and patch events accumulate in the buffer indefinitely, eventually exhausting memory. The SSE handler needs a way to opt out of body recording after headers are written.

## Points

- `DisableBodyRecording` walks the full recorder chain (middleware can nest recorders) to ensure no layer accumulates bytes
- The SSE handler calls `DisableBodyRecording` before `WriteHeader` so no event data is ever buffered


cre8ivejp (approved):
Thanks!

PR #2688, opened 2026-07-09
feat(stream): add SSE-dedicated service and timeout config for streaming

t-kikuc:
Part of #2152, Follows #2684

## What this PR does

Adds SSE-dedicated Kubernetes Service, BackendConfig, and MultiClusterService so that the streaming endpoint (`/v1/gateway/stream_evaluations`) can have its own timeout configuration separate from the main API service.

## Background

The main API service has a 60s timeout, which kills long-lived SSE connections. SSE streams need heartbeat-aware idle timeouts (envoy) and a longer hard deadline (GCP BackendConfig). Splitting into a separate Service lets each path have its own timeout without affecting existing API traffic.

This is one of several infra PRs for the SSE streaming feature — further config and deployment changes will follow.

## Points

- Envoy route for SSE uses `timeout: 0s` (no hard deadline) + `idle_timeout: 300s` (reset by heartbeat)
- GCP BackendConfig `timeoutSec: 86400` is a hard deadline not reset by heartbeat — configurable via `values.yaml`
- nginx config adds SSE-specific location block with `proxy_buffering off` and 120s read/send timeout
- Ingress `rulePaths` places the SSE path before the catch-all so it matches first

### DIFF
--- a/manifests/bucketeer/charts/api/templates/multi-cluster-service-sse.yaml
+++ b/manifests/bucketeer/charts/api/templates/multi-cluster-service-sse.yaml
@@ -1,6 +1,6 @@
 {{- if and .Values.gcpMultiCluster.enabled .Values.gcpMultiCluster.configCluster }}
-# Mirrors multi-cluster-service.yaml with changes: 
-# backendConfig->api-sse, no metrics/admin ports.
+# Mirrors multi-cluster-service.yaml with changes:
+# use '-sse' for backendConfig, no metrics/admin ports.
 apiVersion: networking.gke.io/v1
 kind: MultiClusterService
 metadata:
--- a/manifests/bucketeer/charts/api/templates/multi-cluster-service-sse.yaml
+++ b/manifests/bucketeer/charts/api/templates/multi-cluster-service-sse.yaml
@@ -1,0 +1,29 @@
+{{- if and .Values.gcpMultiCluster.enabled .Values.gcpMultiCluster.configCluster }}
+# Mirrors multi-cluster-service.yaml with changes: 
+# backendConfig->api-sse, no metrics/admin ports.
+apiVersion: networking.gke.io/v1
+kind: MultiClusterService
+metadata:
+  name: {{ .Values.gcpMultiCluster.sseService.name }}
+  namespace: {{ .Values.namespace }}
+  labels:
+    app: {{ template "api.name" . }}
+    chart: {{ template "api.chart" . }}
+    release: {{ template "api.fullname" . }}
+    heritage: {{ .Release.Service }}
+  annotations:
+    cloud.google.com/backend-config: '{"ports": {"{{ .Values.service.externalPort }}":"{{ template "api.fullname" . }}-sse"}}'
+    networking.gke.io/app-protocols: '{"service":"HTTP2"}'
+    cloud.google.com/neg: '{"ingress": true}'
+spec:
+  template:
+    spec:
+      selector:
+        app: {{ template "api.name" . }}
+        release: {{ template "api.fullname" . }}
+      ports:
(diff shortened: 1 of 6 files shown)


manifests/bucketeer/charts/api/templates/multi-cluster-service-sse.yaml
```diff
@@ -0,0 +1,29 @@
+{{- if and .Values.gcpMultiCluster.enabled .Values.gcpMultiCluster.configCluster }}
+# Mirrors multi-cluster-service.yaml with changes: 
+# backendConfig->api-sse, no metrics/admin ports.
```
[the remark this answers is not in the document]
t-kikuc: fixed: [76df3f7](https://github.com/bucketeer-io/bucketeer/pull/2688/commits/76df3f7c32042e3239b18a1b990c804855da98b9)

manifests/bucketeer/charts/api/templates/service-sse.yaml
```diff
@@ -0,0 +1,26 @@
+# Mirrors service.yaml with changes:
+# backendConfig->api-sse, no metrics/admin ports, no envoy/metrics labels.
```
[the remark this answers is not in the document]
t-kikuc: fixed: [76df3f7](https://github.com/bucketeer-io/bucketeer/pull/2688/commits/76df3f7c32042e3239b18a1b990c804855da98b9)

manifests/bucketeer/charts/api/values.yaml
```diff
@@ -125,11 +123,21 @@ envoy:
...
   externalPort: 9000
+sse:
+  backendConfig:
+    timeoutSec: 86400
 ingress:
   name: api
   host:
   staticIPName:
   rulePaths:
+    - path: /v1/gateway/stream_evaluations
```
hvn2k1, line 134: In https://github.com/bucketeer-io/bucketeer/pull/2688/changes?diff=unified#diff-f8c6b0a2324c7abec1466f8b8fd60b2eb17a0b0738d2361925237bd84df8eb96R1 we have gcp enable guard, so should this rule path have similar config?

manifests/bucketeer/charts/api/templates/multi-cluster-service-sse.yaml
```diff
@@ -0,0 +1,29 @@
+{{- if and .Values.gcpMultiCluster.enabled .Values.gcpMultiCluster.configCluster }}
```
hvn2k1, line 1: should we also check if gcp enabled too?

manifests/bucketeer/charts/api/values.yaml
t-kikuc: GCP vs non-GCP switching is handled in `ingress.yaml` itself (annotations +  ingressClassName).  The BackendConfig needs the guard because it's a GCP-specific CRD, but the Ingress path routing is standard Kubernetes and works on both.

And the existing non-SSE rulePaths don't have a gcpEnabled guard either — `rulePaths` is just data consumed by `ingress.yaml` via `toYaml`.

manifests/bucketeer/charts/api/templates/multi-cluster-service-sse.yaml
t-kikuc: This mirrors the existing `multi-cluster-service.yaml`, which uses the same guard (`gcpMultiCluster.enabled && configCluster`) without a `gcpEnabled` check. Since MultiClusterService is a GKE CRD, `gcpMultiCluster.enabled:
  true` inherently implies a GCP environment.

hvn2k1 (approved):
Thank you!

PR #2689, opened 2026-07-09
refactor(subscription): rename remaining notification aliases in manifest

hvn2k1:
Part of #2215, follow-up to #2685

## What this PR does

Completes the `notification` → `subscription` rename in the deployment-facing names that #2685 intentionally left untouched — CLI flags and the configs that feed them, kept in lockstep in one PR:

- **CLI flags**: `notification-service` → `subscription-service` (api/batch/subscriber, default `notification:9090` → `subscription:9090`), `notification-service-port` → `subscription-service-port` (web). Kingpin derives env var names from flag names, so the env vars rename with them.
- **Env vars**: `BUCKETEER_WEB_NOTIFICATION_SERVICE_PORT` → `BUCKETEER_WEB_SUBSCRIPTION_SERVICE_PORT`, `BUCKETEER_{API,BATCH,SUBSCRIBER}_NOTIFICATION_SERVICE` → `BUCKETEER_*_SUBSCRIPTION_SERVICE` (helm deployments and docker-compose)
- **Helm values**: `notificationServicePort` → `subscriptionServicePort` (web), `notificationService` → `subscriptionService` (api/batch/subscriber, values.dev)
- **Envoy**: the web sidecar cluster `notification` is renamed to `subscription`, with its `SubscriptionService` routes repointed

The upcoming notification center (#2215, RFC in #2680) introduces a real `NotificationService`, so the legacy aliases can no longer squat the notification names: a flag/env var/config key with `notification` in its name that actually points at the subscription service becomes misleading — and collision-prone — once both services exist in the web binary.

## Points

- **Binary and manifests rename together in this PR**, so a build of this branch is self-consistent — deploy the chart and image from the same version and nothing changes behavior. Deploying these manifests with an image built *before* this PR (or vice versa) leaves the subscription flag falling back to its default, so chart and image must not be mixed across this boundary.
- **Deployment values overlays** that set the old keys should add the new ones before upgrading (both keys can coexist across chart generations; the unused one is ignored). Out-of-tree deployments overriding the old `BUCKETEER_*_NOTIFICATION_*` variables must rename them when upgrading. Worth a release-note entry.
- **Envoy cluster rename** resets per-cluster stats (`envoy_cluster_notification_*` → `envoy_cluster_subscription_*`); dashboards/alerts keyed on the old cluster name need updating. The Prometheus subsystem `notification` in the sender/notifier metrics is *not* touched here.
- No routing behavior change: `helm template` output verified identical modulo the renamed identifiers, and the rendered output contains no `notification` references.


hvn2k1:
prior to copilot concern, I added the subscriptionService: localhost:9001 but not delete the old one in the config

cre8ivejp (approved):
Thanks!

PR #2691, opened 2026-07-10
refactor(chart): source envoy ports from Helm values instead of hardcoding

hvn2k1:
Follow-up to review feedback on #2690 (the notification envoy cluster port was templatized there; this applies the same treatment everywhere else).

## What this PR does

The envoy configmaps hardcoded ports that the charts already expose as values and inject into the binaries — overriding such a value would desync envoy routing from the actual bind port. This templatizes them:

- **web**: all sidecar service cluster ports (`account` … `insights`, `healthcheck`) now use `env.*ServicePort`; grpc/transcoder ingress listeners use `envoy.grpcPort`/`envoy.httpPort`
- **web**: adds the missing `env.teamServicePort: 9107` value — `deployment.yaml` already referenced it, so `BUCKETEER_WEB_TEAM_SERVICE_PORT` rendered as an empty string (the binary silently fell back to its flag default)
- **api/batch**: ingress listener uses `envoy.port`
- **all charts**: envoy admin listener uses `envoy.adminPort`

Left hardcoded intentionally: cross-service cluster ports (`web`/`batch` upstreams in other charts), egress listeners (9001), and `api-rest-v1` (8000) — no same-chart value exists to bind them to.

## Points

- No behavior change with default values: rendered output of all four envoy configmaps verified **byte-identical** before/after via `helm template`.
- `BUCKETEER_WEB_TEAM_SERVICE_PORT` changes from `""` to `"9107"`, which equals the binary's flag default.

### DIFF
--- a/manifests/bucketeer/charts/api/templates/envoy-configmap.yaml
+++ b/manifests/bucketeer/charts/api/templates/envoy-configmap.yaml
@@ -19,7 +19,7 @@ data:
       address:
         socket_address:
           address: 0.0.0.0
-          port_value: 8001
+          port_value: {{ .Values.envoy.adminPort }}
     static_resources:
       clusters:
         - name: api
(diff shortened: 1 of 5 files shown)


cre8ivejp (approved):
👍

PR #2704, opened 2026-07-17
chore(devcontainer): add SSH server to devcontainer base image

Rtosshy:
## What this PR does
Reinstates the SSH server in the devcontainer base image so `gh codespace ssh` works again.

## Background

SSH used to be provided by the ghcr.io/devcontainers/features/sshd feature, which was dropped along with every other feature in a616dae2 (#2063) when the devcontainer build moved to plain Dockerfile RUN steps. This reimplements it as a plain openssh-server install, consistent with how everything else in this image is installed.

## Points

- Host keys are stripped at build time and regenerated per container at startup (`ssh-keygen -A`), so containers/Codespaces built from the shared image never share a static host key pair.

- No authorized_keys setup is needed: `gh codespace ssh` provisions per-connection keys automatically, as long as sshd is running and accepts pubkey auth.

- Also fixes real gaps found during verification: Codespaces overrides the container entrypoint (so docker-init.sh never runs there), and `postAttachCommand` only fires when an editor attaches (so a codespace accessed purely via `gh codespace ssh` never got sshd started). sshd is therefore started from `postStartCommand` (`.devcontainer/start-sshd.sh`), which runs at every container start — with or without an editor attach.

## Test plan
- [x] Local image build: `make build-devcontainer` (→ evidence 1)
- [x] Local container: sshd config, negative/positive auth, host key uniqueness across containers (→ evidence 1)
- [x] CI dry-run: full image build pipeline against a personal GHCR namespace (→ evidence 2)
- [x] End-to-end: real Codespace, connected via `gh codespace ssh` without ever attaching an editor (→ evidence 3)

### 1. Local build + container verification (`make build-devcontainer` + `docker run`)

<details>
<summary>Setup: image build + container startup (all checks below run against this)</summary>

Image build:

```bash
$ make build-devcontainer
```

This expands to `devcontainer build --workspace-folder=.github --push=false --image-name="ghcr.io/bucketeer-io/bucketeer-devcontainer:latest"`. The build itself includes an `sshd -t` self-test, so invalid `sshd_config` edits fail the build.

Container startup (the entrypoint starts sshd, so waiting for the process doubles as verifying the entrypoint wiring):

```bash
$ docker run -d --name sshtest -p 2222:2222 ghcr.io/bucketeer-io/bucketeer-devcontainer:latest
$ until docker exec sshtest pgrep -x sshd > /dev/null; do sleep 0.5; done
```

</details>

<details>
<summary>sshd config/process check</summary>

```bash
[...]

PR #2705, opened 2026-07-20
chore(devcontainer): remove unused files

Rtosshy:
## What this PR does

Removes unused dev container files identified while working on #2704.

## Background

The features configuration was previously removed when the dev container setup was migrated to plain Dockerfile steps, but the following files were left behind:

- .github/.devcontainer/local-features/setup-user/
- .github/.devcontainer/devcontainer-lock.json

The local feature is no longer referenced by devcontainer.json, the Dockerfile, or the image build workflow. The lock file only contains versions for features that are no longer declared.

## Testing

- Confirmed that the removed local feature is not referenced
- Confirmed that the current devcontainer.json does not declare any features
- Ran git diff --check
- No runtime tests performed(unused file cleanup only)

## Related PR

- #2704


Ubisoft-potato (approved):
Thanks! 👍

PR #2697, opened 2026-07-13
feat(notification): add notification center table migrations

hvn2k1:
Part of #2215 

Initiate notification center table

PR #2708, opened 2026-07-21
feat(notification): implement CreateNotification API

hvn2k1:
Part of #2215 (RFC: #2680)

## What this PR does

Implements the first notification center RPC, `CreateNotification`: system admins can create a draft announcement with one or more localizations (language, tags, title, Markdown content), persisted to MySQL or PostgreSQL.

Phase 2 of the notification center (RFC 0047). The proto contract skeleton (#2690) and table migrations (#2697) are already merged; this PR fills in the first admin API following the `team` domain as the reference implementation (`domain/` + `storage/` interface with mysql/postgres impls + `api/` handler), so the remaining nine RPCs can build on the same structure.

## Points

- The request carries `repeated NotificationLocalization`; validation requires at least one localization, non-empty language/title/content, and unique languages. Languages are not restricted to an allowlist so adding a locale later stays a data-only change.
- The draft insert omits `published_by`/`published_at`, relying on the column defaults (NULL / 0) from the migration.
- Tags are serialized to the JSON column via the shared `JSONObject` helper in both storage implementations.
- Per the RFC, no domain/audit events are published for v1 — the `created_by`/`last_edited_by` columns are the audit trail.
- The other nine RPCs still return `Unimplemented`; they land in follow-up PRs.


hvn2k1:
The copilot review is false positive, when the last matcher in an expectation is gomock.Any() and the mocked method is variadic, it matches all remaining arguments regardless of count.

Ubisoft-potato (approved):
LGTM! 👍

PR #2710, opened 2026-07-21
feat(notification): implement ListDraftAdminNotifications API

hvn2k1:
Part of #2215 (RFC: #2680), Follows #2708

## What this PR does

Implements the `ListDraftAdminNotifications` RPC: system admins can list draft announcements with keyword search, sorting (`created_at`/`updated_at`, asc/desc), offset-based pagination, and a total count, backing the drafts panel of the "Publish notification" tab.

Continues the notification center backend (RFC 0047) on top of the CreateNotification API (#2708). Drafts need their own admin-only listing since the viewer-facing `ListNotifications` will never return them. The RPC is named `ListDraftAdminNotifications` (rather than `ListDraftNotifications`) to make the admin-facing surface explicit next to the upcoming viewer-facing RPCs.

## Points

- Keyword search matches localization titles and contents. Since drafts join to multiple localization rows, the list query uses `LEFT JOIN notification_localization` with `SELECT DISTINCT` (and `COUNT(DISTINCT notification.id)`) instead of an `EXISTS` subquery — the postgres `ExistsFilter` helper doesn't support bind parameters, and this keeps both backends on the shared query helpers.
- Localizations are attached in a second `WHERE notification_id IN (...)` query over the returned page only, avoiding row explosion in the main query; admins receive all localizations per the RFC's editor view.
- `page_size` defaults to 200 when unset and is capped at 200 (`maxNotificationPageSize`), following the auditlog convention; negative page sizes and cursors are rejected as invalid arguments.
- `DEFAULT` order maps to `created_at` (drafts have no `published_at` to sort by).
- Filter construction lives in `listDraftAdminNotificationsFilters`, mirroring the orders helper, in both storage backends.
- The MySQL and PostgreSQL implementations differ only where their conventions differ: MySQL appends WHERE/ORDER/LIMIT via `ConstructQueryAndWhereArgs`, PostgreSQL uses `%s` template placeholders in the SQL files, matching the `team` storage in each backend.


pkg/notification/storage/mysql/notification.go
```diff
@@ -78,3 +86,154 @@ func (s *notificationStorage) CreateNotification(
...
+
+func (s *notificationStorage) ListDraftNotifications(
+	ctx context.Context,
+	p notificationstorage.ListDraftNotificationsParams,
+) ([]*proto.Notification, int, int64, error) {
+	orders, err := listDraftNotificationsOrders(p.OrderBy, p.OrderDirection)
+	if err != nil {
+		return nil, 0, 0, err
+	}
+	filters := []*mysqlstorage.FilterV2{
```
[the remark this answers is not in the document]
hvn2k1: Good catch. Fixed in 3ad594cd1: negative page sizes are now clamped to 0 (no limit) before the `ListOptions` and the slice allocation, following the same convention as the coderef storage (`pkg/coderef/storage/mysql/code_reference.go`). Added a test case covering `PageSize: -1`.

pkg/notification/storage/mysql/notification.go
```diff
@@ -78,3 +86,154 @@ func (s *notificationStorage) CreateNotification(
...
+		}
+	}
+	cursor := p.Cursor
+	if cursor == "" {
+		cursor = "0"
+	}
+	offset, err := strconv.Atoi(cursor)
+	if err != nil {
+		return nil, 0, 0, notificationstorage.ErrInvalidListDraftNotificationsCursor
+	}
```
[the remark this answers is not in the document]
hvn2k1: Fixed in 3ad594cd1: the cursor check is now `if err != nil || offset < 0`, returning `ErrInvalidListDraftNotificationsCursor` (mapped to InvalidArgument in the API layer), matching the existing coderef/autoops/experiment storages. Added a test case for `Cursor: "-1"`.

pkg/notification/storage/postgres/notification.go
```diff
@@ -78,3 +87,159 @@ func (s *notificationStorage) CreateNotification(
...
+
+func (s *notificationStorage) ListDraftNotifications(
+	ctx context.Context,
+	p notificationstorage.ListDraftNotificationsParams,
+) ([]*proto.Notification, int, int64, error) {
+	orders, err := listDraftNotificationsOrders(p.OrderBy, p.OrderDirection)
+	if err != nil {
+		return nil, 0, 0, err
+	}
+	filters := []*pgstorage.Filter{
```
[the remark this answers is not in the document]
hvn2k1: Good catch. Fixed in 3ad594cd1: negative page sizes are now clamped to 0 (no limit) before the `ListOptions` and the slice allocation, following the same convention as the coderef storage (`pkg/coderef/storage/postgres/code_reference.go`). Added a test case covering `PageSize: -1`.

pkg/notification/storage/postgres/notification.go
```diff
@@ -78,3 +87,159 @@ func (s *notificationStorage) CreateNotification(
...
+		}
+	}
+	cursor := p.Cursor
+	if cursor == "" {
+		cursor = "0"
+	}
+	offset, err := strconv.Atoi(cursor)
+	if err != nil {
+		return nil, 0, 0, notificationstorage.ErrInvalidListDraftNotificationsCursor
+	}
```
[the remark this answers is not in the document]
hvn2k1: Fixed in 3ad594cd1: the cursor check is now `if err != nil || offset < 0`, returning `ErrInvalidListDraftNotificationsCursor` (mapped to InvalidArgument in the API layer), matching the existing coderef/autoops/experiment storages. Added a test case for `Cursor: "-1"`.

pkg/notification/api/api.go
```diff
@@ -144,7 +145,36 @@ func (s *NotificationService) ListDraftNotifications(
...
-	return nil, statusNotImplemented
+	_, err := s.checkSystemAdminRole(ctx)
+	if err != nil {
+		return nil, err
+	}
+	params := storage.ListDraftNotificationsParams{
+		SearchKeyword:  req.SearchKeyword,
+		OrderBy:        req.OrderBy,
+		OrderDirection: req.OrderDirection,
+		PageSize:       int(req.PageSize),
```
Ubisoft-potato: If `page_size` is omitted (proto zero value), `PageSize` stays `0`, which the shared query builders treat as "no `LIMIT` clause" — so the request returns all drafts in one response. 

There is also no upper bound on large values. Consider defaulting/capping it like `pkg/auditlog/api/api.go` does.

hvn2k1: Fixed in 6208a6d42: `page_size` now defaults to 200 when unset and is capped at 200, following the auditlog convention (`maxAuditLogPageSize` in `pkg/auditlog/api/api.go`). Added test cases for both the default and the cap. Note the RPC was also renamed to `ListDraftAdminNotifications` in 9cd185a68 to make the admin-facing surface explicit.

proto/notification/service.proto
```diff
@@ -258,8 +258,8 @@ service NotificationService {
     };
   }
 
-  rpc ListDraftNotifications(ListDraftNotificationsRequest)
-      returns (ListDraftNotificationsResponse) {
+  rpc ListDraftAdminNotifications(ListDraftAdminNotificationsRequest)
+      returns (ListDraftAdminNotificationsResponse) {
     option (google.api.http) = {
       get: "/v1/notifications/drafts"
```
cre8ivejp: The REST path must be updated, too. Otherwise, it could conflict with the APIs planned for the next phase.

hvn2k1: I updated the url, thank you

hvn2k1: https://github.com/bucketeer-io/bucketeer/pull/2710/commits/afc249b890032b6ec10d2efba5e15115750d5d62

Ubisoft-potato (commented):
LGTM!

cre8ivejp (approved):
Thank you!

PR #2712, opened 2026-07-24
feat: implement onboard walkthrough for new user

### LINKED ISSUE #2514: feat: improve the onboarding process
### Describe the feature you'd like

For new users, it's important to have onboarding animation when they first visit the web console.
- [ ] Onboarding animation
- [ ] Check and improve the onboarding document, so user can fasten the process to understand the product, how to create flag, flag life cycle, ...

### Background

_No response_

### Alternative solutions

_No response_

hvn2k1:
Resolves https://github.com/bucketeer-io/bucketeer/issues/2514

## What this PR does

Adds an animated, in-console onboarding walkthrough that guides new users through the core journey — **create a feature flag → create an API key → connect an SDK** — using spotlight overlays on the real UI (powered by [driver.js](https://driverjs.com/), ~5 kB, MIT).

### Entry points

- A **Get Started** button pinned at the bottom of the sidebar (new rocket icon) starts the walkthrough at any time.
- **First-time users**: the walkthrough starts automatically on first login, detected via `account.lastSeen == 0` from `GetMe`.

### Walkthrough flow

1. Spotlight on the Feature Flags menu, then on the **+ Create Flag** button.
2. Guides through the create-flag form (general info, variations), step by step.
3. Steps are **action-driven** where possible: they advance when the user actually clicks the real button (`advanceOnClick`), not a "Next" button — users learn by doing in their own console.
4. After the flag is created, a celebration dialog leads to the API Keys page and guides key creation the same way.
5. Once the key exists (and the key-secret dialog is closed), a **Connect an SDK** modal shows install + initialize/evaluate starter code for all 8 supported SDKs (JavaScript, React, React Native, Android, iOS, Flutter, Go, Node.js), with the API endpoint, flag ID, and feature tag pre-filled. The API key is intentionally shown as a `YOUR_API_KEY` placeholder — the real secret is only displayed in the dedicated secret dialog.

### UX guardrails

- Every step has a **Skip walkthrough** button (except the final one); clicking the dimmed overlay or pressing Escape does not exit the tour.
- Portaled dropdowns (environment selector, tags) stay clickable above the walkthrough overlay.
- The API-key submit button is locked (dimmed) until its dedicated step; Enter-key submits are blocked on the form step.
- Full English and Japanese translations.

### Configuration

- Helm: `web.env.walkthroughEnabled` (default `true`) → rendered into `window.env.WALKTHROUGH_ENABLED`, hides the Get Started entry when `false`.
- Docker Compose (`env.js`) and Vite (`VITE_WALKTHROUGH_ENABLED`) equivalents included.

### Backend change

`GetMe` now returns the **pre-update** `lastSeen` (while still bumping it), restoring the first-login signal that #2185 unintentionally removed — previously `lastSeen` always read as "just now", so clients could never detect a first login (this had also broken the existing first-login profile form).

## How to test

1. Click **Get Started** at the bottom of the sidebar and follow the tour end to end (flag → API key → SDK snippet).
2. First-login auto-start: reset an account's `last_seen` to `0` in the DB (requires a rebuilt backend), then sign in.
3. Toggle off: set `web.env.walkthroughEnabled: false` (or `WALKTHROUGH_ENABLED: false` in `env.js`) and confirm the Get Started entry disappears.

## demo

https://github.com/user-attachments/assets/cbe5f480-46a9-4045-962e-9021379be55e


ui/dashboard/src/hooks/index.ts
```diff
@@ -6,3 +6,4 @@ export * from './use-screen';
 export * from './use-action-with-url';
 export * from './useTokenRefresh';
 export * from './use-unsaved-leave-page';
+export * from './use-walkthrough';
```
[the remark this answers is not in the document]
hvn2k1: Moved `WALKTHROUGH_TARGETS` to a side-effect-free `constants/walkthrough.ts` and updated all consumers, so pages no longer pull driver.js (and its CSS) through the hooks barrel. The barrel keeps exporting the hook itself for the navigation. Fixed in 77c515611.

ui/dashboard/src/hooks/use-walkthrough.tsx
```diff
@@ -0,0 +1,343 @@
...
+          setStage(current => (current === tourStage ? 'idle' : current));
+        }
+      }
+    }),
+    [t]
+  );
+
+  const startApiKeyTour = useCallback(() => {
+    const currentEnvironment = getCurrentEnvironment(consoleAccount!);
+
```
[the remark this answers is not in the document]
hvn2k1: Replaced the non-null assertions with guards in `startWalkthrough`, `startApiKeyTour`, and the polling effect. The auto-start effect now also waits for `consoleAccount` before consuming the pending flag, so the walkthrough isn't lost if auth is still loading. Fixed in 77c515611.

ui/dashboard/src/hooks/use-walkthrough.tsx
```diff
@@ -0,0 +1,343 @@
...
+        });
+        popover.footerButtons.insertBefore(skipButton, popover.nextButton);
+      },
+      onCloseClick: (_element, _step, { driver: driverInstance }) => {
+        cancelledRef.current = true;
+        driverInstance.destroy();
+      },
+      onDestroyed: (_element, step) => {
+        // Ending on the last step's action click hands off to the next stage.
+        if (!cancelledRef.current && step?.element === lastStepTarget) {
```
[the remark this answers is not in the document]
hvn2k1: Added `document.removeEventListener('keydown', blockEnterSubmit, true)` to `onDestroyed` as an idempotent safety net covering every exit path. Fixed in 77c515611.

ui/dashboard/src/elements/connect-sdk-modal/snippets.ts
```diff
@@ -0,0 +1,281 @@
...
+    installLanguage: 'shell',
+    codeLanguage: 'go',
+    install: () => 'go get github.com/bucketeer-io/go-server-sdk',
+    code: v => `import (
+  "github.com/bucketeer-io/go-server-sdk/pkg/bucketeer"
+  "github.com/bucketeer-io/go-server-sdk/pkg/bucketeer/user"
+)
+
+ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+defer cancel()
```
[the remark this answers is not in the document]
hvn2k1: Added `context`, `log`, and `time` to the Go snippet's import block so it compiles as pasted. Fixed in 77c515611.

ui/dashboard/src/hooks/use-walkthrough.tsx
```diff
@@ -0,0 +1,350 @@
...
+        },
+        {
+          element: tourTarget(WALKTHROUGH_TARGETS.CREATE_APIKEY_BUTTON),
+          advanceOnClick: true,
+          popover: {
+            title: t('walkthrough.create-apikey-button.title'),
+            description: t('walkthrough.create-apikey-button.description'),
+            side: 'bottom',
+            showButtons: ['close']
+          }
```
t-kikuc, line 232: The walkthrough ends after showing SDK snippets, but the newly created flag is likely in a disabled state at that point. A user who copies the SDK code and runs it will get only the default value, which may look like a broken integration.

Consider adding a step (or a note in the SDK modal) that guides the user to enable the flag on the targeting page. The walkthrough already navigates to the targeting page after flag creation, so this would be a natural addition.

ui/dashboard/src/components/navigation/index.tsx
```diff
@@ -193,7 +200,17 @@ const Navigation = ({ onClickNavLink }: { onClickNavLink: () => void }) => {
...
+        {WALKTHROUGH_ENABLED && (
+          <MenuItemComponent
+            icon={IconSystem.IconRocket}
+            label={t(`navigation.get-started`)}
+            onClick={() => {
+              onClickNavLink();
+              startWalkthrough();
+            }}
+          />
+        )}
```
t-kikuc: Two suggestions on the "Get Started" entry:

**1. Visibility:** This button is shown to _all_ users at all times (when `WALKTHROUGH_ENABLED` is true), including experienced users who don't need it. Consider moving it to the Settings page (e.g. under General) as a "Replay tutorial" action, so it remains discoverable without permanently occupying sidebar space.

**2. Label:** "Get Started" is generic and doesn't convey what it does. Something like "Tutorial" / "チュートリアル" or "Onboarding" / "オンボーディング" would be more descriptive.

ui/dashboard/src/components/navigation/index.tsx
```diff
@@ -193,7 +200,17 @@ const Navigation = ({ onClickNavLink }: { onClickNavLink: () => void }) => {
           </div>
         </div>
 
-        <Divider className="mb-3 bg-primary-50 opacity-10" />
+        {WALKTHROUGH_ENABLED && (
+          <MenuItemComponent
+            icon={IconSystem.IconRocket}
+            label={t(`navigation.get-started`)}
+            onClick={() => {
+              onClickNavLink();
```
steveninhle: When startWalkthrough() is called at the `Settings panel ` navigated to the Feature Flags page, but the Settings panel is never closed, causing the tour's highlight and popover to be positioned incorrectly — first pointing at elements that weren't visible

ui/dashboard/src/hooks/use-walkthrough.tsx
hvn2k1: Good point — done in fe8e18fe7. The SDK modal now shows a note that the new flag is OFF by default, and its primary button ("Go to the flag") lands on the flag's targeting page so the user can enable it right away.

ui/dashboard/src/components/navigation/index.tsx
hvn2k1: Done in fe8e18fe7 — removed the permanent sidebar entry and added a "Replay tutorial" action to Settings > General instead (the walkthrough state moved into a shared WalkthroughProvider so the Settings page can start it). First-time users still get the walkthrough automatically on first login.

ui/dashboard/src/components/navigation/index.tsx
hvn2k1: Nice catch — fixed in fe8e18fe7. The sliding settings panel now syncs with the current path (opens for settings-area pages, closes otherwise), so starting the tour from the settings area no longer leaves the panel covering the highlighted elements.

ui/dashboard/src/pages/settings/page-content.tsx
```diff
@@ -206,6 +210,25 @@ const PageContent = ({ organization }: { organization: Organization }) => {
...
+            {t('walkthrough.tutorial')}
+          </p>
+          <p className="typo-para-small text-gray-600 mt-2">
+            {t('walkthrough.replay-description')}
+          </p>
+          <Button
+            type="button"
+            variant="secondary"
+            className="mt-4"
+            onClick={() => startWalkthrough()}
```
steveninhle: I noticed a couple of things while clicking the "Replay tutorial" button:

1. If there are unsaved changes on the page, clicking "Replay tutorial" is blocked by the leave-page confirmation dialog, so the tutorial never starts.
2. The walkthrough popover doesn't seem to update its position after the modal opens. It stays misaligned until the user interacts with the page(e.g, scrolls)

Could you check if these are expected?

hvn2k1: Thanks, both were real issues — fixed in e5d20277b:

1. `startWalkthrough` no longer starts driving immediately: if you're not on the flag list page it navigates first and only starts the tour once the page is actually reached. So with unsaved changes, the leave-page dialog appears as usual — choosing **Stay** simply doesn't start the tour (the pending request expires after 30s), choosing **Leave** starts it normally.

2. The popover misalignment happened because driver.js measures the target once, while panels/modals are still animating in. The walkthrough now refreshes the spotlight position on `transitionend`/`animationend` (same mechanism as the nested-scroll refresh), so it snaps into place as soon as the animation settles.

hvn2k1: thank you, I just fix that, can you help me check https://github.com/bucketeer-io/bucketeer/pull/2712/commits/e5d20277b126f03130a3194f577c9ddc509550a6

steveninhle: Thanks for fixing this! I retested it and found one minor UI issue that still reproduces.
Steps to reproduce (all within 30 seconds of step 2, per PENDING_TOUR_TIMEOUT):

1. Edit a field on the create/edit flag form so it becomes dirty (unsaved changes).
2. Click Restart tutorial.
3. In the "unsaved changes" confirmation dialog, click Continue editing (cancel the navigation).
4. Click a Settings-menu item in the sidebar (e.g., Organization) to navigate away from the form.
5. In the "unsaved changes" confirmation dialog that appears again, click Leave page this time.
6. Click Back to Main in the sidebar.
7. The walkthrough tour auto-starts.
8. Click Skip walkthrough.

Result: The UI breaks

ui/dashboard/src/components/navigation/index.tsx
steveninhle, line 192: I noticed a UI issue while testing.

Clicking **Tutorial** from the **Settings** menu correctly returns to the **main menu** and starts the walkthrough. However, after clicking **Skip** on the first step, the sidebar switches back to the **Settings** panel instead of remaining on the **main menu**.

I tried this change locally, and it seems to resolve the issue. Would it make sense to use something like this?

```suggestion
className={cn(
'w-full absolute ease-in-out transition-[left,right] duration-500 -left-[100%] invisible',
{ 'left-0 visible': !isOpenSetting }
)}
```

hvn2k1: I move the start tutorial button to user setting, and it fixes this problem too
image

ui/dashboard/src/hooks/use-walkthrough.tsx
```diff
@@ -0,0 +1,466 @@
...
+
+  useEffect(() => {
+    const pending = pendingTourRef.current;
+    if (!pending || !consoleAccount) return;
+    if (Date.now() - pending.requestedAt > PENDING_TOUR_TIMEOUT) {
+      pendingTourRef.current = null;
+      return;
+    }
+    const currentEnvironment = getCurrentEnvironment(consoleAccount);
+    if (pathname === `/${currentEnvironment.urlCode}${PAGE_PATH_FEATURES}`) {
```
steveninhle, line 466: I found a scenario where the pending walkthrough request isn't cleared if the user cancels the navigation due to unsaved changes. As a result, the walkthrough can unexpectedly start later when the user eventually returns to the Feature Flags page. If the user then clicks Skip walkthrough, the tour is torn down mid-transition, and the navigative UI breaks.
### Repro

1. Switch Setting menu.
2. Edit a field on the create/edit flag form so it becomes dirty (unsaved changes).
3. Click **Restart tutorial**.
4. In the "unsaved changes" confirmation dialog, click **Continue editing** (cancel the navigation).
5. Click a Settings-menu item in the sidebar (e.g. Organization) to navigate away from the form.
6. In the "unsaved changes" confirmation dialog that appears again, click **Leave page**.
7. Click **Back to Main** in the sidebar.
8. The walkthrough tour auto-starts unexpectedly.
9. Click **Skip walkthrough**

hvn2k1: Nice catch, thanks for the detailed repro — fixed in 31d226524.

Two changes:
1. Cancelling the leave-page dialog ("Continue editing") now dispatches an event that immediately drops the pending tour request, so it can't fire later.
2. As a safety net, the first navigation of any kind consumes the pending request — the tour starts only if that navigation lands directly on the flag list; any other destination discards it.

Verified against your repro: after step 4 the request is already cleared, so step 8's return to the flag list no longer auto-starts the tour. Choosing **Leave page** on the tutorial's own dialog still starts the tour as expected, since that navigation goes straight to the flag list.

ui/dashboard/src/index.css
```diff
@@ -145,3 +145,51 @@
...
+
+body.driver-active [data-radix-popper-content-wrapper] {
+  z-index: 10001 !important;
+}
+body.driver-active [data-radix-popper-content-wrapper],
+body.driver-active [data-radix-popper-content-wrapper] * {
+  pointer-events: auto;
+}
+
+body.driver-active .driver-active-element [data-tour='submit-apikey-button'] {
```
steveninhle, line 160: Overall, it looks good! Only one small CSS issue.
```suggestion
body.driver-active .driver-active-element[data-tour='submit-apikey-button'] {
```

hvn2k1: Thanks for the close look! The space is actually intentional — this needs to be a **descendant** selector, not a compound one:

- On the *form step*, driver.js puts `driver-active-element` on the form container, and the submit button inside it matches `.driver-active-element [data-tour=…]` → locked/dimmed, so users can't submit before reaching the dedicated step.
- On the *submit step*, the class moves onto the button itself; the descendant selector stops matching (no ancestor carries the class) → the button unlocks exactly when it should be clicked.

With the compound form `.driver-active-element[data-tour=…]` the behavior inverts: the button stays clickable during the form step and gets locked on its own step, which would deadlock the tour.

Added a comment above the rule in d18c78fff so the intent is clear to future readers.

steveninhle: Got it, thanks for the explanation!

ui/dashboard/src/hooks/use-walkthrough.tsx
```diff
@@ -0,0 +1,482 @@
...
+      showProgress: true,
+      overlayOpacity: 0.6,
+      stagePadding: 6,
+      stageRadius: 8,
+      // Steps may target elements on a page the previous step navigates to.
+      waitForElement: 5000,
+      overlayClickBehavior: () => undefined,
+      allowKeyboardControl: false,
+      nextBtnText: t('walkthrough.next'),
+      doneBtnText: t('walkthrough.done'),
```
steveninhle, line 143: I also noticed a couple of minor issues while testing:

1. Some text is still missing i18n support, such as the previous button and the progress text.
2. The walkthrough dialog could be slightly wider to accommodate both the English and Japanese versions better.

Image

cre8ivejp (dismissed):
Thank you! Nice work!

ui/dashboard/src/hooks/use-walkthrough.tsx
hvn2k1: thank you, I fixed in https://github.com/bucketeer-io/bucketeer/pull/2712/commits/d54291a1dbbbbe3bc4f9ca2f4681ef11d4a7f6a5

steveninhle (approved):
LGTM. Thanks!

PR #2734, opened 2026-07-30
fix(subscriber): discard goal events older than the user's latest evaluation

cre8ivejp:
## Summary

The client SDK sets the event timestamps when evaluation and goal events are generated. If a goal event's timestamp is older than the user's latest evaluation timestamp, the app tracked the goal before evaluating the user (incorrect SDK usage), so the event can never be linked. Pub/Sub reordering cannot cause this case: reordering only delays delivery, it doesn't change the timestamps inside the payloads.

The goal event DWH subscriber handled this condition inconsistently:

- **Fresh delivery**: the event was acked and dropped silently, while being
  counted in the `Linked` metric — it looked like a successful link but
  nothing was written to the DWH.
- **Retry path**: the message was re-queued with backoff and retried until it
  expired (12–24h), even though retrying can never succeed because the query
  always returns the user's latest evaluation and timestamps only grow.

## What

- Both paths now discard the event explicitly: the fresh path acks it, the
  retry path deletes the Redis key, each with a warning log and the
  `GoalEventIssuedBeforeEvaluation` metric.
- Renamed the sentinel error to `ErrGoalEventOlderThanEvaluation`.
- The linker no longer stores a retry message when called from the retry
  processor, which previously reset the backoff retry count.
- The legitimate out-of-order Pub/Sub case (evaluation row not yet in the
  DWH) is **unchanged**: it still goes to the Redis retry queue and links on
  a later pass.
- Added unit tests (`goal_events_dwh_test.go`) covering the link, discard,
  and retry/re-queue paths.

### DIFF
--- a/pkg/subscriber/processor/errors.go
+++ b/pkg/subscriber/processor/errors.go
@@ -30,11 +30,18 @@ var (
 	ErrReasonNil                                 = errors.New("eventpersister: reason is nil")
 	ErrEvaluationsAreEmpty                       = errors.New("eventpersister: evaluations are empty")
 	ErrEvaluationEventIssuedAfterExperimentEnded = errors.New("eventpersister: evaluation event issued after experiment ended") //nolint:lll
-	ErrFailedToEvaluateUser                      = errors.New("eventpersister: failed to evaluate user")
-	ErrAutoOpsRuleNotFound                       = errors.New("eventpersister: auto ops rule not found")
-	ErrFeatureEmptyList                          = errors.New("eventpersister: list feature returned empty")
-	ErrFeatureVersionNotFound                    = errors.New("eventpersister: feature version not found")
-	ErrUnknownEvent                              = errors.New("metricsevent persister: unknown metrics event")
-	ErrInvalidDuration                           = errors.New("metricsevent persister: invalid duration")
-	ErrUnknownApiId                              = errors.New("metricsevent persister: unknown api id")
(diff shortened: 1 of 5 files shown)


Ubisoft-potato:
Nit: `codeRetryMessageNoEvaluations` in `pkg/subscriber/processor/metrics.go:49` is now dead code. 
It was only incremented in the `len(evals) == 0` branch of `handleNewRetry` that this PR removes (since `linkGoalEventByExperiment` no longer returns that state without an error — it now returns `ErrGoalEventOlderThanEvaluation` explicitly). 

Worth deleting the constant in this PR or a fast follow-up.

cre8ivejp:
> Nit: `codeRetryMessageNoEvaluations` in `pkg/subscriber/processor/metrics.go:49` is now dead code. It was only incremented in the `len(evals) == 0` branch of `handleNewRetry` that this PR removes (since `linkGoalEventByExperiment` no longer returns that state without an error — it now returns `ErrGoalEventOlderThanEvaluation` explicitly).
> 
> Worth deleting the constant in this PR or a fast follow-up.

@Ubisoft-potato, I have deleted it https://github.com/bucketeer-io/bucketeer/pull/2734/commits/afc6f760b5eff8699d0a784dc26759a96813a58d.

Ubisoft-potato (approved):
Thanks! 👍

PR #2736, opened 2026-07-30
fix(subscriber): retain in-memory buffers when periodic writes fail

cre8ivejp:
## Why

`TestFeatureAutoArchiver_RecentlyUsedNotArchived` (and the other
`TestFeatureAutoArchiver` tests) fail intermittently in the dev cluster with:

    LastUsedInfo not recorded for feature ... after 60 retries

The test registers a single evaluation event and polls `GetFeature` for
120 seconds waiting for `LastUsedInfo`. Normally the data lands within
~20 seconds (`flushInterval` + `writeCacheInterval: 10`), so this is not a
timing problem — the data was being lost.

Investigating the `evaluationCountEventPersister` subscriber revealed the
same data-loss pattern in two of its three periodic writers. The subscriber
acks the Pub/Sub messages as soon as the evaluation counts are written to
Redis, so the in-memory buffers are the **only copy** of the remaining data.
If a periodic flush fails and the buffer is discarded, that data is lost
permanently — there is no redelivery.

## Fix 1: Feature last-used info (MySQL)

`writeEnvLastUsedInfo` flushes the in-memory last-used info buffer to MySQL
every `writeCacheInterval` seconds. When the upsert failed for an
environment (e.g. a transient MySQL deadlock, common in the dev cluster
while parallel e2e suites write to `feature_last_used_info`), the error was
logged and the environment skipped — but the entire buffer was
unconditionally reset right after the loop, discarding the failed
environment's entries too:

    if err := p.upsertMultiFeatureLastUsedInfo(...); err != nil {
        p.logger.Error(...)
        continue
    }
    ...
    // Reset the cache
    p.envLastUsedCache = make(environmentLastUsedInfoCache)

One failed flush permanently lost the last-used info until the next
evaluation event arrived for the same flag. Since the e2e test sends exactly
one event, `LastUsedInfo` never appeared and the test timed out.

Beyond the flaky test, this affects production: silently lost last-used
info skews `UnusedDays`, which feeds auto-archive and stale-flag detection.

**Fix:** remove an environment from the buffer only after a successful
write. Failed environments are kept and retried on the next cycle — the
same pattern `writeUserAttributes` already uses. Retrying is safe because
the upsert merges with the existing DB row and only updates when
`LastUsedAt` is newer.

## Fix 2: DAU buffer (Redis HyperLogLog)

`writeDAU` had the same pattern: the in-memory DAU buffer was swapped out
*before* the Redis write, and on failure the error was only logged — the
swapped-out buffer was discarded, permanently losing those users' DAU
records. `RecordDAUBatch` also fails fast on the first Redis error, so all
records after the failing key in the batch were dropped as well.

**Fix:** merge the buffer back on failure and retry on the next cycle.
Re-sending already-recorded user IDs is safe because `PFADD` (HyperLogLog)
is idempotent, so there is no double counting.

## Audited but not changed

- `writeUserAttributes` already handles failures correctly: it removes
  entries from its cache only after a successful write.
- The evaluation counts themselves are flushed synchronously with proper
  ack/nack handling, so failures there are redelivered by Pub/Sub.

## Tests

Added unit tests covering:

- A failed last-used info flush keeps the buffer; the next cycle writes it
  and clears it.
- With one failing and one succeeding environment, only the successful one
  is removed from the buffer.
- A failed DAU flush restores the buffer; entries buffered in the meantime
  are merged, and the next cycle flushes both together.

## Impact on the e2e tests

Worst case for `TestFeatureAutoArchiver_*` is now one extra flush cycle
(~10s), well within the 120s polling budget, so no test-side changes are
needed.

### DIFF
--- a/pkg/subscriber/processor/evaluation_events_evaluation_count_event_persister.go
+++ b/pkg/subscriber/processor/evaluation_events_evaluation_count_event_persister.go
@@ -696,17 +696,23 @@ func (p *evaluationCountEventPersister) writeEnvLastUsedInfo() {
 			info = append(info, v)
 		}
 		if err := p.upsertMultiFeatureLastUsedInfo(context.Background(), info, environmentId); err != nil {
-			p.logger.Error("Failed to write feature last-used info", zap.Error(err),
+			// Keep the entries so they are retried on the next cycle.
+			// The Pub/Sub messages were already acked when the evaluation counts
+			// were written to Redis, so dropping the entries here would lose the
+			// last-used info until the next evaluation event arrives for the flag.
+			p.logger.Error("Failed to write feature last-used info, will retry next cycle",
+				zap.Error(err),
 				zap.String("environmentId", environmentId))
 			continue
 		}
+		// Remove only the environments that were successfully written.
+		// The failed ones will remain for the next attempt.
+		delete(p.envLastUsedCache, environmentId)
 		p.logger.Debug("Cache has been written",
 			zap.String("environmentId", environmentId),
 			zap.Int("cacheSize", len(info)),
 		)
 	}
-	// Reset the cache
-	p.envLastUsedCache = make(environmentLastUsedInfoCache)
 }
 
 func (p *evaluationCountEventPersister) upsertMultiFeatureLastUsedInfo(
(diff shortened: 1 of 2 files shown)


Ubisoft-potato (approved):
Thanks! 👍

PR #2740, opened 2026-07-30
fix: publish segment domain events after commit to prevent stale cache refresh

cre8ivejp:
## Summary

Fixes the flaky e2e test `TestGrpcGetSegmentUsersRuleBasedSegment` ("updated segment rules should be delivered through the diff sync"), which fails in the dev container but passes in the GCP dev cluster.

### Root cause

`UpdateSegment` (and the other segment mutations) published the `SEGMENT_UPDATED` domain event **inside** the MySQL transaction, before commit. In the dev container, redis-stream pub/sub delivers the event to the `cacheRefresher` subscriber within milliseconds — often before the transaction commits — so the refresher re-read MySQL, saw the **pre-update** segment, and overwrote the segment-users Redis (L2) cache with stale rules and a stale `updated_at`. The cache-invalidation announcement then evicted the API pod's correct L1 entry and replaced it with the stale L2 data on the next read.

Recovery required the every-minute batch `SegmentUserCacher` **plus** the 1-minute L1 in-memory TTL to expire (the batch cacher does not publish invalidations), which in the worst case exceeds the test's 60s `Eventually` window. In the GCP dev cluster, Google Pub/Sub delivery latency exceeds the commit time, so the race never manifests there.

### Changes

`CreateSegment` / `UpdateSegment` / `DeleteSegment` now build the domain event inside the transaction (where the previous entity state is available) but publish it only **after** the transaction commits, matching the existing `UpdateFeature` / `CloneFeature` convention. With the event published post-commit, the cache refresher always reads committed state, so the stale overwrite can no longer happen.

Also fixes two pre-existing bugs where a `domainevent.NewEvent` failure was swallowed with `return nil` (in `DeleteSegment` this returned success without deleting the row).

### Behavior note

A mutation whose commit succeeds but whose post-commit publish fails now returns an error to the caller while the DB change persists — same semantics as `UpdateFeature`. The every-minute batch cachers reconcile the caches in that scenario.

## Test plan

- [x] `go test ./pkg/feature/api/... ./pkg/subscriber/processor/...` passes
- [x] New regression tests:
  - `TestSegmentDomainEventPublishedAfterCommitMySQL` asserts `Publish` fires only after the transaction returns for all three mutations
  - `TestUpdateSegmentPublishFailureMySQL` asserts the request fails and the cache refresh is skipped when the publish fails
- [x] Re-run `make e2e` in the dev container to confirm `TestGrpcGetSegmentUsersRuleBasedSegment` no longer flakes


Ubisoft-potato:
`BulkUploadSegmentUsers` in `pkg/feature/api/segment_user.go` publishes its `SegmentBulkUploadUsersEvent` domain event before the transaction commits, same issue this PR fixes in `segment.go`. We might need to apply the same fix to `segment_user.go`.

Ubisoft-potato (approved):
Thanks! 👍

PR #2741, opened 2026-07-31
chore(claude): add dev container skills for Claude Code

hvn2k1:
## What this PR does

Adds three Claude Code project skills — `devcontainer-run`, `devcontainer-generate`, `devcontainer-deploy` — that let Claude detect the running Bucketeer dev container (local devcontainer or GitHub Codespace) and run development actions inside it, and updates `.claude/CLAUDE.md` to prefer the dev container + minikube flow for those actions.

Codegen and deploys are environment-sensitive: generated files must be produced with exactly protoc v23.4, minikube must be started via `make start-minikube`, and the `*.bucketeer.io` hosts entries only resolve inside the container. When Claude (or a human) runs these on the host, a protoc version mismatch silently churns every generated `.pb.go` file, and bare `kubectl`/`helm` can hit the wrong cluster — the host kubectl context may point at a real GKE cluster instead of minikube. These skills route all such actions through the dev container, where the toolchain is guaranteed.

## Points

- `devcontainer-run/scripts/exec.sh` detection order: already inside the container → local devcontainer (docker label `devcontainer.local_folder`) → GitHub Codespace (`gh codespace ssh`). When nothing is found it exits 2 with instructions to start one — deliberately **no silent fallback to the host**.
- Ambiguity is never resolved by guessing: the local-devcontainer match is an exact `devcontainer.local_folder` label match against the repo root, and if several Bucketeer Codespaces are available the script exits 2 with the list and asks for `BUCKETEER_CODESPACE=<name>` rather than taking the first.
- Every cluster command names its context explicitly (`kubectl --context minikube`, `helm --kube-context minikube`) and `status` warns when the active context is not `minikube`, so a stray `kubectl config use-context` cannot silently redirect a deploy at a real cluster.
- The status health line counts only genuinely failing pods (CrashLoopBackOff, ImagePull errors, OOMKilled, Evicted) rather than a raw ready count, because batch CronJobs constantly spawn short-lived pods and make a ready count flap.
- Codespace mode is flagged as a separate clone (generated files need `gh codespace cp` or a push to get back to the host); a local devcontainer is a bind mount, so no sync is needed. Verification steps therefore run through `$DEVC` too, so they inspect the clone where generation actually happened.
- Verified against a live devcontainer: `make proto-all` run through the skill reproduced the committed generated files byte-for-byte (zero diff), and the single-service rollout commands were validated against the running minikube cluster.
- Docker Compose stays documented in CLAUDE.md as the host-based alternative; build/test/lint commands are unchanged.


.claude/skills/devc/scripts/devc-exec.sh
```diff
@@ -0,0 +1,120 @@
...
+    command -v mockgen >/dev/null && echo "go-tools: OK" || echo "go-tools: MISSING (run bash .devcontainer/setup.sh)"
+    docker info >/dev/null 2>&1 && echo "dockerd: running" || echo "dockerd: NOT running (start with: nohup sudo dockerd > /tmp/dockerd.log 2>&1 &)"
+    if minikube status >/dev/null 2>&1; then
+      # Ignore transient states (Pending/ContainerCreating/Init) — batch CronJobs
+      # constantly spawn short-lived pods and would make the count flap.
+      pods=$(kubectl get pods --no-headers 2>/dev/null)
+      total=$(echo "$pods" | grep -c . || true)
+      failing=$(echo "$pods" | grep -cE "CrashLoopBackOff|ImagePull|ErrImage|Error|OOMKilled|Evicted" || true)
+      echo "minikube: running ($total pods, $failing failing)"
+      [ "$failing" -gt 0 ] && echo "$pods" | grep -E "CrashLoopBackOff|ImagePull|ErrImage|Error|OOMKilled|Evicted"
```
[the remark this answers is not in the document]
hvn2k1: Fixed in f53ffa286 — `kubectl get pods` is now guarded: if kubectl fails while `minikube status` still reports running, the status output prints an explicit "kubectl failed to list pods — check kubectl config" message instead of a misleading 0-pod count.

.claude/skills/devc/scripts/devc-exec.sh
```diff
@@ -0,0 +1,120 @@
...
+  # Case 2: local devcontainer (VS Code "Reopen in Container" / devcontainer CLI)
+  local repo_root
+  repo_root="$(git rev-parse --show-toplevel 2>/dev/null || pwd)"
+  CID="$(docker ps -q --filter "label=devcontainer.local_folder=$repo_root" 2>/dev/null | head -1 || true)"
+  if [ -z "$CID" ]; then
+    # Fallback: any running devcontainer whose local folder looks like a bucketeer checkout
+    CID="$(docker ps --filter "label=devcontainer.local_folder" \
+      --format '{{.ID}} {{.Label "devcontainer.local_folder"}}' 2>/dev/null \
+      | awk '$2 ~ /bucketeer/ {print $1; exit}' || true)"
+  fi
```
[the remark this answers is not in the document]
hvn2k1: Fixed in f53ffa286 — the fuzzy fallback is removed. The repo root is now derived from the script's own location (`<repo>/.claude/skills/devc/scripts/`), so the exact `devcontainer.local_folder` label match works from any cwd and detection now strictly follows exact match → Codespace → not-found. Multiple running checkouts can no longer be confused. Verified against a live devcontainer, including invocation from outside the repo.

.claude/skills/devcontainer-run/SKILL.md
```diff
@@ -0,0 +1,75 @@
+---
+name: devc
+description: >-
```
t-kikuc, line 3: The skill name `devc` does not convey its responsibility. The actual function is "detect the dev container and execute commands inside it" — the name should reflect that (e.g. `container-exec`, `devcontainer-run`).

The current name reads as an abbreviation of "dev container" but says nothing about *what it does* with the container.

.claude/skills/devcontainer-generate/SKILL.md
```diff
@@ -0,0 +1,62 @@
+---
```
t-kikuc, line 1: The core of `devc-generate` and `devc-deploy` is running existing make targets through the `devc` wrapper — `$DEVC 'make proto-all'` and `$DEVC 'make deploy-bucketeer'`. The verify steps and gotchas they document are useful, but the question is whether they justify separate skills.

Two options:
1. **Consolidate into the `devc` skill** — merge the generation/deploy workflows and verify steps into a single skill, reducing the number of skill files to maintain.
2. **Keep the current 3-skill split** — if the context-size savings from loading only the relevant skill are worth the maintenance cost of three files.

Which direction does the author prefer?

.claude/skills/devcontainer-run/SKILL.md
```diff
@@ -0,0 +1,75 @@
...
+- **Local devcontainer**: `/workspaces/bucketeer` is a bind mount of the host
+  repo. Files generated inside appear in the host working tree immediately.
+- **Codespace**: a separate clone. Generated or edited files stay in the
+  codespace. To get them back: commit and push from inside, or
+  `gh codespace cp 'remote:/workspaces/bucketeer/<path>' <local-path>`.
+  Always tell the user which mode you're in when file changes are involved
+  (`status` prints it).
+
+## Environment facts and gotchas
+
```
t-kikuc, line 65: **Applies to all three skill files (`devc`, `devc-generate`, `devc-deploy`):**

These skills contain operational knowledge (environment gotchas, connection details, verification steps, error diagnosis tips, etc.) that is useful for all developers, not just Claude. Currently `DEVELOPMENT.md` already has detailed minikube/docker-compose docs, but none of the skill files or `CLAUDE.md` reference it — and some of this knowledge exists only in the `.claude/` tree where human contributors won't find it.

Worth reviewing each item and deciding: is this human-facing knowledge that belongs in `DEVELOPMENT.md` (with skills referencing it), or is it Claude-specific instruction (verbose step-by-step, timeout hints, etc.) that should stay in `.claude/`? The two have different audiences — keeping the split intentional avoids both duplication and hidden knowledge.

.claude/skills/devcontainer-run/SKILL.md
hvn2k1: Agreed — renamed the whole family in 19dbe51a4: `devc` → `devcontainer-run`, `devc-generate` → `devcontainer-generate`, `devc-deploy` → `devcontainer-deploy` (wrapper script is now `.claude/skills/devcontainer-run/scripts/exec.sh`). Kept "devc" as a trigger alias in the skill description so the short form still works when invoking.

.claude/skills/devcontainer-generate/SKILL.md
hvn2k1: Author preference: keep the 3-skill split. Rationale: (1) skills load into context on demand, so a generation task only pays for the generation instructions, not the deploy runbook; (2) each workflow stays directly invocable as its own slash command (`/devcontainer-generate`, `/devcontainer-deploy`); (3) the actual shared logic — detection + exec — lives once in the wrapper script, so the two workflow skills are thin and the maintenance overlap is small. If they drift or grow stale we can consolidate later; merging is easier than splitting.

.claude/skills/devcontainer-run/SKILL.md
hvn2k1: Good point — done in 19dbe51a4. Moved the human-facing knowledge into `DEVELOPMENT.md`: a new "Working with the dev container from the host" section (kubectl-context hazard, protoc v23.4 requirement, gateway hosts resolving only inside the container), the `make start-minikube` exits-1-when-already-running note, and the faster single-service redeploy flow under "Deploy Bucketeer". Each skill now cross-references the relevant `DEVELOPMENT.md` section and keeps only the Claude-oriented operational detail (wrapper usage, timeouts, verification steps), with a note to keep the two in sync.

.claude/skills/devcontainer-deploy/SKILL.md
```diff
@@ -0,0 +1,88 @@
...
+The Bucketeer deployments are `api`, `web`, `batch-server`, and `subscriber`
+(confirm with `$DEVC 'kubectl get deployments'`). Chart-level
+changes instead: `helm upgrade bucketeer manifests/bucketeer/ --values
+manifests/bucketeer/values.dev.yaml`.
+
+## 4. Verify
+
+```bash
+$DEVC 'kubectl get pods'   # everything Running/Completed, restarts not climbing
+$DEVC 'curl -sk https://web-gateway.bucketeer.io/health'   # must run INSIDE the container
```
t-kikuc: `web-gateway.bucketeer.io/health` isn't a health endpoint here: the web chart's envoy has no `/health` route, so the catch-all `prefix: /` sends it to the `dashboard` cluster, which falls back to `index.html` for unknown paths (`pkg/web/cmd/server/web_console.go:50`). It returns 200 + HTML even with `api` / `batch-server` / `subscriber` all down.

`api-gateway` does route it (`manifests/bucketeer/charts/api/templates/envoy-configmap.yaml`):

```suggestion
$DEVC 'curl -sk https://api-gateway.bucketeer.io/health'   # must run INSIDE the container
```

hvn2k1: updated!

.claude/skills/devcontainer-run/scripts/exec.sh
```diff
@@ -0,0 +1,120 @@
...
+  echo "---"
+  run_inside '
+    echo "user: $(id -un)  workdir: $(pwd)"
+    echo "protoc: $(protoc --version 2>/dev/null || echo MISSING)"
+    command -v mockgen >/dev/null && echo "go-tools: OK" || echo "go-tools: MISSING (run bash .devcontainer/setup.sh)"
+    docker info >/dev/null 2>&1 && echo "dockerd: running" || echo "dockerd: NOT running (start with: nohup sudo dockerd > /tmp/dockerd.log 2>&1 &)"
+    if minikube status >/dev/null 2>&1; then
+      # Ignore transient states (Pending/ContainerCreating/Init) — batch CronJobs
+      # constantly spawn short-lived pods and would make the count flap.
+      if pods=$(kubectl get pods --no-headers 2>/dev/null); then
```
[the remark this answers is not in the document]
hvn2k1: Fixed in ab6feeb36. Every cluster command now names the context explicitly — `kubectl --context minikube` in the status pod query and in all the deploy/diagnose commands, `helm --kube-context minikube` for the chart upgrade — so the active context no longer decides which cluster gets touched.

`status` also reports the hazard rather than hiding it: it reads `kubectl config current-context` and, when that is not `minikube`, prints a `kube-context: WARNING ...` line naming the active context and reminding the caller to pass the context flags. The kubectl-failure branch now says the failure was in the minikube context specifically and points at `kubectl config get-contexts`.

Confirmed the context name is right: `tools/dev/Makefile:65` runs `kubectl config use-context minikube` after `minikube start` with the default profile, so `minikube` is the literal context name.

.claude/skills/devcontainer-deploy/SKILL.md
```diff
@@ -0,0 +1,88 @@
...
+
+```bash
+$DEVC 'make build-go-embed && TAG=localenv make build-docker-images && TAG=localenv make minikube-load-images'
+$DEVC 'kubectl rollout restart deployment <name> && kubectl rollout status deployment <name>'
+```
+
+The Bucketeer deployments are `api`, `web`, `batch-server`, and `subscriber`
+(confirm with `$DEVC 'kubectl get deployments'`). Chart-level
+changes instead: `helm upgrade bucketeer manifests/bucketeer/ --values
+manifests/bucketeer/values.dev.yaml`.
```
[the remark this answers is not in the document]
hvn2k1: Fixed in ab6feeb36. The chart-level `helm upgrade` was the one command left inline in prose; it is now a proper fenced block routed through the wrapper and pinned to the context:

```bash
$DEVC 'helm upgrade bucketeer manifests/bucketeer/ --kube-context minikube --values manifests/bucketeer/values.dev.yaml'
```

I also added an explicit note under it that both `$DEVC` and the context flag are mandatory, because a bare `helm`/`kubectl` on the host often points at a real cluster. Grepped the three skill files afterwards — no command outside `$DEVC` remains (the only bare `helm`/`kubectl`/`protoc` hits left are prose mentions).

.claude/skills/devcontainer-generate/SKILL.md
```diff
@@ -0,0 +1,63 @@
...
+
+## 3. Verify before declaring success
+
+- `git status --porcelain` (on the host for a local devcontainer; inside via
+  `$DEVC 'git status --porcelain'` for a codespace) — the changed files should
+  be only the ones related to your proto/interface change plus their generated
+  outputs. **A diff touching every `.pb.go` in the repo means a wrong protoc
+  version — abort and check `$DEVC 'protoc --version'`.**
+- Spot-check one regenerated file's header still says `protoc v4.23.4`:
+  `grep -m1 "protoc " proto/<domain>/<file>.pb.go`
```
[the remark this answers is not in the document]
hvn2k1: Fixed in ab6feeb36. The spot-check now runs inside the container:

```
$DEVC 'grep -m1 "protoc " proto/<domain>/<file>.pb.go'
```

with the reason stated inline — in codespace mode the host clone is a different checkout, so a bare `grep` would validate a file that was never regenerated. Good catch: this was the one verification step that could have reported success against stale output.

.claude/skills/devcontainer-run/scripts/exec.sh
```diff
@@ -0,0 +1,120 @@
...
+  CID="$(docker ps -q --filter "label=devcontainer.local_folder=$repo_root" 2>/dev/null | head -1 || true)"
+  if [ -n "$CID" ]; then
+    MODE=local
+    return
+  fi
+
+  # Case 3: GitHub Codespace (requires `gh` with the codespace scope)
+  CODESPACE="$(gh codespace list --json name,repository,state \
+    -q '.[] | select((.repository | endswith("/bucketeer")) and .state == "Available") | .name' 2>/dev/null \
+    | head -1 || true)"
```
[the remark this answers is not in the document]
hvn2k1: Fixed in ab6feeb36. `head -1` is gone. Detection now collects all available `*/bucketeer` codespaces and branches on the count:

- exactly one → use it
- more than one → print the list and `exit 2` with `Multiple available Bucketeer codespaces found — refusing to guess`, plus the `export BUCKETEER_CODESPACE=<name>` instruction
- `BUCKETEER_CODESPACE` set → validated against the candidate list with `grep -qx` and rejected with the available names if it does not match

So a fork also named `bucketeer` can no longer be silently selected. Documented the override and the new ambiguous-exit-2 case in the `devcontainer-run` skill, and `bash -n` passes.

.claude/skills/devcontainer-run/SKILL.md
```diff
@@ -0,0 +1,80 @@
+---
+name: devcontainer-run
```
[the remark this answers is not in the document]
hvn2k1: Fixed by updating the PR description — the names in the body now match what the change actually exposes (`devcontainer-run`, `devcontainer-generate`, `devcontainer-deploy`, and `devcontainer-run/scripts/exec.sh`).

The rename happened in 19dbe51a4 in response to an earlier review comment, and I left the description stale. Thanks for catching it. Note that `devc` still works as a trigger alias in the `devcontainer-run` skill description, so the short form remains usable when invoking.

t-kikuc (approved):
let's truy

PR #2751, opened 2026-08-06
feat(notification): implement GetNotificationUnreadCount API

hvn2k1:
Follows #2750

## What this PR does

Implements the `GetNotificationUnreadCount` RPC: returns the requesting user's unread notification count, backing the bell badge in the console.

This is the last read-side API of the notification center besides mark-all-as-read. The badge count must always agree with the Unread tab, so it uses the identical definition: published, not soft-deleted, no read marker for the viewer, and only notifications published after the viewer's account was created.

## Points

- The count is a single `COUNT` query with a correlated `NOT EXISTS` on `notification_read` — no fan-out rows, O(1) with respect to user count, per the RFC's read-time derivation design.
- The account-created bound reuses the same `earliestAccountCreatedAt` lookup as ListNotifications' UNREAD filter, guaranteeing badge/tab consistency.
- This branch is stacked on #2750 (mark-as-read), which is approved but blocked from merging by the `notify-success-to-slack` required-check misconfiguration in the main ruleset; the first commits shown here belong to that PR and will disappear once it merges.
- `MarkAllNotificationsAsRead` is the only remaining stub.


Ubisoft-potato (approved):
Thanks! LGTM!

PR #2760, opened 2026-08-07
feat: add variation value schema UI to the variations tab

cre8ivejp:
Implements the console UI for variation value schema validation from #2499.

The backend (model/persistence #2578, validation #2683) is already merged; this PR adds the
missing frontend so users can define and manage schemas from the console.

Screenshot 2026-08-07 at 10 55 58 PM

Screenshot 2026-08-07 at 10 55 34 PM

## Changes

- Add a "Value Schema" card at the top of the flag Variations tab showing the schema type,
  description, and a preview (enum chips, regex pattern, or read-only JSON Schema editor)
- Add an info tooltip on the card title explaining which validators each flag type supports
- Add an add/edit schema dialog with type-specific editors:
  - Enum: creatable multi-value select (STRING and NUMBER flags)
  - Regex: pattern input with live compile check (STRING flags)
  - JSON Schema: Monaco editor validated as draft 2020-12 (JSON flags)
- Add a "Test schema against current variations" action in the dialog that shows pass/fail
  per variation before saving
- Validate variation values client-side against the active schema (mirroring the backend
  matrix in `pkg/feature/domain/variation_value_schema.go`), blocking save with inline
  errors when values violate the schema
- Render variation value inputs as a dropdown of allowed values when an enum schema is active
- Save schema changes through the existing "Review and save" flow via `PATCH /v1/feature`
  (`variationValueSchema` / `clearVariationValueSchema`); scheduling is disabled when the
  schema changed since scheduled changes do not support schema updates
- Hide the card for BOOLEAN and YAML flags, which have no supported validators in v1
- Add en/ja i18n keys

## Dependency

- Adds `ajv` for client-side JSON Schema (draft 2020-12) validation

## Related Issue

Refs #2499

## Related PRs

- RFC: #2577
- Model and persistence: #2578
- Backend validation: #2683

## Tests

- `tsc --noEmit` and `eslint --max-warnings=0` pass
- Verified manually in the browser (en and ja): enum, regex, and JSON Schema flows —
  add/edit/remove, test-against-variations, persistence after reload, inline value
  validation blocking save, the scheduling guard, and the info tooltip


pradhankukiran (commented):
I tested the latest head locally against the dev stack. I found three client-side validation gaps that can produce behavior different from the backend or less detail than the issue expects.

ui/dashboard/src/utils/variation-value-schema.ts
```diff
@@ -0,0 +1,136 @@
...
+// this runs on every keystroke while editing a JSON Schema.
+const ajv2020 = new Ajv2020({ strict: false, allErrors: true });
+
+const compileJsonSchema = (schema: string) =>
+  ajv2020.compile(JSON.parse(schema));
+
+// Mirrors the backend's strconv.ParseFloat, which rejects surrounding
+// whitespace and non-finite values.
+const isStrictFiniteNumber = (value: string): boolean =>
+  value !== '' && value === value.trim() && Number.isFinite(Number(value));
```
pradhankukiran: `Number()` accepts numeric syntaxes that `strconv.ParseFloat` rejects. I verified that `0x10` and `0b10` are accepted as NUMBER enum values by the schema dialog, but the backend rejects the update. Could this use a decimal-only check, or otherwise mirror the backend parser, so the UI does not accept values that cannot be saved?

ui/dashboard/src/utils/variation-value-schema.ts
```diff
@@ -0,0 +1,136 @@
...
+      if (variationType === 'NUMBER') {
+        const allowed = values.map(Number);
+        return value =>
+          isStrictFiniteNumber(value) && allowed.includes(Number(value));
+      }
+      return value => values.includes(value);
+    }
+    case 'REGEX': {
+      try {
+        const pattern = new RegExp(schema.regexValidator?.pattern ?? '');
```
pradhankukiran: The client and backend still disagree because this uses JavaScript `RegExp` while the backend uses Go RE2. I verified that `(?=a)` can be tested successfully here but is rejected when saved, while a backend-valid pattern such as `(?i)^a+$` disables the Test action. Could compilation/testing use a backend-compatible RE2 implementation or an API-backed validation result?

ui/dashboard/src/utils/variation-value-schema.ts
```diff
@@ -0,0 +1,136 @@
...
+      }
+    }
+    case 'JSON_SCHEMA': {
+      try {
+        const validate = compileJsonSchema(
+          schema.jsonSchemaValidator?.schema ?? ''
+        );
+        return value => {
+          try {
+            return validate(JSON.parse(value)) === true;
```
pradhankukiran: This reduces AJV's result to a boolean and discards `validate.errors`. I verified that a value missing a required `name` field only shows `This value doesn't match the flag's value schema`, without the field or failed rule. Since #2499 asks the UI to identify relevant JSON fields when possible, could the validator preserve AJV's `instancePath` and message for the inline error?

ui/dashboard/src/utils/variation-value-schema.ts
cre8ivejp: I have replaced the Number()-based check with a decimal-only pattern (optional sign, digits, decimal point, exponent) plus a finiteness check, mirroring what the backend's strconv.ParseFloat accepts in practice. Hex, binary, Infinity, and padded values are now rejected in the dialog before save, while normal forms like 1e3, .5, and -1.5 still pass.

Fix: https://github.com/bucketeer-io/bucketeer/commit/3c29466db1be1db95dfde72527d980d8d51cb3a2

ui/dashboard/src/utils/variation-value-schema.ts
cre8ivejp: I added the re2js dependency (a JS port of RE2, the engine Go's regexp uses) and now compile and test patterns with it. Both directions of the mismatch are fixed, and I verified them in the browser: (?=a) is rejected in the dialog with a new "The pattern is not a valid RE2 regular expression" message (localized in en/ja), and (?i)^…$ compiles and the Test button works, with matching done unanchored like Go's MatchString.

Let me know what you think.

Fix: https://github.com/bucketeer-io/bucketeer/commit/3c29466db1be1db95dfde72527d980d8d51cb3a2

ui/dashboard/src/utils/variation-value-schema.ts
cre8ivejp: @pradhankukiran, thank you so much for catching many bugs!

I fixed it so the value validator now returns the first AJV error's instancePath and message. It shows up in two places, both verified against the dev stack:

Screenshot 2026-08-08 at 2 53 02 AM

And inline on the variation value field: "This value doesn't match the flag's value schema (/ must have required property 'name')", with save still blocked.

Two things worth knowing are that the AJV detail is appended verbatim to the localized message, so that fragment is English-only (AJV doesn't produce Japanese messages). I'd consider that acceptable for a technical detail, and re2js is a new frontend dependency (~pure JS, no WASM). Typecheck and lint pass, and none of the browser testing persisted changes to the dev backend.

Let me know what you think.

Fix: https://github.com/bucketeer-io/bucketeer/commit/3c29466db1be1db95dfde72527d980d8d51cb3a2

ui/dashboard/src/utils/variation-value-schema.ts
pradhankukiran: Retested commit `3c29466d` in the rebuilt dev stack. `(?=a)` is now rejected with the RE2 validation message and Save/Test disabled. `(?i)^a+$` compiles, enables Test, and correctly marks `aaa` valid and `bbb` invalid. This fixes both directions of the mismatch. Thanks!

ui/dashboard/src/utils/variation-value-schema.ts
pradhankukiran: Retested commit `3c29466d` in the rebuilt dev stack. `0x10` and `0b10` now show "All allowed values must be numbers", and both Save and Test remain disabled. Normal decimal values still work. This fixes the issue for me. Thanks!

ui/dashboard/src/utils/variation-value-schema.ts
pradhankukiran: Retested commit `3c29466d` in the rebuilt dev stack. Required and type failures now include useful details, for example `/ must have required property 'name'` and `/maxItems must be integer`. A few details are still lost: `{"name":"a","extraField":"x"}` shows `/ must NOT have additional properties` without naming `extraField`; an invalid enum shows `/theme must be equal to one of the allowed values` without listing them; and a value with multiple violations shows only the first error. AJV exposes these through `error.params` and the full `validate.errors` array. Since #2499 asks for relevant fields when possible and demonstrates multiple errors, could we retain those details too? I am leaving this thread open.

cre8ivejp: @pradhankukiran, I have pushed it. Can you check it again?
https://github.com/bucketeer-io/bucketeer/commit/4c28a6ae435bd1de5cbc0dca5318ed7632e6be3e

Test_results_showing_multiple_violations_with_prop

pradhankukiran: Thanks, I checked the latest commit. The extra property name and allowed enum values now show up, and multiple violations are included too. This addresses the remaining issue I found.

cre8ivejp: Thanks for confirming! Is that ok to merge?

pradhankukiran: yes, looks good to me.

cre8ivejp: Thank you so much for your contribution!

PR #2767, opened 2026-08-10
fix: obfuscate the API key in audit log

### LINKED ISSUE #2766: fix: obfuscate the API key in audit log
We have obfuscate the API key in the console, however the Audit log still showing the key, actions:
- Obfuscate the API key when create audit log
- Obfuscate the API key when reading audit log

hvn2k1:
Fixes #2766

## What this PR does

Obfuscates the API key both when the audit log is created and when it is read, so the raw key is no longer stored in domain events / audit logs, nor returned by the audit log APIs.

The Get/List API key endpoints already obfuscate the key, but `CreateAPIKey` and `UpdateAPIKey` publish the API key entity as-is. The raw key therefore ends up in `audit_log.entity_data`, `previous_entity_data` and the `APIKeyCreatedEvent`, and is shown in the audit log detail on the console.

## Points

- `obfuscateAPIKey` moved from `pkg/account/api` to `pkg/account/domain` as `ObfuscateAPIKey`, so the account and auditlog services share one function. Behavior is unchanged (first 4 + `....` + last 4).
- `CreateAPIKey` still returns the raw key in its response, which is the only time its owner can read it. The domain event is built from an obfuscated copy, so the response is untouched.
- The read side masks `entity_type == APIKEY` logs in `GetAuditLog`, `ListAuditLogs`, `ListAdminAuditLogs` and `ListFeatureHistory`. It is needed for the rows saved before this change. The public API gateway endpoints proxy to this service, so they are covered too.
- When the entity data or the event data cannot be parsed, it is dropped instead of being returned as-is, so a raw key never leaks through the fallback path.
- Not included: a backfill for the rows already saved. They still hold the raw key at rest, and `search_keyword` can still match it in the DB.
- The entity data fields of API key logs come back alphabetically sorted, because the JSON is decoded into a map and re-encoded. The console parses the JSON before rendering, so the diff view is unaffected.

### DIFF
--- a/pkg/account/api/api_key.go
+++ b/pkg/account/api/api_key.go
@@ -20,7 +20,6 @@ import (
 	"strconv"
 
 	"go.uber.org/zap"
-	pb "google.golang.org/protobuf/proto"
 
 	"github.com/bucketeer-io/bucketeer/v2/pkg/account/domain"
 	v2as "github.com/bucketeer-io/bucketeer/v2/pkg/account/storage/v2"
@@ -30,16 +29,6 @@ import (
 	proto "github.com/bucketeer-io/bucketeer/v2/proto/account"
 	eventproto "github.com/bucketeer-io/bucketeer/v2/proto/event/domain"
 )
-
-// obfuscatedAPIKeyEntity returns a copy of the API key with its key obfuscated.
-func obfuscatedAPIKeyEntity(k *proto.APIKey) *proto.APIKey {
-	if k == nil {
-		return nil
-	}
-	obfuscated := pb.Clone(k).(*proto.APIKey)
-	obfuscated.ApiKey = domain.ObfuscateAPIKey(obfuscated.ApiKey)
-	return obfuscated
-}
 
 func (s *AccountService) CreateAPIKey(
 	ctx context.Context,
@@ -96,8 +85,6 @@ func (s *AccountService) CreateAPIKey(
 		return nil, api.NewGRPCStatus(err).Err()
 	}
 
-	// The raw key is returned to its creator only.
-	obfuscatedKey := obfuscatedAPIKeyEntity(key.APIKey)
 	e, err := domainevent.NewEvent(
 		editor,
 		eventproto.Event_APIKEY,
@@ -111,10 +98,12 @@ func (s *AccountService) CreateAPIKey(
 			CreatedAt:  key.CreatedAt,
 			UpdatedAt:  key.UpdatedAt,
 			Maintainer: key.Maintainer,
-			ApiKey:     obfuscatedKey.ApiKey,
+			ApiKey:     key.ApiKey,
 		},
 		req.EnvironmentId,
-		obfuscatedKey,
+		// The domain event keeps the raw key: it is where the api key cache pipeline resolves the
+		// cache key from. It is obfuscated when the audit log is created.
+		key.APIKey,
 		nil,
 	)
 	if err != nil {
@@ -370,8 +359,8 @@ func (s *AccountService) UpdateAPIKey(
 			Id: req.Id,
 		},
 		req.EnvironmentId,
-		obfuscatedAPIKeyEntity(current),
-		obfuscatedAPIKeyEntity(prev),
+		current,
+		prev,
 	)
 	if err != nil {
 		return nil, err
--- a/pkg/account/api/api_key.go
+++ b/pkg/account/api/api_key.go
@@ -20,6 +20,7 @@ import (
 	"strconv"
 
 	"go.uber.org/zap"
+	pb "google.golang.org/protobuf/proto"
 
 	"github.com/bucketeer-io/bucketeer/v2/pkg/account/domain"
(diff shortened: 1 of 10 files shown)


pkg/account/domain/api_key.go
```diff
@@ -27,7 +27,19 @@ import (
...
 )
 
-const keyBytes = 32
+const (
+	keyBytes           = 32
+	apiKeyVisibleChars = 4
+)
+
+// ObfuscateAPIKey shows the first and last apiKeyVisibleChars characters, with dots in the middle.
+// It must be used everywhere an API key is returned, stored or logged, except when it is created.
```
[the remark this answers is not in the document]
hvn2k1: this is intended, we don't log the API key

pkg/account/api/api_key.go
```diff
@@ -372,8 +370,8 @@ func (s *AccountService) UpdateAPIKey(
 			Id: req.Id,
 		},
 		req.EnvironmentId,
-		current,
-		prev,
+		obfuscatedAPIKeyEntity(current),
+		obfuscatedAPIKeyEntity(prev),
```
t-kikuc: This turns the API key cache invalidation into a no-op.

`ExtractAPIKeySecrets` pulls the raw secret out of `entity_data`, and it is the only source for both `cacheRefresher.refreshAPIKey` and `cacheInvalidator`, which evict by the raw secret because that is the cache key. They now get `0123....cdef`: `GetEnvironmentAPIKey` returns `ErrAPIKeyNotFound`, `refreshAPIKey` reads that as "rotated away" and evicts a nonexistent key, so nothing is logged and the real entries survive.

Effect: disabling a key takes up to a minute instead of applying immediately (L1 TTL 1m, batch `apiKeyCacher` every minute). Bounded, but silent and untested.

Could the invalidation resolve the secret from `event.EntityId` instead?

pkg/auditlog/api/obfuscate_api_key.go
```diff
@@ -0,0 +1,91 @@
...
+	"google.golang.org/protobuf/types/known/anypb"
+
+	accountdomain "github.com/bucketeer-io/bucketeer/v2/pkg/account/domain"
+	proto "github.com/bucketeer-io/bucketeer/v2/proto/auditlog"
+	eventproto "github.com/bucketeer-io/bucketeer/v2/proto/event/domain"
+)
+
+const apiKeyJSONField = "api_key"
+
+// obfuscateAPIKeys obfuscates the API keys saved before they were obfuscated at the source.
```
t-kikuc: Asking rather than requesting: would mask-at-creation plus a backfill be better than masking on read?

The write-side change already keeps the key out of new rows, so this file (91 lines + 188 of tests) exists only for the older ones -- a backfill would cover those instead.

It also looks hard to avoid: `AuditLogStorage` has no delete method and there is no retention job, and `UpdateAPIKeyRequest` cannot regenerate a key, so anything left in `audit_log` stays a live credential.

Masking on read also has to cover every read path as it is added -- two of the four call sites here are already unreachable (`ListFeatureHistory` pins `EntityType` to `Event_FEATURE`).

What do you think?

pkg/account/api/api_key.go
hvn2k1: Confirmed, thanks — this was a real break, not just a slow path. `cacheInvalidator` has no storage at all (caches + dispatcher only), so resolving the secret from `EntityId` works in the refresher but not in the invalidator.

Fixed by taking the obfuscation off the domain event entirely and doing it in `NewAuditLog` instead: the event keeps the raw key, so `ExtractAPIKeySecrets` and both consumers behave exactly as before, and the key is obfuscated before the audit log row is inserted.

`TestCreateAPIKeyDomainEvent` / `TestUpdateAPIKeyDomainEvent` now assert `ExtractAPIKeySecrets(event)` returns the raw secret, so this cannot regress silently.

pkg/auditlog/api/obfuscate_api_key.go
hvn2k1: Agreed on mask-at-creation, and I moved it there: `NewAuditLog` now obfuscates every audit log row, and it is the single funnel (the persister plus the two direct writers in `account.go`), so a new writer cannot skip it. The read path is left only for the rows written before this change, and both paths now share one implementation in `pkg/auditlog/domain`.

Also dropped the `ListFeatureHistory` call site you flagged. I kept the `ListAdminAuditLogs` one: api key events are never admin events today, so it is unreachable as well, but `admin_audit_log` takes generic audit logs and it is a single line.

On the backfill: we decided not to do it in this PR, so the older rows keep their raw key in the DB and the read path is what protects them. Happy to file a follow-up if you would rather have them scrubbed.

pkg/account/api/api_key.go
hvn2k1: Pushed as 12b4ef672.

pkg/auditlog/api/obfuscate_api_key.go
hvn2k1: Pushed as 12b4ef672.

t-kikuc (approved):
thank you!

PR #2097, opened 2025-08-29
deps: upgrade go version to go1.25

nnnkkk7:
Upgrade go version to 1.25.
Execute `go mod edit -go=1.25`.

PR #2782, opened 2026-08-24
deps: upgrade go version to go1.26

### LINKED ISSUE #2474: deps: upgrade go version to Go1.26
Since Go1.26 has been released, we can upgrade it and should fix lint errors as it.
https://go.dev/doc/go1.26

kimurakazuhiro-c:
Follows #2097
Fixes #2474

## What this PR does

Bumps the Go toolchain from 1.25 to 1.26.7 in every place the version is pinned: the `go` directive in all six modules (root plus the five `hack` modules), the devcontainer's Go tarball, and the `golang` builder base images used by the `hack` Dockerfiles.

Keeps the project on a current Go release, following the same steps as the 1.25 upgrade in #2097. No CI changes are needed because every workflow resolves the version through `go-version-file`.

The devcontainer had also drifted from `go.mod`: it installed go1.25.1 while `go.mod` required 1.25.8, so under the default `GOTOOLCHAIN=auto` the toolchain baked into the image was never the one actually used — Go downloaded 1.25.8 on demand instead. Both are now 1.26.7.

## Points

- `go mod tidy` on the `hack` modules also refreshed their indirect dependencies (`golang.org/x/{crypto,net,text}`, `google.golang.org/api`, `cloud.google.com/go/auth`, `enterprise-certificate-proxy`). The root module's dependencies are untouched — only its `go` directive moved.
- The pin granularity of the `hack` Dockerfiles is intentionally left as it was. `create-e2e-accounts` stays patch-pinned (`golang:1.26.7`) because CI builds and publishes that image on every change under its directory, while the other three stay minor-pinned (`golang:1.26`) as they are only built by hand from their local Makefiles. Aligning them is out of scope here.
- The proto and mockgen checks are left to CI, since they need protoc v23.4 and clang-format locally. This PR touches no `.proto` files and no Go interfaces, so no generated files should change.


kimurakazuhiro-c:
I am addressing errors that occurred in the following file when running the `go vet` command:
pkg/api/api/save_api_key_last_used_at_test.go

kimurakazuhiro-c:
To check for any missed corrections, I performed a grep search on the pre-update version (1.25) and found no missed corrections.
[grep-1.25-result.txt](https://github.com/user-attachments/files/31398455/grep-1.25-result.txt)

kimurakazuhiro-c:
I reviewed the release notes and performed a grep search on libraries with specification changes to verify their usage.
https://go.dev/doc/go1.26
[grep-Minor.changes.to.the.library-result.txt](https://github.com/user-attachments/files/31398493/grep-Minor.changes.to.the.library-result.txt)



I assessed the potential impact on each library in use and confirmed that there are no adverse effects.

・net/url: No impact
→ `url.Parse` appears only once (at `middleware_test.go:152`). Zero instances of colons or IPv6 literals in the host component.
・crypto/tls: No impact
→ The change involves default behavior (enabling PQ key exchange). Although it applies because `CurvePreferences` is unset,
`X25519MLKEM768` has been the default since Go 1.24, so there is no new risk. Issues can be avoided using `tlssecpmlkem=0`.
・net/http: No impact
→ The change regarding trailing slashes (301 → 307) affects only static delivery paths (`/assets/`, `/static/js/`). 
Other muxes use exact-match path registration; `gRPC-Gateway`'s `runtime.ServeMux` is unaffected. 
Regarding Cookie scope changes: zero instances of `http.Cookie` or `CookieJar` usage.
・crypto/rsa: No impact
→ The change ignores the random io.Reader parameter, so what matters is whether we pass one. We never do: zero direct calls to rsa.Sign*/Decrypt*/GenerateKey, and the only signing path (go-jose at signer.go:58) already uses crypto/rand.Reader. 
・crypto/rand: No impact
→ Used only as a randomness source (rand.Read, io.ReadFull(rand.Reader, ...)), never injected into a crypto function. crypto/{dsa,ecdh,ecdsa,ed25519} are not imported.
・os/signal: No impact
→ No calls to functions that underwent changes.
・net/http/httptest: No impact
→ No calls to functions that underwent changes.
・testing: No impact
→ No calls to functions that underwent changes.
・os: No impact
→ No calls to functions that underwent changes.
・fmt: No impact
→ Zero allocation verification tests; zero instances of error type determination using `%T`.
・time: No impact
→ The changes relate to a future release, so they are not addressed in this update.

t-kikuc (approved):

PR #2789, opened 2026-08-26
feat(migration): add organization_id column to audit_log

hvn2k1:
Part of #1982

## What this PR does

Adds an `organization_id` column (default `''`) and an `(organization_id, timestamp DESC)` index to the `audit_log` table, for both MySQL and Postgres.

Organization-level events (account changes, environment/project creation, organization updates) currently go to `admin_audit_log`, which only system admins can read, so organization admins cannot see their own organization's activity. The plan is to store these events in `audit_log` with an empty `environment_id` and a populated `organization_id`, and serve them from the existing audit log console with role-scoped filters. This PR is step 1 (schema only); the write path, history migration, and read path follow in separate PRs.

## Points

- MySQL uses `ALGORITHM=INSTANT` so the column add doesn't rebuild the table.
- Postgres uses `CREATE INDEX CONCURRENTLY`, so the migration file is marked `-- atlas:txmode none` (`CONCURRENTLY` cannot run inside a transaction).
- Index names follow each side's existing convention: `idx_organization_id_timestamp_desc` (MySQL, matching `idx_environment_id_timestamp_desc`) and `idx_audit_log_organization_timestamp` (Postgres, matching `idx_audit_log_environment_timestamp`).
- Both statements were verified against MySQL 8.0 and Postgres 18 locally; `atlas migrate validate` passes for both directories.


t-kikuc (approved):
LGTM

PR #2791, opened 2026-08-27
feat(auditlog): write organization events to audit_log

hvn2k1:
Part of #1982, Follows #2789

## What this PR does

Adds `organization_id` to the domain event and stores organization-level (admin) events in the `audit_log` table — with an empty `environment_id`, scoped by their organization id — instead of `admin_audit_log`.

Organization-level events (account changes, environment/project creation, organization updates) go to `admin_audit_log`, which only system admins can read, so organization admins cannot see their own organization's activity. With the `organization_id` column in place (#2789), this switches the write path so new organization events land in `audit_log` where they can be served per-organization. After this PR the only writes to `admin_audit_log` are a rollout-compatibility fallback for organization-scoped events published by older producers (no `organization_id` yet), so the history migration that follows can still resolve them; once old producers and queued events are drained, the fallback goes away and the table is frozen. The read path and console filters come in a separate PR.

## Points

- Every `NewAdminEvent` producer now passes its entity's organization id (organization, project, environment, account, team). Admin subscription events are system-level and pass an empty organization id, so they land in `audit_log` with both `environment_id` and `organization_id` empty — visible to system admins only.
- Rollout compatibility: an organization-scoped admin event arriving without an organization id (published by an older producer) falls back to `admin_audit_log` instead of being stranded in `audit_log` with both scopes empty; a TODO marks the fallback for removal once old messages drain.
- `Event.IsAdminEvent` is kept and its notification semantics are untouched (the domain event informer still uses it); only the storage routing changed.
- The account service's synchronous in-transaction audit writes were switched from `admin_audit_log` to `audit_log`. The dedup between that write and the published event is unchanged: same event id, duplicate-key insert → ack.
- `AdminAuditLogStorage` is removed from the subscriber persister and wiring, but the storage itself and the read APIs stay until the read-path switch and cleanup.
- New organization events are not visible in any console until the read-path PR — expected, since nothing reads `admin_audit_log` in production.


proto/event/domain/event.proto
```diff
@@ -287,6 +287,7 @@ message Event {
   string entity_data = 11;           // JSON string of the entity data
   string previous_entity_data = 12;  // JSON string of the previous entity data
   string environment_id = 13;
+  string organization_id = 14;
```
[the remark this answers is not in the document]
hvn2k1: That's the purpose of this pr, to write to another table

pkg/subscriber/processor/auditlog_persister.go
```diff
@@ -115,21 +111,18 @@ func (a auditLogPersister) Process(
...
 func (a auditLogPersister) flushChunk(chunk map[string]*puller.Message) {
-	auditlogs, adminAuditLogs, messages, adminMessages := a.extractAuditLogs(chunk)
+	auditlogs, messages := a.extractAuditLogs(chunk)
 	ctx, cancel := context.WithTimeout(
 		context.Background(),
 		time.Duration(a.auditLogPersisterConfig.FlushTimeout)*time.Second,
 	)
 	defer cancel()
-	// Environment audit logs
 	a.createAuditLogs(ctx, auditlogs, messages, a.storage.CreateAuditLog)
```
Ubisoft-potato, line 123: After this PR, `CreateAdminAuditLog` / `CreateAdminAuditLogs` have no production callers (`CreateAdminAuditLogs` was already unused), and nothing matches on `ErrAdminAuditLogAlreadyExists` anymore. Not blocking — `AdminAuditLogStorage` stays for `ListAdminAuditLogs`. 

Suggest removing the write methods + sentinel in the read-path PR.

hvn2k1: We will remove this in another pr. We don't have admin audit log in production anyway

pkg/subscriber/processor/auditlog_persister.go
```diff
@@ -138,16 +131,10 @@ func (a auditLogPersister) extractAuditLogs(
...
-				storage.AdminEnvironmentID,
-			))
-			adminMessages = append(adminMessages, msg)
-		} else {
-			auditlogs = append(auditlogs, domain.NewAuditLog(event, event.EnvironmentId))
-			messages = append(messages, msg)
-		}
+		// Organization-level (admin) events are stored in the audit_log table
+		// with an empty environment id, scoped by their organization id.
+		auditlogs = append(auditlogs, domain.NewAuditLog(event, event.EnvironmentId))
```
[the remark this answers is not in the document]
hvn2k1: https://github.com/bucketeer-io/bucketeer/pull/2791/changes/0bdd0aee56a5424f882192713d08bb69242bfeb3

pkg/subscriber/processor/auditlog_persister_test.go
```diff
@@ -57,32 +57,72 @@ func TestExtractAuditLogs(t *testing.T) {
...
 	}
-	for i, al := range adminAuditLogs {
-		msg, ok := chunk[al.Id]
-		assert.True(t, ok)
-		assert.Equal(t, msg.ID, al.Id)
-		assert.Equal(t, adminMessages[i].ID, al.Id)
-	}
+	// The legacy organization-scoped event goes to admin_audit_log so the
+	// history migration can resolve its organization later.
+	assert.Len(t, adminAuditLogs, 1)
```
t-kikuc, line 123: `assert.Len` does not stop the test, so if a routing regression sends the legacy event to `auditlogs` instead, `adminAuditLogs[0].Id` on the next line panics with an index-out-of-range and the package's other tests never report. Use `require.Len` here.

```suggestion
	require.Len(t, adminAuditLogs, 1)
```

t-kikuc (approved):
Nit: The bullet "AdminAuditLogStorage is removed from the subscriber persister and wiring" is stale — the rollout fallback (0bdd0aee) keeps it. Consider removing that bullet.

PR #2792, opened 2026-08-31
test(api): fix flaky GetEvaluations tests caused by CreatedAt

### LINKED ISSUE #1023: TestGrpcGetEvaluationsValidation is flaky
### Version

main

### Describe the bug
```
=== RUN   TestGrpcGetEvaluationsValidation/success
    api_grpc_test.go:1906: 
        	Error Trace:	/home/runner/work/bucketeer/bucketeer/pkg/gateway/api/api_grpc_test.go:1906
        	Error:      	Not equal: 
        	            	expected: &gateway.GetEvaluationsResponse{state:impl.MessageState{NoUnkeyedLiterals:pragma.NoUnkeyedLiterals{}, DoNotCompare:pragma.DoNotCompare{}, DoNotCopy:pragma.DoNotCopy{}, atomicMessageInfo:(*impl.MessageInfo)(nil)}, sizeCache:0, unknownFields:[]uint8(nil), State:2, Evaluations:(*feature.UserEvaluations)(0xc000023680), UserEvaluationsId:"no_evaluations"}
        	            	actual  : &gateway.GetEvaluationsResponse{state:impl.MessageState{NoUnkeyedLiterals:pragma.NoUnkeyedLiterals{}, DoNotCompare:pragma.DoNotCompare{}, DoNotCopy:pragma.DoNotCopy{}, atomicMessageInfo:(*impl.MessageInfo)(nil)}, sizeCache:0, unknownFields:[]uint8(nil), State:2, Evaluations:(*feature.UserEvaluations)(0xc0006cc600), UserEvaluationsId:"no_evaluations"}
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -28,3 +28,3 @@
        	            	   },
        	            	-  CreatedAt: (int64) 1718339983,
        	            	+  CreatedAt: (int64) 1718339984,
        	            	   ArchivedFeatureIds: ([]string) {
        	Test:       	TestGrpcGetEvaluationsValidation/success
        	Messages:   	success
```

_No response_

### Expected behavior

_No response_

### Relevant log output

_No response_

### Additional information

_No response_

kimurakazuhiro-c:
Fixes #1023

## What this PR does

Removes the flakiness in the `GetEvaluations` tests by asserting `UserEvaluations.CreatedAt` with a time tolerance instead of an exact equality check.

`UserEvaluations.CreatedAt` is stamped at request time inside the handler, so the hard-coded expectation and the actual value differ whenever the second boundary is crossed between building the expectation and running the request. The tests compared the whole response with `assert.Equal`, so this occasionally failed. `api_test.go` carried a `FIXME: This is a flaky test. CreateAt may not be equal ocasionally.` comment acknowledging the problem without fixing it; that comment is now removed.

## Points

- `normalizeUserEvaluationsCreatedAt` asserts the actual `CreatedAt` is within 5 seconds of `time.Now().Unix()`, then copies it into the expected value so the subsequent `assert.Equal` still verifies every other field. This keeps the field covered rather than ignoring it.
- The 5-second delta is a deliberately loose bound: these are in-process handler calls that finish in milliseconds, so the tolerance only needs to absorb clock-boundary skew and CI scheduling stalls, not real latency.
- Applied to all four affected tests: `TestGrpcGetEvaluationsValidation`, `TestGrpcGetEvaluationsZeroFeature` (gRPC) and `TestGetEvaluationsValidation`, `TestGetEvaluationsZeroFeature` (HTTP).
- The helper is a no-op when either side is nil, so the error-path cases that expect a nil response are unaffected.
- Test-only change: no production code is touched.
- #1023 reports the failure at `pkg/gateway/api/api_grpc_test.go`; that package has since moved to `pkg/api/api/`, and the failing test is the same `TestGrpcGetEvaluationsValidation/success` case fixed here.

### DIFF
--- a/pkg/api/api/api_grpc_test.go
+++ b/pkg/api/api/api_grpc_test.go
@@ -2360,7 +2360,7 @@ func TestGrpcGetEvaluationsValidation(t *testing.T) {
 			})
 			actual, err := gs.GetEvaluations(ctx, p.input)
 			if p.expected != nil && actual != nil {
-				normalizeUserEvaluationsCreatedAt(t, p.expected.Evaluations, actual.Evaluations, "%s", p.desc)
+				normalizeUserEvaluationsCreatedAt(t, p.expected.Evaluations, actual.Evaluations)
 			}
 			assert.Equal(t, p.expected, actual, "%s", p.desc)
 			assert.Equal(t, p.expectedErr, err, "%s", p.desc)
@@ -2413,7 +2413,7 @@ func TestGrpcGetEvaluationsZeroFeature(t *testing.T) {
 			"authorization": []string{"test-key"},
 		})
 		actual, err := gs.GetEvaluations(ctx, p.input)
-		normalizeUserEvaluationsCreatedAt(t, p.expected.Evaluations, actual.Evaluations, "%s", p.desc)
+		normalizeUserEvaluationsCreatedAt(t, p.expected.Evaluations, actual.Evaluations)
 		assert.Equal(t, p.expected, actual, "%s", p.desc)
 		assert.Equal(t, p.expected.State, actual.State, "%s", p.desc)
 		assert.Equal(t, p.expectedErr, err, "%s", p.desc)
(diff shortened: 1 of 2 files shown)


pkg/api/api/api_grpc_test.go
```diff
@@ -4959,6 +4963,20 @@ func emptyUserEvaluations(t *testing.T) *featureproto.UserEvaluations {
...
+	expected, actual *featureproto.UserEvaluations,
+	msgAndArgs ...interface{},
+) {
+	t.Helper()
+	if expected == nil || actual == nil {
+		return
+	}
+	assert.InDelta(t, time.Now().Unix(), actual.CreatedAt, 5, msgAndArgs...)
+	expected.CreatedAt = actual.CreatedAt
+}
```
t-kikuc: The comment describes *what* the function does (readable from the signature + body). Replace with *why* — the flaky CreatedAt timing that motivates this helper.

pkg/api/api/api_grpc_test.go
```diff
@@ -4959,6 +4963,20 @@ func emptyUserEvaluations(t *testing.T) *featureproto.UserEvaluations {
 	}
 }
 
+// normalizeUserEvaluationsCreatedAt asserts CreatedAt is roughly now, then aligns expected so Equal stays stable.
+func normalizeUserEvaluationsCreatedAt(
+	t *testing.T,
+	expected, actual *featureproto.UserEvaluations,
+	msgAndArgs ...interface{},
```
t-kikuc: `msgAndArgs` can be dropped — `t.Helper()` already attributes the failure to the caller line, and the call sites run inside `t.Run(p.desc)` which names the case in the output.

pkg/api/api/api_grpc_test.go
kimurakazuhiro-c: @t-kikuc 
I have updated the description to address the instability (flaky behavior) regarding the timing of `CreatedAt`.

pkg/api/api/api_grpc_test.go
kimurakazuhiro-c: @t-kikuc 
Modified to omit `msgAndArgs`.

t-kikuc (approved):

PR #2809, opened 2026-09-07
deps: bump grpc and fast-uri to fix HIGH CVEs

Ubisoft-potato:
Fixes five HIGH CVEs flagged by `pr-trivy`.

| Package | From | To | CVEs |
|---|---|---|---|
| `google.golang.org/grpc` | v1.82.1 | v1.83.2 | CVE-2026-84304 |
| `fast-uri` | 3.1.5 | 3.1.7 | CVE-2026-75899, CVE-2026-75931, CVE-2026-75975, CVE-2026-76172 |

These are already present on `main` — recently published CVEs against unchanged dependency versions. `pr-trivy` only runs on `pull_request`, so `main` is never re-scanned and the failure surfaces on unrelated PRs.

- **grpc**: the root `go.mod` is already on v1.83.2. The five `hack/*` modules are separate Go modules that had drifted behind, carrying grpc as an indirect dependency.
- **fast-uri**: transitive via `ajv@^8.20.0` with range `^3.0.1`, so only the lock entry moves. Fixed in 3.1.6; 3.1.7 is the latest in range.

Verified locally: all five `hack/*` modules build, `yarn install --frozen-lockfile` accepts the lockfile unchanged, and the CI trivy invocation reports 0 vulnerabilities across all 9 targets.


hvn2k1 (approved):
Thank you!

PR #2816, opened 2026-09-14
fix(api): re-evaluate rule-based flags on SSE reconnect

### LINKED ISSUE #2814: Question: does SSE reconnect fully re-evaluate rule-based flags after a user attribute update?
## Summary

While writing e2e tests for the JS SDK, I noticed a behavior around SSE reconnects that
I'd like to confirm here. When a user's attributes change and the client reconnects the
SSE stream, a flag with a targeting rule based on that attribute did not come back
updated in my test run, while the same attribute update over REST did return the
updated value. I'm sharing the details below so we can confirm together whether this is
intended or worth adjusting.

## Observed behavior

**Environment / tag:** e2e test environment for `bucketeer-io/javascript-client-sdk`,
feature tag `javascript`

**Feature flag under test:**
- ID: `feature-js-e2e-string`, version 5
- Default variation: `variation 1` (id `87e0a1ef-a0cb-49da-8460-289948f117ba`) → value
  `"value-1"`, reason `DEFAULT`
- Targeting rule: `app_version == "0.0.1"` → `variation 2` (id
  `802f2b29-a5c5-47d1-b5ba-f457d224c7b2`) → value `"value-2"`, reason `RULE`

**User:** id `bucketeer-js-user-id-1`, starts with no attributes (`{}`)

**e2e test:**
https://github.com/bucketeer-io/javascript-client-sdk/blob/b96545d8d48851902bc5ea4eca5b12bed706139d/e2e/streaming.spec.ts#L100-L144

### Step 1 — initial connect

`POST /v1/gateway/stream_evaluations`
```json
{"tag":"javascript","user":{"id":"bucketeer-js-user-id-1","data":{}},"sourceId":7,"sdkVersion":"2.6.1","userEvaluationsId":"","evaluatedAt":"0"}
```
The server responded with `put`, `forceUpdate: true`, including `feature-js-e2e-string`
= `value-1` (`DEFAULT`), as expected. The client cached `userEvaluationsId` (`UEID1`,
e.g. `186937244628824709`) and `evaluatedAt` (`T1`, e.g. `1789051042`).

### Step 2 — update attributes, SDK reconnects

`POST /v1/gateway/stream_evaluations` (fresh connection, same as any reconnect)
```json
{"tag":"javascript","user":{"id":"bucketeer-js-user-id-1","data":{"app_version":"0.0.1"}},"sourceId":7,"sdkVersion":"2.6.1","userEvaluationsId":"186937244628824709","evaluatedAt":"1789051042"}
```

**I expected** the response to include `feature-js-e2e-string` = `value-2`, reason
`RULE`.

**I observed** the captured response came back without `feature-js-e2e-string` at all:
```json
{"evaluations":{"id":"2618820213135455269","evaluations":[{"id":"feature-javascript-e2e-sse-string:13:bucketeer-js-user-id-1", "featureId":"feature-javascript-e2e-sse-string", ...}],"createdAt":"1789051042","archivedFeatureIds":[],"forceUpdate":false},"userEvaluationsId":"7419377868834335765"}
```
The one flag that is present, `feature-javascript-e2e-sse-string`, appears to be there
because its own definition happened to be touched by an unrelated concurrent test, not
because of the attribute change.

**For comparison**, I ran the same attribute update via REST (`fetchEvaluations()`
after `updateUserAttributes()`), which correctly returned `value-2` / `RULE` for the
same flag and user, using `getEvaluations`.

## Question

Should a flag with a targeting rule come back updated on an SSE reconnect after a user
attribute change like this, the same way it does over REST? Or is there a reason the
SSE path handles this differently? Keen to hear how the backend side of this works
before looking any further.

hvn2k1:
Fixes #2814

## What this PR does

Makes the `stream_evaluations` (SSE) endpoint re-evaluate flags with targeting rules when a client reconnects after a user attribute change, matching the behavior of `GetEvaluations`.

## Background

The SSE diff path called `EvaluateFeaturesByEvaluatedAt` with `userAttributesUpdated` hard-coded to `false`, and `StreamEvaluationsRequest` has no field for the client to signal an attribute change. On reconnect the diff only included flags whose `UpdatedAt` fell within the grace window, so a rule-based flag whose evaluation changed only because of new user attributes was silently missing — the client kept the stale variation. The same update over REST returned the correct rule-matched variation, since `GetEvaluationsRequest.UserEvaluationCondition.UserAttributesUpdated` reaches the evaluator there.

Since the server cannot tell whether the attributes changed while the client was disconnected, the initial put after a (re)connect now always passes `userAttributesUpdated=true`, pulling every rule-bearing flag into the diff. This fixes all SDK versions already in the field without a protocol or SDK change.

## Points

- Rule-based flags are re-evaluated on every reconnect initial put, even when attributes did not change — bounded to reconnects; live patches on an open connection pass `checkUserAttributes=false` since the user cannot change mid-connection.
- A follow-up can add `UserEvaluationCondition` to `StreamEvaluationsRequest` so SDKs signal the change explicitly and the server only re-evaluates rule flags when attributes actually changed.
- e2e: `TestStreamReconnectAfterUserAttributesUpdate` reproduces the exact issue scenario, advancing the `evaluatedAt` cursor past `featureFlagDiffGracePeriod` instead of sleeping through the 10-minute window.


pkg/api/api/api.go
```diff
@@ -442,17 +443,34 @@ func (s *gatewayService) evaluateFeaturesForStream(
...
+		prevUEID, evaluatedAt, userAttributesUpdated, tag,
 	)
 	if err != nil {
 		return "", nil, err
 	}
-	ueid = evaluation.UserEvaluationsID(user.Id, user.Data, features)
 	return ueid, evaluations, nil
 }
 
+func filterFeaturesByTag(fs []*featureproto.Feature, tag string) []*featureproto.Feature {
```
Ubisoft-potato, line 461: This is identical to `(*grpcGatewayService).filterByTag` in `api_grpc.go:2097`, and that one doesn't use its receiver. Could we make a single package-level helper and call it from both paths, so the tag-filter semantics for `GetEvaluations` and the stream can't drift apart?

pkg/api/api/api.go
```diff
@@ -442,17 +443,34 @@ func (s *gatewayService) evaluateFeaturesForStream(
 		return "", nil, err
 	}
 
+	ueid = evaluation.UserEvaluationsID(user.Id, user.Data, filterFeaturesByTag(features, tag))
```
Ubisoft-potato, line 446: Nit: the in-place sort in `UserEvaluationsID` now runs on the fresh slice returned by `filterFeaturesByTag`, and `TopologicalSort` builds a new slice too, so the `features` copy above (L417-420) no longer protects anything. Either drop it or reword its comment so it doesn't point at a sort that no longer touches `features`.

pkg/api/api/api.go
```diff
@@ -442,17 +443,34 @@ func (s *gatewayService) evaluateFeaturesForStream(
...
+	// The stream request cannot signal a user attribute change, but the
+	// attributes are part of the UEID hash, so a mismatch on reconnect forces
+	// rule-based flags to be re-evaluated.
+	userAttributesUpdated := checkUserAttributes && prevUEID != "" && prevUEID != ueid
```
t-kikuc: `prevUEID != ueid` fires on feature-definition changes too — not only on attribute changes. This makes the heuristic imprecise: when a flag config changes server-side (but attributes are the same), `userAttributesUpdated` is set to `true` unnecessarily.

I think we don't need to calculate UEID here, and this is enough:

```go
userAttributesUpdated := checkUserAttributes && prevUEID != ""
```

Or, `prevUEID != nil` is checked in `EvaluateFeaturesByEvaluatedAt()`, so this is also enough:
```go
	evaluations, err := evaluator.EvaluateFeaturesByEvaluatedAt(
		features, user, segmentUsersMap, segmentsMap,
		prevUEID, evaluatedAt, checkUserAttributes, tag,
	)
```

pkg/api/api/api.go
```diff
@@ -442,17 +443,34 @@ func (s *gatewayService) evaluateFeaturesForStream(
...
 		features, user, segmentUsersMap, segmentsMap,
-		prevUEID, evaluatedAt, false, tag,
+		prevUEID, evaluatedAt, userAttributesUpdated, tag,
 	)
 	if err != nil {
 		return "", nil, err
 	}
-	ueid = evaluation.UserEvaluationsID(user.Id, user.Data, features)
 	return ueid, evaluations, nil
 }
```
t-kikuc, line 457: If the UEID comparison is dropped in favor of `userAttributesUpdated := checkUserAttributes && prevUEID != ""`, this helper can be removed.

pkg/api/api/api.go
```diff
@@ -442,17 +443,34 @@ func (s *gatewayService) evaluateFeaturesForStream(
...
+				break
+			}
+		}
+	}
+	return result
+}
+
 func (s *gatewayService) getTargetFeatures(fs []*featureproto.Feature, id string) ([]*featureproto.Feature, error) {
 	feature, err := s.findFeature(fs, id)
 	if err != nil {
```
t-kikuc: This `ueid` cannot be deleted since we have to return the last evaluated UEID.

pkg/api/api/api.go
```diff
@@ -442,17 +443,34 @@ func (s *gatewayService) evaluateFeaturesForStream(
...
+	userAttributesUpdated := checkUserAttributes && prevUEID != "" && prevUEID != ueid
 	evaluations, err := evaluator.EvaluateFeaturesByEvaluatedAt(
 		features, user, segmentUsersMap, segmentsMap,
-		prevUEID, evaluatedAt, false, tag,
+		prevUEID, evaluatedAt, userAttributesUpdated, tag,
 	)
 	if err != nil {
 		return "", nil, err
 	}
-	ueid = evaluation.UserEvaluationsID(user.Id, user.Data, features)
```
t-kikuc, line 452: This cannot be deleted because we have to return the last evaluated UEID

t-kikuc (commented):
Thanks a lot
Basically it's right

pkg/api/api/api.go
hvn2k1: Resolved in 103d1ab55 by removing `filterFeaturesByTag` entirely: per the other review thread, the UEID-mismatch comparison is dropped, and the stream UEID is computed over all features again (as before this PR), so there is no second copy of the tag-filter logic left to dedupe.

pkg/api/api/api.go
hvn2k1: After 103d1ab55 the UEID is computed over `features` again, so the copy and its comment are accurate as-is. Note the copy is load-bearing either way: `evaluate()` also calls `UserEvaluationsID` on the slice it receives (evaluation/go/evaluation.go:222), which sorts it in place — so the singleflight-shared slice needs the copy regardless of which slice the handler hashes.

pkg/api/api/api.go
hvn2k1: kept in 103d1ab55
The UEID is still computed after evaluation and returned in the response. Only the mismatch comparison and the `filterFeaturesByTag` helper were removed; `checkUserAttributes` is now passed directly to `EvaluateFeaturesByEvaluatedAt` as you suggested.

t-kikuc (approved):
thank you!

PR #2817, opened 2026-09-16
feat(auditlog): organization audit logs for organization admins

hvn2k1:
Part of #1982, Follows #2791

## What this PR does

Lets organization admins read their organization's audit logs, API and console. `ListAuditLogs`/`GetAuditLog` accept an `organization_id` (organization admin role required): alone it scopes to the organization's organization-level logs (rows with an empty `environment_id`); combined with `environment_id` it returns that environment's logs plus the organization-level logs in one timeline, with correct pagination and total count. The console's audit logs page sends both IDs for organization admins, so they see the combined timeline by default; Environment, Project, and Organization are added to the entity type dropdown as filters.

## Background

Since #2791, organization-level events are written to `audit_log` scoped by `organization_id` — but only system admins could read them. This is the read path: organization admins see their organization's activity inline with the environment they're viewing, environment members see their environment exactly as before. Same console page, no new screen.

## Points

- **The scope comes from the verified role, never from a client-supplied filter.** `organization_id` present → organization ADMIN check (org admins implicitly pass environment checks, so one check covers the union). Environment-only requests must carry a non-empty `environment_id` — previously an empty one could reach the environment role check, where accounts holding a role on the empty-string environment would match.
- **One storage path.** A single scope helper builds the WHERE parts for list and get on both MySQL and PostgreSQL: both IDs → `(environment_id = ? OR (organization_id = ? AND environment_id = ''))`, one ID → that scope alone. The organization branch structurally pins `environment_id = ''`, so a cross-organization read stays impossible at the storage layer. A new `AndFilter` primitive in the query builders expresses the grouped OR term.
- Editor avatars are stored per environment, so organization-only responses skip avatar enrichment.
- The console never lets the user pick an organization — it comes from the signed-in account; non-admin members are unchanged, and the server enforces the role regardless of what the client sends.
- Intentionally NOT done here: no unfiltered system-admin view on `ListAuditLogs` (that stays on `ListAdminAuditLogs` until the history merge lands).


pkg/auditlog/api/api.go
```diff
@@ -109,21 +109,40 @@ func (s *auditlogService) GetAuditLog(
...
 	localizer := locale.NewLocalizer(ctx)
-	_, err := s.checkEnvironmentRole(
-		ctx, accountproto.AccountV2_Role_Environment_VIEWER,
-		req.EnvironmentId)
-	if err != nil {
-		return nil, err
+	if req.EnvironmentId == "" && req.OrganizationId == "" {
+		return nil, statusMissingEnvironmentOrOrganization.Err()
+	}
+	if req.EnvironmentId != "" {
```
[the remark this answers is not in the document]
hvn2k1: Right — this contradicted the ListAuditLogs contract. Fixed in b763859: GetAuditLog now selects the organization scope whenever organization_id is set (org admin check + GetOrganizationAuditLog + avatar skip all keyed on the same decision), with a test asserting a request carrying both IDs takes the organization path.

ui/dashboard/src/pages/audit-logs/page-content.tsx
```diff
@@ -189,6 +191,11 @@ const PageContent = () => {
       {!!auditLogId && (
         <AuditLogDetailsModal
           auditLogId={auditLogId}
+          organizationId={
+            isOrganizationAdmin && isOrganizationEntityType(filters?.entityType)
+              ? consoleAccount?.organization.id
+              : undefined
```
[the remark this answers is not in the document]
hvn2k1: Good catch. Fixed in 8f291f6: the copied entry link now keeps the current query string, so the entityType the details scope is derived from survives the deep link.

pkg/auditlog/storage/v2/mysql/audit_log.go
```diff
@@ -205,6 +239,17 @@ func listAuditLogsOptionsFromParams(p v2als.ListAuditLogsParams) (*mysqlstorage.
...
+	if p.OrganizationID != "" {
+		filters = append(filters, &mysqlstorage.FilterV2{
+			Column:   "organization_id",
+			Operator: mysqlstorage.OperatorEqual,
+			Value:    p.OrganizationID,
+		}, &mysqlstorage.FilterV2{
+			Column:   "environment_id",
+			Operator: mysqlstorage.OperatorEqual,
+			Value:    "",
+		})
```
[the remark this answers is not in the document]
hvn2k1: Added in b763859: TestListAuditLogsOrganizationScopeMySQL builds the options from an organization-scoped params struct and asserts both the list and count queries contain organization_id = ? AND environment_id = ? with args ["org-1", ""].

pkg/auditlog/storage/v2/postgres/audit_log.go
```diff
@@ -211,6 +245,17 @@ func listAuditLogsOptionsFromParams(p v2als.ListAuditLogsParams) (*pgstorage.Lis
...
+	if p.OrganizationID != "" {
+		filters = append(filters, &pgstorage.Filter{
+			Column:   "organization_id",
+			Operator: pgstorage.OperatorEqual,
+			Value:    p.OrganizationID,
+		}, &pgstorage.Filter{
+			Column:   "environment_id",
+			Operator: pgstorage.OperatorEqual,
+			Value:    "",
+		})
```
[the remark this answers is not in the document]
hvn2k1: Added in b763859: TestListAuditLogsOrganizationScopePostgres asserts both the list and count queries contain organization_id = $1 AND environment_id = $2 with args ["org-1", ""].

pkg/auditlog/storage/v2/mysql/audit_log.go
```diff
@@ -37,6 +37,8 @@ var (
 	insertAuditLogV2SQL string
 	//go:embed sql/auditlog/select_audit_logs_v2.sql
 	selectAuditLogsV2SQL string
+	//go:embed sql/auditlog/select_organization_audit_log_v2.sql
+	selectOrganizationAuditLogV2SQL string
```
t-kikuc: [ask] Why not just extend the existing SQLs and UI, instead of adding new ones?
Since we use a single table for both env auditLog and admin auditLog, i guess we can make it simpler.

hvn2k1: We are extending the existing UI. But for the query, at first I use another method to check if that's org audit log (by event type) and it was cumbersome.
However right now I'm using the org ID itself, so let me try implement the reuse first and decide

hvn2k1: updated in dc0f34c. There's now one storage method and one handler

hvn2k1: Reopening to this thread: I find that when we enter audit log console, we need to retrieve both organization audit log and the env log

pkg/auditlog/api/api.go
```diff
@@ -173,36 +193,89 @@ func (s *auditlogService) ListAuditLogs(
...
 ) (*proto.ListAuditLogsResponse, error) {
 	localizer := locale.NewLocalizer(ctx)
-	_, err := s.checkEnvironmentRole(
-		ctx, accountproto.AccountV2_Role_Environment_VIEWER,
-		req.EnvironmentId)
+	// organization_id requires the organization admin role; combined with
+	// environment_id it lists that environment's logs plus the organization's
+	// organization-level logs in one timeline.
+	if req.OrganizationId != "" {
+		_, err := s.checkOrganizationRole(
```
Ubisoft-potato, line 200: When both IDs are sent, only the organization ADMIN check runs, and that check never ties `environment_id` to the organization. An admin of org A can send `organization_id=A` with an `environment_id` from org B, and the union clause `environment_id = ? OR (organization_id = ? AND environment_id = '')` returns org B's environment logs. Environment rows carry an empty `organization_id` (`pkg/domainevent/domain/event.go:55`), so storage can't pin this either.

I verified this by running the handler with an org-A admin and `environment_id=env-of-org-B`: only `GetAccountV2(organization_id=org-A)` is called, `GetAccountV2ByEnvironmentID` is never called, and storage is reached with both IDs.

Suggest running `checkEnvironmentRole` whenever `environment_id` is non-empty, in addition to the org check. Its account lookup joins `account_v2.organization_id` to `environment_v2.organization_id`, so a foreign environment is rejected while org admins still pass for their own environments. Same applies to `GetAuditLog` (line 117). Worth an API test for the foreign-environment case.

proto/auditlog/service.proto
```diff
@@ -53,7 +57,13 @@ message ListAuditLogsRequest {
   int64 from = 7;
   int64 to = 8;
   google.protobuf.Int32Value entity_type = 9;
+  // Scopes the list to one environment; requires an environment role.
   string environment_id = 10;
+  // Requires the organization admin role. Alone, lists the organization's
+  // organization-level logs (environment_id is empty on those); combined with
+  // environment_id, lists that environment's logs plus the organization-level
+  // logs in one timeline.
+  string organization_id = 11;
```
t-kikuc: [ask] Could the server derive `organization_id` instead of taking it from the client?

The console only sends what it already knows about the signed-in account (`collection-loader/index.tsx:47`, `page-content.tsx:194`), and `GetAccountV2ByEnvironmentID` already `INNER JOIN`s `account_v2.organization_id = environment_v2.organization_id`, so `checkEnvironmentRole` can return the caller's organization and admin flag. That would drop this field, the swagger/`pb.go`/`proto.lock` churn and 6 UI files, and the cross-org read @Ubisoft-potato found becomes unrepresentable rather than guarded.

The cost I can see is the organization-only mode (no `environment_id`) — is anything relying on it? #1982 asks only for the existing page, which always sends both. Anything else you would lose?

pkg/auditlog/api/api.go
```diff
@@ -140,6 +154,12 @@ func (s *auditlogService) GetAuditLog(
 	auditlog.LocalizedMessage = domainevent.LocalizedMessage(auditlog.Type, localizer)
 	s.obfuscateAPIKey(auditlog)
 
+	// Editor avatars are stored per environment; organization-only scope skips them.
```
t-kikuc: `select_avatar_accounts_v2.sql` is `FROM account_v2 a INNER JOIN environment_v2 e ON a.organization_id = e.organization_id`, and `GetAvatarAccountsV2` adds only `e.id = ?` (`pkg/account/storage/v2/mysql/account.go:239`). The avatar columns live on `account_v2`, so the row is per organization and `e.id` only selects which one — in the combined scope this same call already returns avatars for organization-level editors who hold no role in the environment.

So the stated reason is wrong, and the skip costs the organization-only response avatars it could serve. Please fix the comment (same wording at line 227) and either enrich that path too or say it is deferred.

pkg/auditlog/api/api.go
hvn2k1: nice catch, I've fixed this in https://github.com/bucketeer-io/bucketeer/pull/2817/commits/8e511d10def7ed9635646a31f4fc49e3eb89555e

Ubisoft-potato (approved):
The rest LGTM! Thanks!

pkg/auditlog/storage/v2/mysql/audit_log.go
```diff
@@ -49,15 +47,53 @@ func NewAuditLogStorage(qe mysqlstorage.QueryExecer) v2als.AuditLogStorage {
...
 	ctx context.Context,
-	id string,
-	environmentID string,
+	id, environmentID, organizationID string,
 ) (*proto.AuditLog, error) {
+	scopeFilters, orFilters := auditLogScopeFilters(environmentID, organizationID)
+	filters := append([]*mysqlstorage.FilterV2{
+		{Column: "id", Operator: mysqlstorage.OperatorEqual, Value: id},
+	}, scopeFilters...)
+	query, whereArgs := mysqlstorage.ConstructQueryAndWhereArgs(selectAuditLogsV2SQL, &mysqlstorage.ListOptions{
```
t-kikuc: The organization scope is unverified on this path. `TestListAuditLogsOrganizationScope{MySQL,Postgres}` only exercise `listAuditLogsOptionsFromParams`, but `GetAuditLog` builds its `ListOptions` inline here, so nothing asserts the WHERE clause or args it generates — `TestGetAuditLogMySQL`/`Postgres` pass `gomock.Any()` for both and always call with an empty `organizationID`.

Deleting `select_audit_log_v2.sql` removed the one place this scope could be read, so the generated query is now its only definition. Please add the assertions you added for the list path — `id = ?` plus the scope parts, with args — on both drivers.

t-kikuc (approved):
let's try

PR #2818, opened 2026-09-17
feat(migration): merge the admin audit log history into audit_log

hvn2k1:
Part of #1982, Follows #2811

Merges the resolved rows from the `audit_log_migration` working table into `audit_log`. Idempotent and a no-op where the working table is empty. `admin_audit_log` and the working table are kept and dropped in a later release.

the `audit_log_migration` table will be deleted in another pr


migration/mysql/20260917000000_merge_audit_log_migration.sql
```diff
@@ -0,0 +1,10 @@
+INSERT INTO audit_log (
+  id, timestamp, entity_type, entity_id, type,
+  event, editor, options, entity_data, previous_entity_data,
+  environment_id, organization_id
+)
+SELECT id, timestamp, entity_type, entity_id, type,
+       event, editor, options, entity_data, previous_entity_data,
+       '', organization_id
+FROM audit_log_migration
```
t-kikuc: The merge does not distinguish "no organization" from "organization unresolved". `audit_log_migration` holds both: system-level rows (`entity_type` 7 / 11) that legitimately have no organization, and rows the eight ordered rules declined. Both land in `audit_log` with `organization_id = ''`, and re-applying this file after an improved resolution pass cannot correct the second group, because the `ON DUPLICATE KEY UPDATE` is a deliberate no-op — so "idempotent" holds only in the sense of not duplicating. Could the insert be restricted to `organization_id <> '' OR entity_type IN (7, 11)`, so declined rows stay in the working table until they resolve?

migration/mysql/20260917000000_merge_audit_log_migration.sql
```diff
@@ -0,0 +1,10 @@
+INSERT INTO audit_log (
```
t-kikuc: Both preceding migrations in this series (`20260826000000`, `20260907000000`) open with a header comment stating intent and why the statement is safe; this one has none. The two decisions a later reader will stop on are the bare `''` for `environment_id` (organization-level rows are stored with an empty environment id) and the self-assignment in `ON DUPLICATE KEY UPDATE` (intentional no-op for re-runs). Could you add a short header covering both?

t-kikuc (approved):
thanks

